PR #60, opened 2026-03-25
mcp-logging

KIvanow:
> [!NOTE]
> **Medium Risk**
> Adds a new authenticated `POST /mcp/telemetry` endpoint and background client-side flushing, which could affect MCP request flow/performance and increases event ingestion surface area. Failures are largely swallowed/optional, but payload validation and volume limits should be reviewed.
> 
> **Overview**
> Adds end-to-end **MCP tool-call telemetry**: the MCP CLI now wraps each tool in `withTelemetry`, buffers success/duration/error info, periodically flushes it to `POST /mcp/telemetry`, and flushes on shutdown (opt-out via `BETTERDB_TELEMETRY=false`).
> 
> On the API side, `McpController` gains `POST /mcp/telemetry` (guarded with the existing `AgentTokenGuard`), validates 1–100 events, and forwards them to a new `UsageTelemetryService.trackMcpToolCall`; `McpModule` now imports `TelemetryModule` (optional injection keeps community/self-hosted behavior graceful).
> 
> Bumps package versions to `0.12.0` across the monorepo and `@betterdb/mcp` to `1.1.0`, and tightens MCP API prefix detection to require `res.ok`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6920f0bfc1edae5b1c4991681a5837e337145c9d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "web",
-  "version": "0.11.0",
+  "version": "0.12.0",
   "homepage": "https://betterdb.com",
   "private": true,
   "type": "module",
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "api",
-  "version": "0.11.0",
+  "version": "0.12.0",
   "homepage": "https://betterdb.com",
   "private": true,
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "betterdb-monitor",
-  "version": "0.11.0",
+  "version": "0.12.0",
   "private": true,
   "workspaces": [
     "apps/*",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@betterdb/monitor",
-  "version": "0.11.1",
+  "version": "0.12.0",
   "description": "Monitor and observe your Valkey/Redis databases",
   "bin": {
     "betterdb": "./bin/betterdb.js"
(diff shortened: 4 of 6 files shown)


apps/web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "web",
-  "version": "0.11.0",
+  "version": "0.12.0",
```
jamby77, line 3: So each package version should be bumped to the same version?

packages/mcp/src/index.ts
jamby77, line 1: I think this file is doing too much stuff, I would break it down a little bit

packages/mcp/src/telemetry.ts
```diff
@@ -0,0 +1,54 @@
...
+function flush(): Promise<void> {
+  if (buffer.length === 0 || !apiRequestFn) return Promise.resolve();
+  const events = buffer;
+  buffer = [];
+  return apiRequestFn('POST', '/mcp/telemetry', { events })
+    .then(() => {})
+    .catch(() => {
+      // swallow — telemetry must never break the MCP server
+    });
+}
```
jamby77, line 28: Is there a reason why it can't be just async function?
```suggestion
asyn function flush(): Promise<void> {
  if (buffer.length === 0 || !apiRequestFn) return;
  const events = buffer;
  buffer = [];
  try {
    await apiRequestFn('POST', '/mcp/telemetry', { events })
  } catch() {
      // swallow — telemetry must never break the MCP server
  }
}
```

apps/web/package.json
KIvanow: A few are connected - shared/api/web all create the general application for instance, while mcp, agent, etc are completely separate

PR #145, opened 2026-05-04
feature: add PostHog reverse proxy to bypass ad blockers

KIvanow:
Adds a /ingest proxy route so PostHog events are sent through the app's own domain instead of directly to us.i.posthog.com, preventing ad blocker interference.

- Vite dev server proxies /ingest → https://eu.i.posthog.com
- NestJS controller handles /ingest/* in production via native fetch
- Unit and e2e tests cover URL routing, body/header forwarding, and error codes

## Changes
- 

## Checklist
- [ ] Docs added / updated
- [x] [Roborev](https://www.roborev.io/) review passed — run `roborev review --branch` or `/roborev-review-branch` in Claude Code *(internal)*

---

> [!NOTE]
> **Medium Risk**
> Adds a new unauthenticated `/ingest/*` proxy path and adjusts global API prefixing to exclude it, which can affect routing and introduces an outbound request surface area. Risk is moderate but scoped because the upstream host is fixed via `POSTHOG_HOST` and behavior is covered by unit/e2e tests.
> 
> **Overview**
> Adds a new `PosthogProxyModule` exposing `/ingest/*` that forwards requests to `POSTHOG_HOST` via `fetch`, passing through method, JSON body (for non-GET/HEAD), and `content-type`, and returning upstream status/body/`content-type`.
> 
> Updates production routing to keep `/ingest/*` *outside* the `/api` global prefix, and configures Vite dev server to proxy `/ingest` to PostHog. Adds unit tests for URL rewriting/headers/body handling and an e2e suite validating happy-path and upstream error status forwarding.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a30163f85daa40785329b9a0a8ba54767d9f1f25. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/api/src/posthog-proxy/posthog-proxy.controller.ts
+++ b/apps/api/src/posthog-proxy/posthog-proxy.controller.ts
@@ -2,7 +2,7 @@ import { All, Controller, Req, Res } from '@nestjs/common';
 import { SkipThrottle } from '@nestjs/throttler';
 import { FastifyRequest, FastifyReply } from 'fastify';
 
-const POSTHOG_HOST = 'https://us.i.posthog.com';
+const POSTHOG_HOST = 'https://eu.i.posthog.com';
 
 @SkipThrottle()
 @Controller('ingest')
--- a/apps/api/src/posthog-proxy/posthog-proxy.controller.ts
+++ b/apps/api/src/posthog-proxy/posthog-proxy.controller.ts
@@ -2,7 +2,7 @@ import { All, Controller, Req, Res } from '@nestjs/common';
 import { SkipThrottle } from '@nestjs/throttler';
 import { FastifyRequest, FastifyReply } from 'fastify';
 
-const POSTHOG_HOST = 'https://eu.i.posthog.com';
+const POSTHOG_HOST = process.env.POSTHOG_HOST ?? 'https://eu.i.posthog.com';
 
 @SkipThrottle()
 @Controller('ingest')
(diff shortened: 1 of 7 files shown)


apps/api/src/posthog-proxy/posthog-proxy.controller.ts
```diff
@@ -0,0 +1,34 @@
+import { All, Controller, Req, Res } from '@nestjs/common';
+import { SkipThrottle } from '@nestjs/throttler';
+import { FastifyRequest, FastifyReply } from 'fastify';
+
+const POSTHOG_HOST = 'https://eu.i.posthog.com';
```
jamby77: We have this (or should have this) as env variable, having it in 2 places may lead to drift in case of a change

apps/web/vite.config.ts
```diff
@@ -8,6 +8,13 @@ export default defineConfig({
   envDir: path.resolve(__dirname, '../..'),
   server: {
     port: 5173,
+    proxy: {
```
jamby77, line 11: If we have the controller, why don't we proxy to it directly?

apps/api/src/posthog-proxy/posthog-proxy.controller.ts
KIvanow: Nice catch. Fixed it

apps/web/vite.config.ts
KIvanow: It's one fewer network hop (browser → Vite → PostHog vs browser → Vite → NestJS → PostHog)

jamby77 (approved):

PR #156, opened 2026-05-07
refactor: replace hardcoded 'env-default' strings with ENV_DEFAULT_ID

Vswaroop04:
## Summary

* Removes the stale `TODO` comment from `connection-registry.service.ts` (the constant was already defined but not consistently used)
* Introduces a single source of truth by importing `ENV_DEFAULT_ID` in:
  * `connections.dto.ts` (Swagger example values)
  * `connection-registry.service.spec.ts` (test fixtures)
  * Seed scripts
* Seed scripts define a local alias with a reference comment to avoid introducing a NestJS dependency in standalone scripts

## Why

Previously, renaming the sentinel value required updating 28+ hardcoded string literals across multiple files. This change centralizes the value, making future updates safer and easier.

## Note

SQL schema defaults (`DEFAULT 'env-default'`) in SQLite and PostgreSQL adapters are intentionally unchanged. These are database-level literals within migration/template strings, not part of application logic.

## Test Plan

* [x] Run `pnpm test` and verify all existing registry tests pass
* [x] Confirm via grep that there are no remaining bare `'env-default'` string literals in application code (excluding SQL schema and migration strings)

### DIFF
--- a/apps/api/scripts/seed-demo-data-pg.ts
+++ b/apps/api/scripts/seed-demo-data-pg.ts
@@ -9,12 +9,15 @@
 import { Pool } from 'pg';
 import { randomUUID } from 'crypto';
 
+// Canonical value lives in connection-registry.service.ts (ENV_DEFAULT_ID).
+const ENV_DEFAULT_ID = 'env-default';
+
 const STORAGE_URL =
   process.env.STORAGE_URL ||
   `postgresql://${process.env.STORAGE_POSTGRES_USER || 'betterdb'}:${process.env.STORAGE_POSTGRES_PASSWORD || 'devpassword'}@${process.env.STORAGE_POSTGRES_HOST || 'localhost'}:${process.env.STORAGE_POSTGRES_PORT || '5432'}/${process.env.STORAGE_POSTGRES_DATABASE || 'betterdb'}`;
 
 // Auto-detected at runtime from the most recently active connection
-let CONNECTION_ID = process.env.CONNECTION_ID || 'env-default';
+let CONNECTION_ID = process.env.CONNECTION_ID || ENV_DEFAULT_ID;
 let SOURCE_HOST = process.env.SOURCE_HOST || 'localhost';
 let SOURCE_PORT = parseInt(process.env.SOURCE_PORT || '6379', 10);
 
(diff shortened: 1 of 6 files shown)


Vswaroop04:

Vswaroop04:
Hey @KIvanow 👋

I’ve been playing around with BetterDB and really like what you’re building, especially the MCP integration and semantic cache. While going through the codebase, I came across a few areas that could use some cleanup or improvements and put together a few contributions.

Happy to iterate based on your feedback. No rush at all

KIvanow:
Hey @Vswaroop04 , thank you so much for the contribution! 

Thanks for the clean refactor - the intent is right. A few things to address before this can merge:

1. Dependency direction in connections.dto.ts
  Importing from connection-registry.service inside a DTO inverts the layering — DTOs should have no knowledge of the service layer. While there's no circular dependency today, this is fragile. The fix is to extract the constant to a dedicated file:                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                          
  // apps/api/src/connections/connection.constants.ts                                                                                                                                                                                                                                                                                                                                                                                     
  export const ENV_DEFAULT_ID = 'env-default';                                                                                                                                                                                                                                                                                                                                                                                            
  
  Then import from there in both the service and the DTO. As a bonus, the seed scripts could also import from this file directly (it has no NestJS dependencies), eliminating the local duplicate copies.                                                                                                                                                                                                                                 
                  
2. SQL injection pattern in seed-demo-data-pg.ts                                                                                                                                                                                                                                                                                                                                                                                        
  // current                                                                                                                                                                                                                                                                                                                                                                                                                              
  `WHERE connection_id <> '${ENV_DEFAULT_ID}'`

  String interpolation in SQL is the injection pattern regardless of whether the value is currently safe. Please use a parameterized query:                                                                                                                                                                                                                                                                                               
  
  pool.query(                                                                                                                                                                                                                                                                                                                                                                                                                             
    `SELECT connection_id, source_host, source_port FROM command_log_entries
     WHERE connection_id <> $1
     ORDER BY captured_at DESC LIMIT 1`,
    [ENV_DEFAULT_ID]                                                                                                                                                                                                                                                                                                                                                                                                                      
  )
                                                                                                                                                                                                                                                                                                                                                                                                                                          
3. Unsigned commits

  The commit 130c0e4 is not GPG-signed.

KIvanow:
Hey @Vswaroop04  all three issues from the first review are cleanly addressed, thanks for the quick turnaround. One last
  minor thing before this merges:

  connection-registry.service.spec.ts still imports ENV_DEFAULT_ID from '../connection-registry.service' (the re-export)
  rather than from '../connection.constants' directly:

  // current
  import { ConnectionRegistry, ENV_DEFAULT_ID } from '../connection-registry.service';

  // preferred
  import { ConnectionRegistry } from '../connection-registry.service';
  import { ENV_DEFAULT_ID } from '../connection.constants';

  The re-export works, but it slightly undermines the extraction - the spec is still coupled to the service for a value that now lives independently. A one-line import split fixes it. After that this is good to go.

KIvanow (approved):

PR #159, opened 2026-05-08
test(inference-latency): add unit tests for InferenceLatencyService

Vswaroop04:
## Summary

Adds `inference-latency.service.spec.ts`. The core service previously had no direct unit test coverage despite existing tests for related components like `bucketing`, `percentiles`, `correlation`, and the controller layer.

Includes 15 test cases covering both `getProfile` and `getTrend`, including:

* Empty result handling for both slowlog and commandlog sources
* Source selection logic based on connection capabilities
* Bucket grouping and per-bucket entry aggregation
* `FT.SEARCH` buckets being flagged as `unhealthy` when p50 exceeds the configured threshold
* Correct propagation of `thresholdUs` from connection config, with fallback to `0` on error
* Proper forwarding of explicit `[startTime, endTime]` windows to storage
* Validation error handling for:

  * only `startTime` provided
  * `endTime <= startTime`
  * exceeded bin count limits
  * non-positive `bucketMs`
* Trend bin assignment and cross-bucket filtering behavior

## Why

The service contained a significant amount of business and validation logic but had no direct unit tests around the orchestration layer itself. This fills the coverage gap around source selection, validation paths, threshold handling, and time-window aggregation behavior.

## Test Plan

* [ ] `cd apps/api && pnpm test`

### DIFF
--- a/apps/api/src/inference-latency/__tests__/inference-latency.service.spec.ts
+++ b/apps/api/src/inference-latency/__tests__/inference-latency.service.spec.ts
@@ -1,0 +1,307 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { Test, TestingModule } from '@nestjs/testing';
+import { NotFoundException } from '@nestjs/common';
+import { InferenceLatencyService, InferenceLatencyValidationError } from '../inference-latency.service';
+import { ConnectionRegistry } from '../../connections/connection-registry.service';
+import { PrometheusService } from '../../prometheus/prometheus.service';
+
+// ---------------------------------------------------------------------------
+// Shared test fixtures
+// ---------------------------------------------------------------------------
+
+function makeSlowLogEntry(overrides: Record<string, unknown> = {}) {
+  return {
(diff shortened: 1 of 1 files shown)


KIvanow (changes requested):
Hi @Vswaroop04 great PR overall and thank you for your contribution! Please sign the commit so it can be merged

apps/api/src/inference-latency/__tests__/inference-latency.service.spec.ts
```diff
@@ -0,0 +1,307 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
KIvanow, line 1: Blanket eslint-disable @typescript-eslint/no-explicit-any at file level is a bit of a code smell. The any is only needed for the buildModule parameters and storage mock arrays - could be scoped to just those lines, or typed with Partial<StoragePort> etc. No need to change, just keep in mind in general.

KIvanow (approved):

PR #160, opened 2026-05-09
feat(semantic-cache): add Google AI (Gemini) embedding provider

Vswaroop04:
## Summary

Adds Google AI embedding support to both packages, following the same provider pattern already used for Cohere and Voyage.

### Added providers

* packages/semantic-cache/src/embed/google.ts
* TypeScript provider using native fetch
* packages/semantic-cache-py/betterdb_semantic_cache/embed/google.py
* Python provider using httpx

Both providers target the embedContent REST API and support:

* Default model: text-embedding-004(768 dimensions)
* Configurable taskType
* defaults to RETRIEVAL_QUERY
* use RETRIEVAL_DOCUMENT when storing documents
* Optional title support for improved retrieval quality with document embeddings
* Optional outputDimensionality truncation (text-embedding-004 and newer)
* API key from GOOGLE_API_KEY or explicit config
* No dependency on the official Google SDK

---

### TypeScript

```ts
import { createGoogleEmbed } from '@betterdb/semantic-cache/embed/google';

const embed = createGoogleEmbed({
  model: 'text-embedding-004'
});

const cache = new SemanticCache({
  client,
  embedFn: embed
});
```

### Python

```python
from betterdb_semantic_cache.embed.google import create_google_embed

embed = create_google_embed(
    model="text-embedding-004"
)

cache = SemanticCache(
    SemanticCacheOptions(
        client=client,
        embed_fn=embed
    )
)
```

---

## Test

* pnpm test passes in packages/semantic-cache
* Manually verified against a live GOOGLE_API_KEY
* createGoogleEmbed()("hello world") returns a valid 768-element float embedding array

### DIFF
--- a/packages/semantic-cache-py/betterdb_semantic_cache/embed/google.py
+++ b/packages/semantic-cache-py/betterdb_semantic_cache/embed/google.py
@@ -1,0 +1,90 @@
+"""Google AI (Gemini) embedding helper for betterdb-semantic-cache.
+
+Uses the Google AI REST API directly via httpx.
+Requires the 'httpx' extra: pip install betterdb-semantic-cache[httpx]
+
+Usage::
+
+    from betterdb_semantic_cache.embed.google import create_google_embed
+    embed = create_google_embed(model="text-embedding-004")
+    cache = SemanticCache(SemanticCacheOptions(client=client, embed_fn=embed))
+"""
+from __future__ import annotations
+
+import os
+from typing import Any, Literal
+
+from betterdb_semantic_cache.types import EmbedFn
+
+GoogleEmbedTaskType = Literal[
+    "RETRIEVAL_QUERY",
+    "RETRIEVAL_DOCUMENT",
+    "SEMANTIC_SIMILARITY",
+    "CLASSIFICATION",
+    "CLUSTERING",
+]
+
+
+def create_google_embed(
+    *,
+    model: str = "text-embedding-004",
+    api_key: str | None = None,
+    base_url: str = "https://generativelanguage.googleapis.com/v1beta",
+    task_type: GoogleEmbedTaskType = "RETRIEVAL_QUERY",
+    title: str | None = None,
+    output_dimensionality: int | None = None,
+) -> EmbedFn:
+    """Create an EmbedFn backed by the Google AI (Gemini) Embeddings API.
+
+    Args:
+        model: Google AI embedding model. Default: 'text-embedding-004' (768-dim).
+               Other options: 'text-multilingual-embedding-002', 'embedding-001'.
+        api_key: Google AI API key. Default: GOOGLE_API_KEY env var.
+        base_url: API base URL.
+        task_type: Task type hint. Default: 'RETRIEVAL_QUERY'.
+                   Use 'RETRIEVAL_DOCUMENT' when storing documents.
+        title: Optional document title. Only used with task_type='RETRIEVAL_DOCUMENT'.
+        output_dimensionality: Optional output dimensionality (truncation).
+                               Supported by text-embedding-004+.
+    """
+    _client: list[Any] = []
+
+    async def _get_client() -> Any:
+        if not _client:
+            try:
+                import httpx
+            except ImportError:
+                raise ImportError(
+                    'betterdb-semantic-cache embed/google requires the "httpx" package. '
+                    "Install it: pip install betterdb-semantic-cache[httpx]"
+                )
+            _client.append(httpx.AsyncClient(timeout=30))
+        return _client[0]
+
+    async def embed(text: str) -> list[float]:
+        key = api_key or os.environ.get("GOOGLE_API_KEY")
+        if not key:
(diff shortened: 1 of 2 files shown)


Vswaroop04:
@KIvanow @jamby77  would appreciate any feedback or thoughts on the approach here

KIvanow (changes requested):
Hey @Vswaroop04 thank you for your contribution! A few issues though:

 - Unsigned commit (259e5f1) - same as your other three open PRs.
  - API key passed as a URL query parameter - inconsistent with every other provider and a security regression.
  Every existing provider (Cohere, Voyage, OpenAI, etc.) passes the key in a header:
  // cohere.ts / voyage.ts
  headers: { Authorization: `Bearer ${apiKey}` }
  - The Google implementation puts the key in the URL:
  // google.ts - don't do this
  `${baseUrl}/models/${model}:embedContent?key=${encodeURIComponent(apiKey)}`
  - And Python:
  params={"key": key}
  - Keys in URLs end up in server access logs, proxy logs, and HTTP Referer headers. Google's API supports x-goog-api-key as a request header - please use that instead, consistent with how every other provider in this package works.                                                                                                                                                                                                  
  - httpx.AsyncClient is never closed (Python). The lazy-initialized client in _client[0] holds a connection pool with no cleanup path. In test environments this produces ResourceWarnings; in production it leaks file descriptors over time. Either expose a close() helper or document that callers are responsible for cleanup, but the current code silently leaks.
  - title behaviour is inconsistent between TS and Python. TypeScript skips title when falsy (if (opts?.title)), so an empty string is silently dropped. Python sends it when title is not None, so "" would be included in the request body. Pick one semantics and apply it to both - !== undefined / is not None is the more predictable choice.

Vswaroop04:
> Hey @Vswaroop04 thank you for your contribution! A few issues though:
> 
>     * Unsigned commit ([259e5f1](https://github.com/BetterDB-inc/monitor/commit/259e5f1eaf77c943d5238fdaea42c762cb752087)) - same as your other three open PRs.
> 
>     * API key passed as a URL query parameter - inconsistent with every other provider and a security regression.
>       Every existing provider (Cohere, Voyage, OpenAI, etc.) passes the key in a header:
>       // cohere.ts / voyage.ts
>       headers: { Authorization: `Bearer ${apiKey}` }
> 
>     * The Google implementation puts the key in the URL:
>       // google.ts - don't do this
>       `${baseUrl}/models/${model}:embedContent?key=${encodeURIComponent(apiKey)}`
> 
>     * And Python:
>       params={"key": key}
> 
>     * Keys in URLs end up in server access logs, proxy logs, and HTTP Referer headers. Google's API supports x-goog-api-key as a request header - please use that instead, consistent with how every other provider in this package works.
> 
>     * httpx.AsyncClient is never closed (Python). The lazy-initialized client in _client[0] holds a connection pool with no cleanup path. In test environments this produces ResourceWarnings; in production it leaks file descriptors over time. Either expose a close() helper or document that callers are responsible for cleanup, but the current code silently leaks.
> 
>     * title behaviour is inconsistent between TS and Python. TypeScript skips title when falsy (if (opts?.title)), so an empty string is silently dropped. Python sends it when title is not None, so "" would be included in the request body. Pick one semantics and apply it to both - !== undefined / is not None is the more predictable choice.

Thanks for the thorough review! All issues addressed Should have caught these before opening nice finds!

KIvanow (changes requested):
Hey @Vswaroop04 - all four issues from the first review are fixed, thanks! One small thing before this merges:

  The close() helper is a good addition but it's invisible to callers — nothing in the module docstring or create_google_embed docstring mentions it exists. Please add a note, e.g.:

```
  When finished, release the connection pool:
      await embed.close()

```
One line in the docstring is enough. After that this is good to go.

Vswaroop04:
> Hey @Vswaroop04 - all four issues from the first review are fixed, thanks! One small thing before this merges:
> 
> The close() helper is a good addition but it's invisible to callers — nothing in the module docstring or create_google_embed docstring mentions it exists. Please add a note, e.g.:
> 
> ```
>   When finished, release the connection pool:
>       await embed.close()
> ```
> 
> One line in the docstring is enough. After that this is good to go.

Done

KIvanow (approved):
Re-reviewed at `4a49295` and everything from both rounds is in - verified against the code:

- API key now sent via `x-goog-api-key` header in both TS and Python, no more key-in-URL.
- `close()` helper releases the httpx pool, and it's documented in the `create_google_embed` docstring.
- `title` handling is consistent across both (`!== undefined` / `is not None`).
- Commits are signed now.

Structure matches the Cohere/Voyage providers cleanly. One optional follow-up (non-blocking): `title` is sent regardless of `task_type`, but Google only honors it with `RETRIEVAL_DOCUMENT` and may reject it otherwise - worth either gating it on the task type or noting the constraint more loudly. Nothing that needs to hold up the merge.

Approving - thanks for the careful iteration, @Vswaroop04! 🚀

PR #174, opened 2026-05-11
feature(monitor): post-capture filters + JSON/CSV export

jamby77:
## Summary

PR 11 of 25 in `docs/plans/specs/monitor-command/plan-implementation.md`. Stacked on top of #173 (PR 10). **Closes Phase 2 frontend MVP.** The session-detail page gains a Filters & export panel that filters the captured stream by command, client, key glob, and time window — then downloads the filtered slice as JSON or CSV via a new server endpoint.

### Backend

- **`monitor-line.parser.ts`** — pure module
  - `parseMonitorLine(text)` → `{ts, tsRaw, db, addr, cmd, args, key, raw}`. Handles backslash-escaped quotes, IPv6 bracket addresses, keyless commands (PING etc.).
  - `matchesFilters(line, {command, client, key, afterTs, beforeTs})`: case-insensitive command exact match, addr substring match, glob match on the first arg / key (`*` and `?`), inclusive timestamp window.
  - `lineToCsvRow` + `CSV_HEADER`: RFC 4180 escaping for commas, quotes, newlines.
- **`GET /monitor/sessions/:id/export?format=json|csv`** (+ filter query params) streams persisted chunks, parses each line, applies filters, emits either `{sessionId, count, lines[]}` JSON or a CSV with a header row. `Content-Disposition: attachment; filename=monitor-session-<id>.<fmt>`. 404 on unknown session. Default format is `json` when an unknown value is supplied.

### Frontend

- New `FiltersAndExport` component on the session-detail page (below the live tail).
  - Five inputs: Command, Client, Key glob, After (datetime-local), Before (datetime-local).
  - **In-page preview count** derived from the live-tail buffer (parser inlined to avoid a server round-trip on every keystroke). Export buttons hit the server endpoint, which sees the full session.
  - Two `<a download>` buttons styled as outline Buttons — clean browser-native download with the URL carrying the current filters.
- `useMonitorTail` lifted from `TailView` up to `MonitorSession` so the Filters panel and the TailView share one buffer.

## Test plan

- [ ] `SKIP_DOCKER_SETUP=true pnpm --filter api test -- --testPathPatterns "monitor|capture-sessions|health-gate|provider-detector|acl-checker|preflight|capture-writer|demo-mode|tail|monitor-line"` → 184 tests across 14 suites, all pass
- [ ] `pnpm --filter web exec tsc --noEmit` → exit 0
- [ ] Run dev: `MONITOR_DEV_PREVIEW=true pnpm dev:api` and `VITE_MONITOR_DEV_PREVIEW=true pnpm dev:web`
- [ ] Start a session, send mixed traffic, let it complete
- [ ] `curl '/monitor/sessions/:id/export?format=json'` → `{count, lines[]}` with parsed structure
- [ ] `curl '/monitor/sessions/:id/export?format=json&command=GET'` → filtered subset
- [ ] `curl '/monitor/sessions/:id/export?format=json&key=user:*'` → globbed subset
- [ ] `curl '/monitor/sessions/:id/export?format=csv'` → starts with `ts,ts_raw,db,addr,cmd,args,key`
- [ ] Browser: navigate to a completed session, type `GET` in Command field → preview "Buffer match: N of M lines" updates
- [ ] Add `user:*` in Key glob → preview narrows further; Export-link URL carries `command=GET&key=user%3A*`
- [ ] Click Export JSON → file downloads with `monitor-session-<id>.json`; opens in browser and matches the preview count
- [ ] Click Export CSV → file downloads with `monitor-session-<id>.csv`; first line is the header

## Notes for reviewers

- **IPv6 bracket addresses caught my parser early** — the initial header regex used a greedy `[^\]]+` that stopped at the inner `]` of `[::1]:port`. Fixed with a non-greedy `(.+?)\]\s+"` anchor. Covered by a unit test.
- **The frontend parser is duplicated.** The in-page preview count needs to filter the live-tail buffer client-side; I inlined a small parser in `filters-and-export.tsx` to avoid round-tripping on every keystroke. The parser is intentionally simpler than the backend's (no `args` array, no escape decoding) since it only feeds filter predicates. If we ever surface the parsed fields in the UI, the right move is a shared `@betterdb/shared` parser — flagged.
- **AUTH redaction** in the CSV output above is a *server-side feature in iovalkey's MONITOR stream* (visible in the docker dev run — Valkey itself redacts AUTH args to `(redacted)`), not something this PR adds. We get it for free from the underlying protocol.
- **Net diff: 707 lines.** Above the 400 target. Composition: parser 142 + parser tests 159 + filters-and-export panel 191 + controller endpoint 83 + controller tests 92 + small wiring. Tight enough to review as one slice but a parser-only split would have left an unverifiable PR.
- **Screenshot at `docs/assets/pr11-filters-export.png`.**

## Stacked PR

Base branch is `feature/monitor-web-live-tail` (#173), so the diff shown is ONLY PR 11 changes.

## Phase 2 MVP done

This closes Phase 2 (frontend MVP). Next is Phase 3: cross-reference engine + UI panel — the actual differentiator the spec calls out.


KIvanow (approved):
```
  monitor-line.parser.ts:114 (backend):
  function globToRegex(glob: string): RegExp {
    const escaped = glob
      .replace(/[.+^${}()|[\]\\]/g, '\\$&')
      .replace(/\*/g, '.*')
      .replace(/\?/g, '.');
    return new RegExp(`^${escaped}$`);
  }
```
and
```
  filters-and-export.tsx:185 (frontend):
  function globMatch(s: string, glob: string): boolean {
    const escaped = glob
      .replace(/[.+^${}()|[\]\\]/g, '\\$&')
      .replace(/\*/g, '.*')
      .replace(/\?/g, '.');
    return new RegExp(`^${escaped}$`).test(s);
  }
```
Are almost identical. Not blocking, but it would be better if they were extracted somehow to avoid odd issues when only one is changed in the future

PR #142, opened 2026-04-30
feat(anomaly): sentinel/cluster failover event tracking with slowlog correlation

SBALAVIGNESH123:
## Summary

Implements [Issue #28](https://github.com/BetterDB-inc/monitor/issues/28) — Sentinel/Cluster failover event tracking with slowlog correlation.

This PR adds end-to-end failover detection, persistence, and correlation to the anomaly detection pipeline, enabling post-incident analysis of failover events alongside slowlog spikes.

### Shared Types (`packages/shared`)
- Added `FAILOVER_STARTED` and `FAILOVER_COMPLETED` to `WebhookEventType` enum
- Registered both as **PRO tier** events in `WEBHOOK_EVENT_TIERS` and `PRO_EVENTS`
- Extended `IWebhookEventsProService` interface with `dispatchFailoverStarted()` and `dispatchFailoverCompleted()`

### Anomaly Detection (`proprietary/anomaly-detection`)
- Added `CLUSTER_STATE` to `MetricType` enum
- **Replication role detection** now tracks both directions:
  - `master -> replica` demotion: `CRITICAL` / `DROP` + `failover.started` webhook
  - `replica -> master` promotion: `WARNING` / `SPIKE` + `failover.completed` webhook
- **Cluster state tracking**: detects `ok -> fail` and `fail -> ok` transitions via `CLUSTER INFO`
- Excluded `CLUSTER_STATE` from z-score buffer loop (handled via state-change detection)
- Added `lastClusterState` cleanup in `onConnectionRemoved()`
- Injected `WebhookEventsProService` (optional, PRO-only)

### Correlation Engine
- Enhanced `NODE_FAILOVER` pattern rule with `optionalMetrics: [SLOWLOG_LAST_ID, CLUSTER_STATE, OPS_PER_SEC]`
- Added dedicated `CLUSTER_STATE` -> `NODE_FAILOVER` pattern rule
- Added `enrichDiagnosis()` method that dynamically includes slowlog correlation context when `SLOWLOG_LAST_ID` co-occurs with failover events in the same 5s window

### Webhook PRO Service
- Added `dispatchFailoverStarted()` and `dispatchFailoverCompleted()` methods with license gating

## Testing

- **50 unit tests pass** across `correlator.spec.ts`, `spike-detector.spec.ts`, `metric-buffer.spec.ts`
[...]

PR #125, opened 2026-04-24
refactor(storage): extract WebhookRepository across all three storage adapters

amitkojha05:
## Summary

Implements the first step of the TODO on `postgres.adapter.ts` line 49:
Extracts all webhook and delivery methods from `PostgresAdapter`, `SqliteAdapter`, and `MemoryAdapter` into dedicated repository classes under `apps/api/src/storage/adapters/repositories/`. Each adapter becomes a thin facade that delegates to its repository. Follows the same pattern established in PR #13 (MultiConnectionPoller base class extraction).

## Changes

- `repositories/webhook.postgres.repository.ts` — new, owns all Postgres webhook/delivery SQL
- `repositories/webhook.sqlite.repository.ts` — new, owns all SQLite webhook/delivery SQL  
- `repositories/webhook.memory.repository.ts` — new, owns all in-memory webhook/delivery logic
- `postgres.adapter.ts` — delegates to `WebhookPostgresRepository`, retains `if (!this.pool)` guards
- `sqlite.adapter.ts` — delegates to `WebhookSqliteRepository`, retains `if (!this.db)` guards
- `memory.adapter.ts` — delegates to `WebhookMemoryRepository`, Maps moved into repository

Zero behavior changes. All three adapters implement the same `StoragePort` interface contract as before.

The updated TODO now lists the remaining domains to extract:

## Checklist

- [x] Unit / integration tests added — existing webhook-storage test suite covers all three adapters via the StoragePort interface; zero test changes required
- [ ] Docs added / updated — no user-facing docs affected by this refactor
- [ ] Roborev review passed — *(internal)*
- [ ] Competitive analysis done / discussed — *(internal)*
- [ ] Blog post about it discussed — *(internal)*

PR #191, opened 2026-05-12
refactor(storage): extract SlowLogRepository across all three storage adapters

amitkojha05:
## Summary

Extracts slow log query logic from the three storage adapters into dedicated `SlowLogRepository` classes, continuing the repository extraction started in #125 (webhooks).

**New files:**
- `repositories/slowlog.postgres.repository.ts`
- `repositories/slowlog.sqlite.repository.ts`
- `repositories/slowlog.memory.repository.ts`

**Modified files:**
- `postgres.adapter.ts` — delegates to `SlowLogPostgresRepository`
- `sqlite.adapter.ts` — delegates to `SlowLogSqliteRepository`
- `memory.adapter.ts` — delegates to `SlowLogMemoryRepository`

## Changes

The TODO comment in `postgres.adapter.ts` (left in #125) lists slowlog as the next domain to extract:

```
// Domain-specific repositories (webhooks extracted). Remaining domains to extract:
// ACL, anomaly, slowlog, commandlog, latency, ...
```

This removes ~200 lines of inline query logic from each SQL adapter and isolates slow log storage behind testable, single-responsibility classes.

## Implementation

Same pattern as #125:

- Repository constructors take **non-nullable** connections (`Pool`, `Database.Database`) — the type system guarantees a live connection
- Adapters retain `if (!this.pool)` / `if (!this.db)` lifecycle guards  and delegate after initialization
- Memory repository is self-contained (owns its own `StoredSlowLogEntry[]`)
- `entries.length === 0` short-circuit lives in both the adapter (preserving original behaviour) and the repository (for standalon  testability)

```typescript
// Adapter — owns lifecycle, keeps guard
async saveSlowLogEntries(entries: StoredSlowLogEntry[], connectionId: string): Promise<number> {
  if (!this.pool || entries.length === 0) return 0;
  return this.slowlogRepo.saveSlowLogEntries(entries, connectionId);
}

// Repository — non-nullable Pool, pure query logic
constructor(private readonly pool: Pool, private readonly mappers: RowMappers) {}

async saveSlowLogEntries(entries: StoredSlowLogEntry[], connectionId: string): Promise<number> {
  if (entries.length === 0) return 0;
  // ... SQL ...
}
```

## Testing

Pure extraction — zero SQL changes, zero logic changes, zero interface changes. All existing tests pass unchanged.

## CI note
CI note: the anomaly.service.spec.ts failure (getClusterInfo mock missing at line 733) was introduced in #142 (feat: sentinel/cluster failover — commit 2fa1c46) and is pre-existing on master. Unrelated to this  SlowLog extraction.

### DIFF
--- a/apps/api/src/storage/adapters/memory.adapter.ts
+++ b/apps/api/src/storage/adapters/memory.adapter.ts
@@ -50,6 +50,7 @@ import type {
 } from '@betterdb/shared';
 import { PROPOSAL_DEFAULT_EXPIRY_MS, variantPayloadSchemaFor } from '@betterdb/shared';
 import { WebhookMemoryRepository } from './repositories/webhook.memory.repository';
+import { SlowLogMemoryRepository } from './repositories/slowlog.memory.repository';
 
 const NULL_SUB_DISCRIMINATOR = '__betterdb_null__';
 
@@ -73,7 +74,6 @@ export class MemoryAdapter implements StoragePort {
   private clientSnapshots: StoredClientSnapshot[] = [];
   private anomalyEvents: StoredAnomalyEvent[] = [];
   private correlatedGroups: StoredCorrelatedGroup[] = [];
-  private slowLogEntries: StoredSlowLogEntry[] = [];
   private commandLogEntries: StoredCommandLogEntry[] = [];
   private latencySnapshots: StoredLatencySnapshot[] = [];
   private latencyHistograms: StoredLatencyHistogram[] = [];
@@ -83,6 +83,7 @@ export class MemoryAdapter implements StoragePort {
   private settings: AppSettings | null = null;
   private readonly MAX_DELIVERIES_PER_WEBHOOK = 1000;
   private readonly webhookRepo = new WebhookMemoryRepository(this.MAX_DELIVERIES_PER_WEBHOOK);
+  private readonly slowlogRepo = new SlowLogMemoryRepository();
   private idCounter = 1;
   private ready: boolean = false;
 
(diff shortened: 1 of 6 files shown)


amitkojha05:
Hi @KIvanow ,PR  is up — extracts SlowLogRepository across all three adapters, continuing the pattern from #125.
Happy to get your feedback.

KIvanow (changes requested):
Thanks @amitkojha05 - I went through this carefully and the extraction itself is clean. I diffed the SQL that moved out of each adapter against the new repositories and it's identical (the postgres `INSERT ... ON CONFLICT DO NOTHING`, the `SELECT ... ORDER BY timestamp DESC LIMIT/OFFSET`, `SELECT MAX(slowlog_id)`, and the `DELETE ... captured_at` all match byte-for-byte), and the adapters now do nothing but keep their lifecycle guards and delegate. No logic or interface changes snuck in. Nice continuation of the #125 pattern.

No concerns with the code. Two things before I approve, both about getting a trustworthy green signal rather than the diff itself:

**1. Sign the commit.** `891a582c` shows `verified=false` - same as your other branches. Please amend with your signing key and force-push; the repo enforces verification.

**2. Rebase onto current master so CI actually validates the extraction.** You're right that the failing `anomaly.service.spec.ts` is pre-existing and unrelated to slowlog - but it's a *TypeScript compile error*, which fails the entire Jest run, so the storage/slowlog tests never actually executed in this CI run. That means "all existing tests pass unchanged" has no green run backing it yet. This branch is also 192 commits behind master and diverged. A rebase onto current master (where that anomaly failure should already be resolved) lets the suite compile and actually exercise the new repositories - which is exactly the confidence a pure-extraction refactor needs before merging.

Once it's signed and I can see a clean run that runs the storage tests, this is a quick approve. Thanks again - this is a tidy refactor.

amitkojha05:
Thanks @KIvanow — appreciate the careful diff-check on the SQL, 

Both addressed in `1cf1526`:

 **1. Signed.** Amended with my SSH signing key and force-pushed; commit now shows `Verified`. 

 **2. Rebased onto current master.** Branch is now up to date — the pre-existing `anomaly.service.spec.ts` compile error is resolved on master, so the Jest run compiles and the storage/slowlog suites actually execute this time. Both checks (`API Tests` + `CLA Assistant`) are green on the current tree, so "existing tests pass unchanged" now has a real signal behind it.

KIvanow (approved):
Approving. ✅ Thanks @amitkojha05 - this is a tidy continuation of the #125 pattern, and both items from the last review are now sorted.

**Prior review items - resolved:**

1. **Signed** - `1cf1526` verifies clean.
2. **Rebased + real CI signal** - the branch is on current master and `api-tests` now passes (3m8s). That's the important one: last time the pre-existing `anomaly.service.spec.ts` TypeScript compile error was failing the whole Jest run, so "all existing tests pass unchanged" had no green run behind it. Now the suite compiles and the storage tests actually execute, which is exactly the confidence a pure-extraction refactor needs.

**On the extraction itself:** I re-diffed all three moved blocks and the SQL/logic is byte-for-byte identical to what left the adapters - the Postgres `INSERT ... ON CONFLICT DO NOTHING`, the `SELECT ... ORDER BY timestamp DESC LIMIT/OFFSET`, `SELECT MAX(slowlog_id)`, the `DELETE ... captured_at`, the SQLite `INSERT OR IGNORE` + transaction, and the in-memory filter/sort/slice all match. Adapters keep their lifecycle guards and delegate; the memory repo cleanly owns its own `StoredSlowLogEntry[]` with no leftover references in the adapter. No logic or interface changes snuck in.

**One optional follow-up (not blocking):** the repos are now isolatable, but there's no dedicated storage test for them - slowlog is currently covered indirectly (`connection-isolation.spec.ts` + the analyzer/analytics specs), and there's no `slowlog-storage.spec.ts` mirroring the `webhook-storage.spec.ts` that #125 added. A small dedicated spec in a later PR would bank the testability this extraction unlocks. Happy to leave that out of scope here.

Nice, low-risk refactor - merging.

PR #198, opened 2026-05-14
fix(storage): enforce cache proposal constraints at application level

Vswaroop04:
Summary

SQLite on the Linux CI build silently skips the expression-based partial unique indexes used for the cache proposal constraints (COALESCE(json_extract(...))). Because of that, the UNIQUE and CHECK constraints are not actually enforced at the database level in CI, which causes the failing cache-proposals-sqlite.spec.ts tests.

This change adds matching validation inside createCacheProposal so the behaviour stays consistent even when SQLite does not create those indexes correctly. The validation mirrors the intended database constraints and prevents duplicate pending proposals from being created.

Some existing PR test runs are currently failing because of this issue as well, so getting this merged should stabilise those SQLite CI failures.

One thing I wanted to clarify with maintainers (@KIvanow , @jamby77)  : should duplicate pending proposals ever be allowed as a fallback when SQLite skips the expression index, or is strict enforcement the intended behaviour across all environments?

Also verified that proposals for different categories are still allowed correctly.

### DIFF
--- a/apps/api/src/storage/adapters/sqlite.adapter.ts
+++ b/apps/api/src/storage/adapters/sqlite.adapter.ts
@@ -3364,6 +3364,43 @@ export class SqliteAdapter implements StoragePort {
 
   async createCacheProposal(input: CreateCacheProposalInput): Promise<StoredCacheProposal> {
     if (!this.db) throw new Error('Database not initialized');
+
+    const validTypes: Record<string, string[]> = {
+      semantic_cache: ['threshold_adjust', 'invalidate'],
+      agent_cache: ['tool_ttl_adjust', 'invalidate'],
+    };
+    if (!validTypes[input.cache_type]?.includes(input.proposal_type)) {
+      throw new Error(
+        `CHECK constraint failed: invalid combination cache_type='${input.cache_type}' proposal_type='${input.proposal_type}'`,
+      );
+    }
+
+    const payload = input.proposal_payload as Record<string, unknown>;
+    if (input.proposal_type === 'threshold_adjust') {
+      const category = payload.category ?? null;
+      const dupe = this.db
+        .prepare(
+          `SELECT id FROM cache_proposals
+           WHERE connection_id = ? AND cache_name = ? AND proposal_type = 'threshold_adjust'
+             AND status = 'pending'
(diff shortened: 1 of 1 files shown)


Vswaroop04:
@KIvanow Can you pls have a look at this PR?

KIvanow (changes requested):
Thanks @Vswaroop04 - nice diagnosis. You correctly traced this to SQLite silently skipping the `json_extract` expression indexes in CI, and I checked the app-level guards against the schema: the `validTypes` map matches the DDL `CHECK` exactly, and the two duplicate-check queries reproduce the partial unique indexes faithfully (the `COALESCE(json_extract(...), '__betterdb_null__')` + `status = 'pending'` predicate). Commit's signed and CI is green. Good fix.

To answer your question directly: **yes, strict enforcement is the intended behavior across all environments**, and app-level validation as defense-in-depth is the right call - it also protects us in production if the same SQLite limitation shows up there, not just in CI. So let's keep the approach. A few things before it merges:

**a) Enforcement is now asymmetric, and that's fine for this PR - but let's be intentional.** Postgres relies on real DB constraints; SQLite now relies on this app-level guard. I'm ok leaving Postgres as-is (its partial indexes work), so no need to touch it here. But please open a follow-up issue to investigate *why* the SQLite expression index no-ops in CI (JSON1 / SQLite version), so we're not silently depending on the app-level path forever. The guard is the safety net; the index should still be the primary mechanism.

**b) De-duplicate the constraint definitions - they're now a second source of truth.** The `validTypes` map and the dup-check predicates mirror the schema DDL, but nothing keeps them in sync. Concretely: we have a `ttl_adjust` proposal type coming in #161, and this hardcoded map would silently reject it. Please pull the `cache_type -> allowed proposal_type` mapping into a shared constant that both the schema/CHECK and this validation reference (or at minimum, a comment on both the map and the DDL pointing at each other), so the next person adding a proposal type can't miss one side.

**c) Rebase onto current master.** This branch is ~200 commits behind. CI is green so the risk is lower than a stale refactor, but please rebase and re-run so we're validating against current master before merge.

Once the shared constant is in and it's rebased, this is an approve. Thanks again - this is a well-scoped fix and the right question to have asked.

PR #202, opened 2026-05-15
fix(mcp): use package version in server metadata

### LINKED ISSUE #152: MCP server reports version 0.1.0 instead of 1.2.0
The `McpServer` instance in `packages/mcp/src/index.ts:183` is hardcoded with `version: '0.1.0'`, while the actual package version is `1.2.0`. This causes MCP clients that rely on the server version to receive incorrect information.

**File:** `packages/mcp/src/index.ts:183`

epowell40:
## Summary
- read the MCP server version from packages/mcp/package.json instead of hard-coding 0.1.0
- keep a 1.2.0 fallback so server metadata stays aligned with the published package

Fixes #152.

## Test
- corepack pnpm install --filter @betterdb/mcp --ignore-scripts
- corepack pnpm --filter @betterdb/mcp build

Note: full workspace install attempted first but the Windows environment is missing the ClangCL build toolset required by hnswlib-node.

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -2,8 +2,12 @@
 
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
+import { createRequire } from 'node:module';
 import { z } from 'zod';
 import { initTelemetry, trackToolCall, stopTelemetry } from './telemetry.js';
+
+const require = createRequire(import.meta.url);
+const { version: PACKAGE_VERSION = '1.2.0' } = require('../package.json') as { version?: string };
 
 // --- CLI arg parsing ---
 
@@ -180,7 +184,7 @@ async function withTelemetry(toolName: string, fn: () => Promise<ToolResult>): P
 
 const server = new McpServer({
   name: 'betterdb',
-  version: '0.1.0',
+  version: PACKAGE_VERSION,
 });
 
 server.tool(


epowell40:

KIvanow (changes requested):
Thanks for this @epowell40 - clean fix, and the right approach. I traced it through the build: `tsc` emits `src/index.ts` to a flat `dist/index.js`, so `createRequire(import.meta.url)` + `require('../package.json')` resolves to the package root both locally and in the published tarball (package.json is always published). Using `createRequire` is also the correct idiom here since the package is `"type": "module"` - no tsconfig change needed. This genuinely closes #152; it'll now report the real version instead of a hardcoded one.

No issue with the mechanism. A few things before it can merge:

**1. Rebase - the PR currently has merge conflicts.** `index.ts` has moved quite a bit in the ~152 commits since this branched, and GitHub shows it as conflicting. Please rebase onto current master and resolve.

**2. Get a green CI run.** Right now only the CLA check has run - `api-tests` never executed (the conflict likely blocked it), so there's no signal on the change yet. After the rebase, let's make sure the full suite runs green.

**3. Sign the commit.** `8cfa7cb2` shows `verified=false`; the repo enforces signature verification. Please amend with your signing key and force-push.

**Minor:** the `= '1.2.0'` fallback reintroduces a smaller version of the same drift this PR removes - it's already stale (the package is at 1.3.1). It should never actually fire since the path is correct, but if it did it'd report a plausible-but-wrong number. I'd fall back to something that's obviously a fallback, like `'0.0.0'` or `'unknown'`, instead.

Once it's rebased, signed, and green, this is an easy approve. Thanks!

PR #222, opened 2026-05-26
fix(mcp): report package version from metadata

RitwijParmar:
## Summary\n- derive the MCP server version from packages/mcp/package.json instead of a hardcoded 0.1.0\n- keeps the MCP handshake version aligned with the published package version\n\nFixes #152.\n\n## Test\n- pnpm --filter @betterdb/mcp build

### DIFF
--- a/packages/mcp/src/index.ts
+++ b/packages/mcp/src/index.ts
@@ -2,8 +2,12 @@
 
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
+import { createRequire } from 'node:module';
 import { z } from 'zod';
 import { initTelemetry, trackToolCall, stopTelemetry } from './telemetry.js';
+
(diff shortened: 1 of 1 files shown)


RitwijParmar:

recheck

RitwijParmar:
recheck

RitwijParmar:

RitwijParmar:
recheck

KIvanow:
Thanks for the fix — the hardcoded `0.1.0` has been stale since the first release, so this is needed.

One minor suggestion: `createRequire` pulls in a CJS shim just to load JSON. A pure-ESM alternative would be more consistent with the rest of the codebase:

```ts
import { readFileSync } from 'node:fs';
const { version } = JSON.parse(readFileSync(new URL('../package.json', import.meta.url), 'utf-8'));
```

Also, can you confirm the `../package.json` path resolves correctly from the compiled output in `dist/`? If the build flattens the directory structure it could break.

Not a blocker — approving as-is.

KIvanow (approved):
LGTM

PR #242, opened 2026-06-15
feat(retrieval): Phase 4 — query (vector + filters + hybrid)

jamby77:
## Phase 4 — `query` (vector + filters + hybrid)

Stacked on **#241 (Phase 3)** — base is the Phase 3 branch, not master.

### What's new
- **`buildFtSearchQuery(schema, k, filter?)`** — pure builder: `(filter)=>[KNN k @vec $vec AS __score]`; TAG → `@f:{escapeTag(v)}`, NUMERIC → `@f:[v v]`. Throws on unknown fields, TEXT fields, and non-numeric values for numeric fields.
- **`Retriever.query(options)`** — embeds `text` or uses a precomputed `vector` (exactly one), runs `FT.SEARCH … PARAMS 2 vec … LIMIT 0 k DIALECT 2`, maps rows → `{ id (prefix stripped), score, text, fields }`; empty → `[]`. Validates `k` is a positive integer and that a precomputed vector matches the index dimension.
- **`hybrid: 'rerank'`** via an injectable `rerankFn(queryText, hits)` (requires `rerankFn` + `text`).
- New `fields.ts` centralizes the reserved field names (`__text`, `__score`); `buildFtCreateArgs` now rejects schema fields using them.

### Public API
Exports `buildFtSearchQuery`, `QueryFilter`, `QueryHit`, `QueryOptions`, `RerankFn`, and the `TEXT_FIELD`/`SCORE_FIELD` constants.

### Tests
20 unit tests (6 builder + 14 query) plus reserved-field guards; full package suite **76/76 green**, `tsc --noEmit` + prettier clean.

### Review-driven changes (pre-PR review)
Added: `k` positive-integer validation, NUMERIC filter value-type check, precomputed-vector dimension check, reserved-field guard at index creation (via shared `fields.ts`), and a stronger rerank assertion verifying it receives mapped hits.

### Deferred (not in this phase)
- **`RETURN` clause** to avoid shipping embedding bytes for every hit — Phase 5 (perf/observability).
- **Rerank over-fetch** (fetch `>k` candidates then trim) — the plan scopes `rerank` to *reordering* k; over-fetch is a future enhancement.
- **`score` semantics**: KNN returns distance (smaller = closer); the field is named `score` per the spec — worth documenting for callers.
- Real-Valkey round-trip lands in **Phase 6** integration.

---

> [!NOTE]
> **Medium Risk**
[...]

PR #243, opened 2026-06-15
feat(retrieval): Phase 5 — discovery, observability & health

jamby77:
## Phase 5 — Discovery, observability & health

Stacked on **#242 (Phase 4)** — base is the Phase 4 branch, not master.

### What's new
- **Discovery** (`discovery.ts`) — `register()`/`unregister()` write a `type:'retrieval'` marker to `__betterdb:caches` (HSET/HDEL); pure `buildRetrievalMarker`.
- **Health** (`health.ts`) — `health()` returns a retrieval-local `IndexHealthSnapshot` (`numDocs`, `indexingState`, `dims`, `percentIndexed` from FT.INFO, `estimatedRecall` via an optional `recallEstimator` hook).
- **Telemetry seam** (`telemetry.ts`) — pluggable `RetrievalMetrics` + `RetrievalTracer` wired into `query`/`upsert` via `instrument()`: operation duration, query-result counts, embedding-call count, optional spans. Zero-cost when absent.
- **Prometheus factory** (`prometheus-metrics.ts`) — `createPrometheusMetrics({registry, prefix})` implementing `RetrievalMetrics`, using get-or-create guards so it is safe to construct twice on a shared registry.

### Tests
14 unit tests (discovery 3, health 3, telemetry 5, prometheus 3); full package suite **90/90 green**; `tsc --noEmit` + prettier clean. `prom-client` added with a frozen-lockfile-valid `pnpm-lock.yaml`.

### Review-driven changes (pre-PR review)
- **get-or-create** Prometheus metrics (mirrors semantic-cache) — avoids a duplicate-registration crash when two Retrievers share a registry.
- Count the dims-probe embedding call so embedding-cost metrics aren't undercounted.
- Added tests for the `instrument()` failure path (metrics + span fire on throw) and the `percentIndexed` missing-field → 0 fallback.

### Deferred (not in this phase)
- **Heartbeat/TTL** for registry markers (stale entry on crash) — inert today since no consumer reads `type:'retrieval'` markers; add when discovery is consumed.
- **Shared `IndexHealthSnapshot`** contract with the Inference Pipeline Latency Profiler — that shared type doesn't exist yet; health stays retrieval-local until the Profiler consumes it.
- Cross-package centralization of `REGISTRY_KEY` / discovery / telemetry (semantic-cache, agent-cache, and retrieval each keep a local copy today) and `RETRIEVAL_VERSION`↔package.json sync — separate refactor.
- Real-Valkey round-trip lands in **Phase 6** integration.

---

> [!NOTE]
> **Low Risk**
> Additive SDK surface and optional observability hooks; registry writes are guarded against clobbering other cache types. No changes to core query/upsert semantics beyond instrumentation wrappers.
> 
> **Overview**
> **Phase 5** extends `@betterdb/retrieval` with **discovery**, **index health**, and **observability** on `Retriever`, plus a **`prom-client`** dependency.
> 
> **Discovery** adds `register()` / `unregister()` that write a `type: 'retrieval'` JSON marker to the shared `__betterdb:caches` hash (same pattern as agent-cache). Registration skips or refuses to overwrite foreign cache types (`agent_cache`, etc.) and only `HDEL`s markers the retriever owns.
> 
> **Health** adds `health()` returning `IndexHealthSnapshot` from `FT.INFO` (`numDocs`, indexing state, dims, `percentIndexed` via `parsePercentIndexed` for both fraction and 0–100 forms). An optional `recallEstimator` hook can populate `estimatedRecall`.
> 
> **Telemetry** introduces pluggable `RetrievalMetrics` and `RetrievalTracer`, wired through `instrument()` on `query` and `upsert` (duration, query hit counts, embedding calls—including dims probe—and spans on success and failure). `createPrometheusMetrics({ registry, prefix })` implements `RetrievalMetrics` with get-or-create metrics safe for shared registries.
> 
> New public exports cover discovery, health helpers, telemetry types, and the Prometheus factory; unit tests cover discovery guards, health parsing, instrumentation, and metrics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8575366bbe84631cf5aae834930e9c2b8a6382b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/retrieval/package.json
+++ b/packages/retrieval/package.json
@@ -35,7 +35,8 @@
     "clean": "rm -rf dist"
   },
   "dependencies": {
-    "@betterdb/valkey-search-kit": "workspace:*"
+    "@betterdb/valkey-search-kit": "workspace:*",
+    "prom-client": "^15.1.3"
   },
   "devDependencies": {
     "@types/node": "^22.19.15",
--- a/packages/retrieval/src/__tests__/discovery.test.ts
+++ b/packages/retrieval/src/__tests__/discovery.test.ts
@@ -1,0 +1,55 @@
+import { describe, it, expect, vi } from 'vitest';
+import { Retriever } from '../retriever';
+import { buildRetrievalMarker, REGISTRY_KEY } from '../discovery';
+import type { RetrievalSchema } from '../schema';
+
+const schema: RetrievalSchema = {
+  fields: { source: { type: 'tag' } },
+  vector: { metric: 'cosine', algorithm: 'hnsw', dims: 4 },
+};
+
+describe('buildRetrievalMarker', () => {
+  it('builds a retrieval registry marker', () => {
+    expect(
+      buildRetrievalMarker({
+        name: 'docs',
+        version: '0.1.0',
+        startedAt: '2026-06-15T00:00:00.000Z',
+      }),
+    ).toEqual({
+      type: 'retrieval',
+      prefix: 'docs',
+      version: '0.1.0',
+      protocol_version: 1,
+      capabilities: ['upsert', 'query', 'delete'],
+      index_name: 'docs:idx',
+      started_at: '2026-06-15T00:00:00.000Z',
+    });
+  });
+});
+
+describe('Retriever discovery', () => {
+  it('registers a retrieval marker on the registry', async () => {
+    const call = vi.fn(async () => 1);
+    const retriever = new Retriever({ client: { call }, name: 'docs', schema });
+
+    await retriever.register();
+
+    const registerCall = call.mock.calls.find((args) => args[0] === 'HSET');
+    expect(registerCall?.[1]).toBe(REGISTRY_KEY);
+    expect(registerCall?.[2]).toBe('docs');
+    const marker = JSON.parse(String(registerCall?.[3]));
+    expect(marker.type).toBe('retrieval');
+    expect(marker.prefix).toBe('docs');
+    expect(typeof marker.started_at).toBe('string');
+  });
+
+  it('unregisters the marker from the registry', async () => {
+    const call = vi.fn(async () => 1);
+    const retriever = new Retriever({ client: { call }, name: 'docs', schema });
+
+    await retriever.unregister();
+
+    expect(call).toHaveBeenCalledWith('HDEL', REGISTRY_KEY, 'docs');
+  });
+});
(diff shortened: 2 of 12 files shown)


KIvanow (changes requested):
Phase 5 is a clean set of seams: telemetry is genuinely zero-cost when absent, `instrument()` wrapping in `finally` means metrics and spans fire on the throw path too, the get-or-create Prometheus registration avoids the duplicate-registration crash on a shared registry, and counting the dims-probe embedding call keeps cost metrics honest. A few things before merging:

**1. `register()` / `unregister()` have no collision guard and share the `{name}` registry field with agent-cache.** Both agent-cache and the memory store in #255 guard cross-type collisions, but retrieval blindly `HSET`s its marker under field `this.name`, which is the exact field agent-cache writes to (`agent-cache/src/discovery.ts:210`). So a retrieval `register()` on a Valkey already hosting a same-named agent-cache silently overwrites that marker, and `unregister()` `HDEL`s whatever marker owns the field, retrieval or not. Since the registry is the shared surface Monitor reads, this is a silent marker loss. Please add a type check before the `HSET` (and only `HDEL` when the field actually holds a retrieval marker), consistent with the sibling packages.

**2. (minor) `RETRIEVAL_VERSION` is hardcoded `'0.1.0'` and will drift from package.json.** The PR notes the sync is deferred, which is fine, but a TODO link in the code pointing at the follow-up would keep it from being forgotten.

**3. (minor) The Prometheus `getSingleMetric(...) as Histogram` / `as Counter` cast is unsafe on a shared registry.** If a different-typed metric already owns that name, the guard passes and the cast throws at first use. Very unlikely given the prefix, but a type check (or a short note) would harden it.

jamby77:
@KIvanow Addressed in ac1cd7d: (1) `register()` now reads the existing marker and skips + `console.warn`s if a different cache type owns the `{name}` field instead of clobbering it; `unregister()` only HDELs a retrieval-owned marker — so no more silent marker loss vs agent-cache. (2) added a TODO on `RETRIEVAL_VERSION` pointing at the package.json-sync follow-up. (3) the Prometheus get-or-create now uses `instanceof Histogram/Counter` instead of a blind cast. Tests added for both collision paths.

KIvanow (approved):

PR #250, opened 2026-06-16
feat(agent-memory): Phase 3 — forget / forgetByScope

jamby77:
## Agent Memory — Phase 3: `forget` / `forgetByScope`

Stacked on **#249 (Phase 2)**.

### What's new
- **`forget(id)`** → DEL `{name}:mem:{id}`; idempotent (returns whether the memory existed).
- **`forgetByScope(scope & {tags?})`** → finds matches via an escaped scope/tag `FT.SEARCH` filter, DELs them in batches, returns the count. **Throws when no scope field or tag is given** (prevents an accidental delete-everything).
- Extracted shared **`buildScopeFilter`** (used by both `recall` and `forget`); `escapeTag` neutralizes glob chars so a `threadId` of `'*'` matches literally, not everything.

### Design note
The plan says "SCAN-matches scope/tags", but scope/tags are HASH fields (TAG-indexed), not part of the key (`{name}:mem:{uuid}`) — a key `SCAN` can't filter on them. So `forgetByScope` uses an `FT.SEARCH` filter query, which is the correct mechanism.

### Tests
30 unit tests (including the batch-cap warning); `tsc --noEmit` + prettier clean.

### Review-driven change
- **Batched pagination instead of a 10000-row cap** — the first pass deleted only the first 10k matches and reported that as success (silent truncation for large scopes). Now loops `FT.SEARCH(LIMIT 0 500)` → DEL batch → repeat, terminating when a batch makes **no progress** (`removed === 0`) so a lagging index that re-lists already-deleted keys can't loop forever. Reaching the `FORGET_MAX_BATCHES` safety cap emits a `console.warn` so a partial delete is surfaced, not silently reported as complete. (Earlier drafts described a `seen`-set guard; the no-progress check supersedes it.)

### Known gap (tracked separately)
`forgetByScope` (like `recall`/`consolidate`) queries `{name}:mem:idx`, which the package does not yet create — index lifecycle (`ensureIndex`) is unimplemented and is being scoped as a dedicated follow-up. `forget(id)` is unaffected (plain `DEL`).

### Next
Phase 4 (recall reinforcement — bump lastAccessedAt/accessCount on recalled items).

---

> [!NOTE]
> **Medium Risk**
> Bulk delete by scope is irreversible and depends on search index correctness; guards (required filter, escaping, batch cap warning) reduce accidental or silent data loss.
> 
> **Overview**
> Adds **memory deletion** to `MemoryStore`: **`forget(id)`** issues a single `DEL` on the memory hash and returns whether anything was removed (idempotent). **`forgetByScope`** finds memories via the same escaped scope/tag **`FT.SEARCH`** filter used for recall, deletes them in **500-key batches** until no matches remain, and returns the total deleted count; it **throws** if no scope field or tag is provided so callers cannot wipe the store by accident.
> 
> **`buildScopeFilter`** is extracted from `buildRecallQuery` so recall and forget share one filter expression (including **`escapeTag`** so literal `*` in IDs cannot broaden the query).
> 
> Large scopes paginate across batches instead of stopping at a fixed row limit; the loop also stops when a batch deletes nothing (stale index re-listing). Hitting **`FORGET_MAX_BATCHES`** logs a **warning** so partial bulk deletes are visible. New unit tests cover single delete, scoped search/DEL, glob escaping, empty scope guard, pagination, and the cap warning.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47693941127364167aca2aef1885a05df65e3ddb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/agent-memory/src/MemoryStore.ts
+++ b/packages/agent-memory/src/MemoryStore.ts
@@ -1,12 +1,13 @@
 import { randomUUID } from 'node:crypto';
 import { encodeFloat32, parseFtSearchResponse } from '@betterdb/valkey-search-kit';
 import { buildMemoryRecord } from './buildMemoryRecord';
-import { buildRecallQuery, SCORE_FIELD } from './buildRecallQuery';
+import { buildRecallQuery, buildScopeFilter, SCORE_FIELD } from './buildRecallQuery';
 import { parseMemoryItem } from './parseMemoryItem';
 import { compositeScore, similarityFromDistance, type RecallWeights } from './compositeScore';
 import type {
   EmbedFn,
   MemoryHit,
+  MemoryScope,
   MemoryStoreClient,
   RecallOptions,
   RememberOptions,
@@ -17,6 +18,8 @@ const DEFAULT_WEIGHTS: RecallWeights = { similarity: 0.6, recency: 0.25, importa
 const DEFAULT_HALF_LIFE_SECONDS = 604800; // 7 days
 const DEFAULT_RECALL_K = 8;
 const RECALL_OVERFETCH = 4;
+const FORGET_BATCH_SIZE = 500;
+const FORGET_MAX_BATCHES = 10000;
 
 export interface MemoryStoreOptions {
   client: MemoryStoreClient;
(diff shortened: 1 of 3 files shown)


KIvanow (changes requested):
A couple of things worth a look.

**1. The PR description and the Cursor summary both describe termination behavior that isn't in the code.** The "Review-driven change" note says `forgetByScope` terminates via "a `seen` set," and the summary says hitting the 10k-batch cap "logs a warning." Neither is present: the loop actually breaks on `removed === 0` (no `seen` set), and it exits the `FORGET_MAX_BATCHES` loop silently with no warning. Looks like the approach changed without updating the prose. Worth syncing the description so reviewers aren't checking against the wrong mechanism, and if the silent cap is intentional, a `console.warn`/throw on cap-hit would match the documented intent.

**2. `forgetByScope` queries `{name}:mem:idx`, which the package still never creates.** This is the same index-creation gap flagged on [#249](https://github.com/BetterDB-inc/monitor/pull/249#issuecomment-4744196227), now landing on the destructive path: `forget(id)` works standalone (plain `DEL`), but `forgetByScope` depends on the missing index, so bulk-forget errors or silently no-ops until the caller hand-rolls the index out-of-band. Disregard if index lifecycle is owned by a follow-up PR, just flag where it lives.

pengo7363:
Thank you

jamby77:
@KIvanow thanks — both addressed:

**1. Description sync.** Updated the PR description: the `seen`-set prose is gone. `forgetByScope` terminates on a no-progress batch (`removed === 0`), and reaching the `FORGET_MAX_BATCHES` cap emits a `console.warn` (in the code at `8ab239a`, lines 157–158) so a partial delete isn't silently reported as complete. So the documented intent and the code now match.

**2. `{name}:mem:idx` is never created by the package.** Confirmed — it was an implicit caller responsibility (the integration suite hand-rolled the `FT.CREATE`), and no phase owned it. Scoped and implemented as a dedicated follow-up: **#261** adds `MemoryStore.ensureIndex()` (idempotent `FT.INFO`→`FT.CREATE` with the full memory schema, reusing `isIndexNotFoundError` from valkey-search-kit), wires it into `AgentMemory.initialize()` so the facade is self-sufficient, and switches the integration suite to dogfood it. That closes the index-lifecycle gap for `forgetByScope`/`recall`/`consolidate`.

KIvanow (approved):

PR #251, opened 2026-06-16
feat(agent-memory): Phase 4 — recall reinforcement

jamby77:
## Agent Memory — Phase 4: recall reinforcement

Stacked on **#250 (Phase 3)**. Small phase.

### What's new
- Recalled items are **reinforced by default**: `recall()` bumps each returned item's `last_accessed_at` (HSET) and `access_count` (HINCRBY).
- **`reinforce: false`** skips the writes.
- Reinforcement is **best-effort** — a failure (`.catch(() => undefined)`) never breaks the recall read path, mirroring agent-cache's "stats failure must not break the cache."

### Tests
3 new unit tests (default-on bump, opt-out, failure-doesn't-break-recall); 30 total. `tsc --noEmit` + prettier clean.

### Next
Phase 5 (eviction & TTL — `selectEvictions` pure selection + capacity/TTL writes).

---

> [!NOTE]
> **Low Risk**
> Localized agent-memory recall behavior with defensive EXISTS checks and swallowed write failures; no auth or shared infra changes.
> 
> **Overview**
> **Recall ranking** now treats recency from `last_accessed_at` (with `Math.max(createdAt, lastAccessedAt)` for clock skew), so bumping access time can raise composite scores on later recalls.
> 
> **After ranking**, `recall()` **reinforces returned hits by default**: `HSET` `last_accessed_at` and `HINCRBY` `access_count` per item, gated by `EXISTS` so stale index hits cannot resurrect deleted hashes. **`reinforce: false`** skips writes; reinforcement errors are swallowed so the read path always returns hits.
> 
> New unit tests cover default reinforcement, opt-out, no resurrection, failure tolerance, and recency-based ranking with `reinforce: false`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd3f48288d383544f1fd23621656440a133399f2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/agent-memory/src/MemoryStore.ts
+++ b/packages/agent-memory/src/MemoryStore.ts
@@ -89,7 +89,11 @@ export class MemoryStore {
         continue;
       }
       const item = parseMemoryItem(this.name, hit);
-      const ageSeconds = (now - item.createdAt) / 1000;
+      // Recency decays from the last access, not creation, so reinforcement
+      // (which bumps last_accessed_at) actually makes a memory more recallable.
+      // max() guards against a clock-skewed last_accessed_at older than created_at.
+      const lastTouched = Math.max(item.createdAt, item.lastAccessedAt);
+      const ageSeconds = (now - lastTouched) / 1000;
       const score = compositeScore({
         similarity: similarityFromDistance(distance),
         ageSeconds,
--- a/packages/agent-memory/src/MemoryStore.ts
+++ b/packages/agent-memory/src/MemoryStore.ts
@@ -104,7 +104,27 @@ export class MemoryStore {
     }
 
     hits.sort((a, b) => b.score - a.score);
-    return hits.slice(0, k);
+    const result = hits.slice(0, k);
+
+    if (options.reinforce !== false) {
+      // Reinforcement is best-effort and must never break the recall read path.
+      await this.reinforce(result, now).catch(() => undefined);
+    }
+    return result;
+  }
+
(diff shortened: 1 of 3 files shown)


KIvanow (changes requested):
One thing I'd like to sort before this merges.

Reinforcement doesn't actually influence recall ranking. `reinforce()` writes `last_accessed_at`, but the composite recency term still decays from `createdAt` (`ageSeconds = (now - item.createdAt) / 1000` in `recall()`); `lastAccessedAt` is parsed onto the item but never read by the scoring path. So reinforcing a memory does not make it any more likely to be recalled next time, which is the feedback loop the name implies. As it stands the write is effectively inert for recall.

If the intent is that `last_accessed_at` only feeds Phase 5 eviction (not recall recency), that's fine, but it's worth saying so explicitly, because right now "reinforcement" reads as if it should boost future recall and it doesn't. If recency is meant to reflect access, the composite should use `max(createdAt, lastAccessedAt)` (or `lastAccessedAt`) for the age calculation. Either way, worth a one-line decision here so the behavior matches the name.

jamby77:
@KIvanow Fixed in fc1be01: recall recency now decays from `max(created_at, last_accessed_at)`, so reinforcing a memory genuinely boosts its future recall (the feedback loop the name implies). Added a test asserting a recently-accessed memory outranks an equally-similar stale one.

KIvanow (approved):

PR #258, opened 2026-06-17
feat(agent-memory): Phase 10 — AgentMemory facade

jamby77:
Stacked on #257 (Phase 9 — observability).

## What
Phase 10 of `@betterdb/agent-memory`: the `AgentMemory` convenience facade so a caller gets all four tiers from one construct.

```ts
const mem = new AgentMemory({
  client, embedFn,
  tierDefaults: { session: { ttl: 1800 } },
  memory: { defaultThreshold: 0.25, recall: { weights, halfLifeSeconds }, maxItemsPerScope: 5000 },
});
await mem.initialize();
mem.llm / mem.tool / mem.session   // short-term, delegated to AgentCache (unchanged)
mem.memory                         // long-term MemoryStore
```

- Wraps `AgentCache` (exposing `.llm/.tool/.session`) + a `MemoryStore` (`.memory`), sharing the **client**, the **resolved name prefix**, and the **telemetry registry**.
- Requires `embedFn` — throws a clear error if missing (checked before any resource is constructed).
- Maps the `memory` sub-config onto `MemoryStore` (verified end-to-end via `currentConfig()`).
- `initialize()` awaits discovery readiness for **both** tiers; `close()` tears down both (memory first, `cache.shutdown()` in a `finally`).

## Design / review notes (fixes applied)
- **Single resolved name**: the name is resolved once (`options.name ?? 'betterdb_ac'`) and passed explicitly to *both* AgentCache and MemoryStore, so their key prefixes / discovery markers / stats keys can never drift even if AgentCache's internal default changes.
- **Symmetric discovery**: AgentCache discovers by default, so the facade now discovers the memory tier by default too (opt-out via `memory.discovery: false`). Standalone `MemoryStore` stays opt-in. Added `MemoryStore.ensureDiscoveryReady()` so `initialize()` can await it.
- **Leak-safe close**: `close()` runs `cache.shutdown()` in a `finally`, so a failing `memory.close()` can't leave cache timers/heartbeats running.
- Distinct metric namespaces (`agent_cache_*` vs `agent_memory_*`) are intentional; only the prom registry is shared.

## Tests (`AgentMemory.test.ts`, 9)
exposes 4 tiers · embedFn-missing throws · memory wired to shared client+prefix · custom name prefix · sub-config mapping via `currentConfig()` · `initialize()`/`close()` drive both (spy) · memory marker registered by default · memory discovery opt-out · one shared prom registry (`agent_cache_*` + `agent_memory_*`).

95/95 package tests green · `tsc` clean · prettier clean.

---

> [!NOTE]
> **Medium Risk**
> New public API and shared Valkey naming/discovery across cache and memory; lifecycle and collision behavior affect how multi-tier apps start and shut down, but changes are mostly composition with tests covering wiring and discovery.
> 
> **Overview**
> Adds the **`AgentMemory`** facade so one constructor exposes short-term **`llm` / `tool` / `session`** (via `AgentCache`) and long-term **`memory`** (`MemoryStore`), with a required **`embedFn`** and optional **`memory`** sub-config for recall thresholds, weights, discovery, and config refresh.
> 
> **Shared wiring:** resolves **`name`** once for both tiers (default `betterdb_ac`), reuses the Valkey **client** and optional **Prometheus registry**; memory discovery is **on by default** on the facade (opt out with `memory.discovery: false`). **`initialize()`** awaits discovery on cache and memory in parallel (cache collisions still throw); **`close()`** shuts down memory then always **`cache.shutdown()`** in a `finally`.
> 
> **`MemoryStore`** gains **`ensureDiscoveryReady()`** so the facade can await registration without changing standalone opt-in discovery behavior. Package **exports** include `AgentMemory`, `AgentMemoryOptions`, and `AgentMemoryConfig`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ee9021ffe02671dabdec4c3f38423efda24d67e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/agent-memory/src/AgentMemory.ts
+++ b/packages/agent-memory/src/AgentMemory.ts
@@ -47,6 +47,9 @@ export class AgentMemory {
 
     const memory = options.memory ?? {};
     this.memory = new MemoryStore({
+      // AgentCacheOptions.client doesn't surface the `.call` method MemoryStore
+      // needs; a real ioredis/iovalkey client has it, so we assert the contract
+      // here. A method-only client/mock would compile but fail at runtime.
       client: options.client as unknown as MemoryStoreClient,
       name,
       embedFn: options.embedFn,
--- a/packages/agent-memory/src/AgentMemory.ts
+++ b/packages/agent-memory/src/AgentMemory.ts
@@ -1,1 +1,80 @@
-export class AgentMemory {}
+import { AgentCache, type AgentCacheOptions } from '@betterdb/agent-cache';
+import {
+  MemoryStore,
+  type MemoryDiscoveryConfig,
+  type MemoryConfigRefreshConfig,
+} from './MemoryStore';
+import type { RecallWeights } from './compositeScore';
+import type { EmbedFn, MemoryStoreClient } from './types';
+
+const DEFAULT_NAME = 'betterdb_ac';
+
+export interface AgentMemoryConfig {
+  defaultThreshold?: number;
+  recall?: {
+    weights?: RecallWeights;
+    halfLifeSeconds?: number;
+  };
+  maxItemsPerScope?: number;
+  discovery?: boolean | MemoryDiscoveryConfig;
+  configRefresh?: boolean | MemoryConfigRefreshConfig;
+}
+
+export interface AgentMemoryOptions extends AgentCacheOptions {
+  embedFn: EmbedFn;
+  memory?: AgentMemoryConfig;
+}
+
+export class AgentMemory {
+  readonly llm: AgentCache['llm'];
+  readonly tool: AgentCache['tool'];
+  readonly session: AgentCache['session'];
+  readonly memory: MemoryStore;
+  private readonly cache: AgentCache;
+
+  constructor(options: AgentMemoryOptions) {
+    if (typeof options.embedFn !== 'function') {
+      throw new Error('AgentMemory requires an embedFn to back the memory tier');
+    }
+
+    // Resolve the name once and hand the same value to both tiers so their key
+    // prefixes, discovery markers, and stats keys can never drift apart.
+    const name = options.name ?? DEFAULT_NAME;
+    this.cache = new AgentCache({ ...options, name });
+    this.llm = this.cache.llm;
+    this.tool = this.cache.tool;
+    this.session = this.cache.session;
+
+    const memory = options.memory ?? {};
+    this.memory = new MemoryStore({
+      client: options.client as unknown as MemoryStoreClient,
+      name,
+      embedFn: options.embedFn,
+      defaultThreshold: memory.defaultThreshold,
+      weights: memory.recall?.weights,
+      halfLifeSeconds: memory.recall?.halfLifeSeconds,
+      maxItemsPerScope: memory.maxItemsPerScope,
+      // The facade is the batteries-included product: discover the memory tier
+      // alongside the cache tiers by default, unless explicitly disabled.
+      discovery: memory.discovery ?? true,
+      configRefresh: memory.configRefresh,
+      telemetry: options.telemetry?.registry ? { registry: options.telemetry.registry } : undefined,
+    });
+  }
+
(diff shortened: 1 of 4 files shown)


KIvanow (changes requested):
The facade composition is clean: resolving `name` once for both tiers, the `finally` on `cache.shutdown()`, and the early `embedFn` check are all the right calls. Three things before merging:

**1. `initialize()` does not bootstrap the FT index, so the facade is not usable end-to-end as documented.** The example in the description is `await mem.initialize()` then use `mem.memory`, but `initialize()` only awaits discovery readiness, nothing issues `FT.CREATE`. The first `recall`/`remember` then hits a non-existent `:mem:idx`. This is the same gap raised on #249, and the answer there was #261 (`MemoryStore.ensureIndex()` wired into `initialize()`), which I confirmed does the right thing. The problem is ordering: #261 is based on `phase11-integration`, downstream of this PR, so as it stands phase10's `initialize()` advertises a readiness it does not deliver until #261 lands and is integrated. Please either pull `ensureIndex()` into this `initialize()` or note clearly in the PR that the facade is not usable until #261.

**2. The `initialize()` collision comment is inaccurate after #255.** It says "the memory tier already propagates [the collision], so the cache side isn't swallowed," but post-#255 the memory tier warns rather than throwing, and `ensureDiscoveryReady()` swallows everything via `.catch(() => undefined)`. So only the cache side surfaces a collision; the memory side never does. The net behavior is fine, but the rationale in the comment is backwards and worth correcting.

**3. `options.client as unknown as MemoryStoreClient` assumes the shared client supports `.call`.** `AgentCacheOptions.client`'s type does not guarantee the `.call` contract `MemoryStore` relies on, so a method-only client or mock would compile and then break at runtime. Real ioredis-like clients are fine, but the double-cast hides the assumption, worth at least a comment.

jamby77:
@KIvanow Addressed in c0efcef: added a comment on the `options.client as unknown as MemoryStoreClient` double-cast documenting that it asserts the `.call` contract MemoryStore needs (real ioredis/iovalkey clients satisfy it; a method-only client would compile but break at runtime).

KIvanow:
@jamby77 thanks, issue 3 is resolved. The cast comment landed (the c0efcef link is orphaned after the branch force-update, but the change survived the rebase as edb1937, AgentMemory.ts:50). Keeping the request-changes open for the other two:

1. initialize() still does not bootstrap the FT index. It only awaits the two ensureDiscoveryReady() calls, so a freshly initialized facade still hits a non-existent :mem:idx on the first remember/recall. ensureIndex() only lands in #261, which is downstream of this PR. Simplest fix is to pull ensureIndex() into this initialize() so the facade is usable end-to-end on its own; otherwise please note the limitation in the PR.

2. The collision comment is still backwards after #255. It says "the memory tier already propagates, so the cache side isn't swallowed," but post-#255 the memory tier warns instead of throwing and ensureDiscoveryReady() swallows via .catch(() => undefined), so only the cache side surfaces a collision. The net behavior is fine, but the rationale should be corrected.

jamby77:
Addressed review feedback:

- **#1 (initialize() FT-index bootstrap):** taking the documented option — the facade's index bootstrap lands in #261 (`MemoryStore.ensureIndex()` wired into `initialize()`), which sits directly atop this PR in the chain (#258 → #259 → #261). Merging the chain down delivers it; the facade is not usable end-to-end until #261 lands, as noted.
- **#2 (collision comment):** corrected — the memory tier warns and continues last-writer-wins, so only the cache tier surfaces a discovery collision (9ee9021).
- **#3 (client cast):** documented at the cast site.
- **Bugbot High (dual registry/heartbeat):** memory discovery namespaces both marker field and heartbeat key under `{name}:mem`, distinct from agent-cache.

Rebased onto current master.

KIvanow (approved):

PR #259, opened 2026-06-17
test(agent-memory): Phase 11 — integration suite (real valkey-search)

jamby77:
Stacked on #258 (Phase 10 — AgentMemory facade). **Final core phase (0–11).**

## What
`MemoryStore.integration.test.ts` — a skip-guarded suite that exercises the full memory stack against a real Valkey + `valkey-search` server.

- Connects via `VALKEY_URL` (default `redis://localhost:6380`), `lazyConnect` + no-retry so it **fails fast and skips** when no server is present (CI stays green) — same pattern as semantic-cache/agent-cache.
- `FT.CREATE`s the `{name}:mem:idx` index (FLAT, FLOAT32, COSINE, dim 16; TAG scope/tags/source; NUMERIC importance/created_at/last_accessed_at/access_count; TEXT content) in `beforeAll`, since the package has no index-creation code of its own.

## Coverage
1. **remember → recall round trip** — stored memory comes back at cosine distance `< 0.01` (≈ exact match), with the right id and content.
2. **scope/tag isolation** — same text in two namespaces; recall filters server-side to the requested namespace and to a tag.
3. **capacity eviction** — `maxItemsPerScope: 3`, write 5; the `evictions` counter in `__mem_stats` increments and the scope stays bounded.
4. **ttl expiry** — a `ttl: 1` memory exists, then is gone after expiry.
5. **consolidation** — old low-importance memories collapse into a `source: 'summary'` memory that is itself recallable; sources deleted; counts correct.

## Robustness
- Polls for async HNSW indexing (`pollUntil`) rather than fixed sleeps; drops the index and deletes keys around the run.
- Adds `iovalkey` as a devDependency for the test client (lockfile updated).

## Verification
Default `vitest run` → 100/100 green (integration **skips**, no server). Against the live `valkey/valkey-bundle:9.1` on port 6384: **5/5 green, non-flaky across 3 consecutive runs.** `tsc` clean · prettier clean.

This completes the agent-memory core (Phases 0–11). Optional follow-ups remain: Phase 12 (Python parity) and Phase 13 (MCP tools).

---

> [!NOTE]
> **Low Risk**
> Test-only and lockfile changes; production MemoryStore code is untouched.
> 
> **Overview**
> Adds **`MemoryStore.integration.test.ts`**, which runs `MemoryStore` end-to-end against a real Valkey + **valkey-search** instance when `VALKEY_URL` is reachable (default `redis://localhost:6380`); otherwise tests **no-op** so CI stays green, matching **agent-cache** / **semantic-cache**.
> 
> The suite **creates and tears down** its own `FT.CREATE` index (the package does not ship index bootstrap), uses **`pollUntil`** for indexing lag, and covers **remember/recall** (near-zero cosine distance), **namespace/tag** isolation, **`maxItemsPerScope`** eviction + stats, **TTL** expiry, and **consolidate** → recallable summary memories.
> 
> Adds **`iovalkey`** as a **devDependency** (lockfile updated); no runtime/package API changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit df18e145f05d9d7dacdf00fbea4dc3d0cc6e7f93. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/agent-memory/package.json
+++ b/packages/agent-memory/package.json
@@ -44,6 +44,7 @@
   "devDependencies": {
     "@opentelemetry/sdk-trace-base": "^1.30.1",
     "@types/node": "^22.19.15",
+    "iovalkey": ">=0.3.0",
     "typescript": "^5.9.3",
     "vitest": "^4.1.1"
   },
--- a/packages/agent-memory/src/__tests__/MemoryStore.integration.test.ts
+++ b/packages/agent-memory/src/__tests__/MemoryStore.integration.test.ts
@@ -165,7 +165,8 @@ describe('MemoryStore integration (real valkey-search)', () => {
 
     const evictions = Number(await client.call('HGET', `${NAME}:__mem_stats`, 'evictions'));
     expect(evictions).toBeGreaterThanOrEqual(1);
-    await pollUntil(async () => (await ftCount('@namespace:{cap}')) <= 4);
+    await pollUntil(async () => (await ftCount('@namespace:{cap}')) <= 3);
+    expect(await ftCount('@namespace:{cap}')).toBeLessThanOrEqual(3);
   });
 
   it('expires a ttl-scoped memory', async () => {
--- a/packages/agent-memory/src/__tests__/MemoryStore.integration.test.ts
+++ b/packages/agent-memory/src/__tests__/MemoryStore.integration.test.ts
@@ -1,0 +1,199 @@
+import { describe, it, expect, beforeAll, afterAll, vi } from 'vitest';
+import Valkey from 'iovalkey';
+import { MemoryStore } from '../MemoryStore';
+import type { MemoryStoreClient } from '../types';
+import { fakeEmbed } from './helpers/fakeEmbed';
+
+const VALKEY_URL = process.env.VALKEY_URL ?? 'redis://localhost:6380';
+const NAME = 'agentmem_it';
+const INDEX = `${NAME}:mem:idx`;
+const DIMS = 16;
+
+let client: Valkey;
+let store: MemoryStore;
+let skip = false;
+
+function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+async function pollUntil(fn: () => Promise<boolean>, timeoutMs = 5000): Promise<void> {
+  const start = Date.now();
+  while (Date.now() - start < timeoutMs) {
+    if (await fn()) {
+      return;
+    }
+    await sleep(100);
+  }
+  throw new Error('pollUntil timed out');
+}
+
+async function ftCount(filter: string): Promise<number> {
+  const raw = await client.call('FT.SEARCH', INDEX, filter, 'LIMIT', '0', '0', 'DIALECT', '2');
+  return Array.isArray(raw) ? Number(raw[0]) : 0;
+}
+
+async function dropAndClean(): Promise<void> {
+  try {
+    await client.call('FT.DROPINDEX', INDEX);
+  } catch {
+    // index may not exist yet
+  }
+  const keys = await client.keys(`${NAME}:*`);
+  if (keys.length > 0) {
+    await client.del(...keys);
+  }
+}
+
+async function createIndex(): Promise<void> {
+  await client.call(
+    'FT.CREATE',
+    INDEX,
+    'ON',
+    'HASH',
+    'PREFIX',
+    '1',
+    `${NAME}:mem:`,
+    'SCHEMA',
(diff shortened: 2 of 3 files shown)


amitkojha05:
Hey @jamby77  👋

Before I kick off Phase 13 (MCP tools for `@betterdb/agent-memory`), wanted to quickly align on a few things to avoid any rework:

**1. Base branch**
Is Phase 11 (#259) merged to `master` yet, or should I branch off `feature/agent-memory-phase11-integration` and rebase later? Don't want to stack on something still in review.

**2. Discovery marker type string**
What's the exact `type` value used in the Phase 7 discovery marker — is it `'agent_memory'` (with underscore) or something else? Want to confirm before I filter on it in the service layer.

**3. `memory_recall` via MCP**
The recall tool needs an embed function at the API layer, which currently isn't passed through from the application. My plan for this PR is to register the tool but have it return a documented `501 Not Implemented` — something like: *"Server-side recall requires the embed function to be available at the API layer, which isn't currently supported through MCP. Use the agent-memory library directly for now."* The interface exists for discovery; the embed-pass-through can be a follow-up (Phase 13b). Does that approach work for everyone, or is there a preferred alternative?

**4. Stats hash field names**
For the `__mem_stats` hash — can someone confirm the exact field names (e.g. `recall_total`, `recall_hits`, `evictions_total`, etc.)? I want to make sure the service reads the right keys rather than assuming from the Phase 9 telemetry counter names.

**5. `escapeTag` utility**
Planning to use `escapeTag()` from `@betterdb/valkey-search-kit` for scope/tag values in `FT.SEARCH` filters. Is this the right import, or has that been moved somewhere?

**6. MCP tool count in existing tests**
How many tools are currently registered in `packages/mcp/src/index.ts`? I'll be adding 6 new ones and need to update the count assertion in the test file.

Happy to work as per your response . Otherwise, answers here are great — this should be a pretty contained PR once the above is clear.

Thanks!

KIvanow (changes requested):
Test-only phase, and the distinct-namespace design is a real improvement over the retrieval integration suite: every test uses its own scope prefix (rt / tenantA / tenantB / cap / ttl / cons), so the cases are genuinely isolated and reorder-safe rather than sharing a single mutated index. pollUntil, the FLAT index for determinism, and the dropAndClean teardown are all sensible.

One blocking issue plus two minors.

**1 (blocking): the skip guard only pings, and the default URL points at a non-search Valkey.**
The guard does connect + ping and nothing else, and the default is redis://localhost:6380 (no auth). That default does not match the search-capable instance (redis://:devpassword@localhost:6384). The failure mode is the bad one: if any plain, non-search Valkey is listening on 6380, ping succeeds, the skip stays false, and then beforeAll's FT.CREATE throws because the module is not loaded. The suite ERRORS instead of skipping. The retrieval phase6 suite already guards this with an FT._LIST probe; please add the same probe here (skip when FT.* is unavailable) and align the default URL with the search-capable instance.

**2 (minor): the hand-rolled FLAT index diverges from the production schema.**
The test builds its own index args instead of using the package's index builder (buildMemoryIndexArgs, landing in the ensure-index follow-up). FLAT is fine for test determinism, but once that builder is integrated the suite should dogfood it so the test exercises the real schema (HNSW, field set) rather than a parallel definition that can drift.

**3 (minor): the iovalkey 'error' handler is only attached on the skip path.**
On the happy path there is no error listener, so a mid-run connection error surfaces as an unhandled rejection. Attach it unconditionally.

jamby77:
@KIvanow Fixed in 71793fa: the integration client's `'error'` listener is now attached unconditionally right after construction, so a mid-run connection drop on the happy path can't surface as an unhandled rejection.

KIvanow (approved):
Verified on 7c2413d.

Issue 3 is resolved: the iovalkey 'error' listener is now attached unconditionally right after construction, so a mid-run drop on the happy path can't surface as an unhandled rejection.

Issue 2 (hand-rolled FLAT index) is effectively handled by the stack: #261 swaps this createIndex() for store.ensureIndex(), so the suite dogfoods the production builder once that lands. No need to duplicate it here.

On issue 1, connect + ping is fine for the MVP. If someone points VALKEY_URL at a non-search Valkey, beforeAll's FT.CREATE fails fast with a descriptive module-not-loaded error rather than a silent or misleading failure, so the misconfiguration is obvious from the test output. An FT._LIST probe to convert that into a skip is a nice-to-have we can layer on later, and aligning the default URL with the search-capable instance can ride along then.

Approving.

PR #269, opened 2026-06-23
feat(retrieval-py): Python port of @betterdb/retrieval

KIvanow:
## Summary

Python port of `@betterdb/retrieval`: a developer-facing retrieval SDK over [Valkey Search](https://valkey.io/topics/search/) (`FT.*`). New package `packages/retrieval-py` (`betterdb-retrieval` 0.1.0), an exact functional equivalent of the TypeScript package, built on `betterdb-valkey-search-kit`. Third step of the TS to Python port (after valkey-search-kit-py and the semantic-cache-py refactor).

## Changes
- `Retriever`: idempotent index lifecycle (`create_index` / `drop_index` / `describe_index` / `health`), `upsert` / `delete`, and `query` (text|vector + filter + `hybrid="rerank"` KNN search), plus discovery `register` / `unregister`.
- Typed `RetrievalSchema` (tag / numeric / text fields; HNSW & FLAT vector specs), `ft_create` / `ft_search` arg builders, version-skew-tolerant health parsing.
- Shared discovery registry marker on `__betterdb:caches`, ownership-checked so it never clobbers a foreign cache type.
- Observability seams: `RetrievalMetrics` / `RetrievalTracer` protocols + `create_prometheus_metrics` (optional `prometheus` extra).
- Python idiom adaptations: client `Protocol` uses `execute_command(*args)`; errors are `ValueError`; `query` is keyword-only; results are snake_case dataclasses; schema `TypedDict`s keep camelCase keys to mirror the TS shape.
- 96 unit tests (against a fake client) mirroring the TS vitest suites; README + CHANGELOG. `uv.lock` left untracked, matching the other `-py` packages.

## Checklist
- [x] Unit / integration tests added — 96 passing (`uv run --extra dev pytest tests -q`); ruff check + format clean. Live-server integration test not ported (TS `Retriever.integration.test.ts`).
- [x] Docs added / updated — README + CHANGELOG

---

> [!NOTE]
> **Medium Risk**
> Large new surface area for vector/index operations on Valkey; TS discovery and index-creation behavior changes affect multi-tenant registry safety and startup races.
> 
> **Overview**
> Adds **`packages/retrieval-py`** (`betterdb-retrieval` 0.1.0): an async Python SDK mirroring `@betterdb/retrieval` on Valkey Search—`Retriever` (index lifecycle, upsert/delete, KNN query with filters and `hybrid="rerank"`), typed `RetrievalSchema`, `__betterdb:caches` discovery, health parsing, and optional Prometheus metrics/tracing hooks, plus README, CHANGELOG, and a large pytest suite against a fake client.
> 
> The TypeScript **`@betterdb/retrieval`** package picks up the same behavioral fixes in this PR: **`register` / `unregister`** now use **Lua `EVAL` scripts** for atomic ownership checks on the shared registry (no raw `HSET`/`HDEL` race with other cache types), **`createIndex`** **ignores “index already exists”** after a not-found `FT.INFO` (multi-worker boot), and **numeric query filters** reject **non-finite** values (`NaN`/`±Infinity`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51b21773530da3bd258673bbc9cb78ef2a4b8dd6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/retrieval-py/betterdb_retrieval/ft_search.py
+++ b/packages/retrieval-py/betterdb_retrieval/ft_search.py
@@ -1,0 +1,45 @@
+from __future__ import annotations
+
+from betterdb_valkey_search_kit import escape_tag
+
+from .fields import SCORE_FIELD
+from .ft_create import resolve_vector_field_name
+from .schema import RetrievalSchema
+
+QueryFilter = dict[str, "str | int | float"]
+
+
+def _is_number(value: object) -> bool:
+    return isinstance(value, (int, float)) and not isinstance(value, bool)
+
+
+def _build_filter_clause(field: str, value: str | int | float, schema: RetrievalSchema) -> str:
+    spec = schema["fields"].get(field)
+    if spec is None:
+        raise ValueError(f"Cannot filter on unknown field '{field}'")
+    field_type = spec.get("type")
+    if field_type == "tag":
+        return f"@{field}:{{{escape_tag(str(value))}}}"
+    if field_type == "numeric":
+        if not _is_number(value):
+            raise ValueError(
+                f"Numeric filter on field '{field}' requires a number, got: {type(value).__name__}"
+            )
+        return f"@{field}:[{value} {value}]"
+    raise ValueError(
+        f"Cannot filter on TEXT field '{field}'; only tag and numeric fields are filterable"
+    )
+
+
+def build_ft_search_query(
+    schema: RetrievalSchema,
+    k: int,
+    filter: QueryFilter | None = None,
+) -> str:
+    vector_field = resolve_vector_field_name(schema["vector"])
+    clauses: list[str] = []
+    if filter is not None:
+        for field, value in filter.items():
+            clauses.append(_build_filter_clause(field, value, schema))
+    filter_expr = f"({' '.join(clauses)})" if clauses else "*"
+    return f"{filter_expr}=>[KNN {k} @{vector_field} $vec AS {SCORE_FIELD}]"
--- a/packages/retrieval-py/betterdb_retrieval/ft_search.py
+++ b/packages/retrieval-py/betterdb_retrieval/ft_search.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import math
+
 from betterdb_valkey_search_kit import escape_tag
 
 from .fields import SCORE_FIELD
@@ -24,6 +26,10 @@ def _build_filter_clause(field: str, value: str | int | float, schema: Retrieval
         if not _is_number(value):
             raise ValueError(
                 f"Numeric filter on field '{field}' requires a number, got: {type(value).__name__}"
+            )
+        if isinstance(value, float) and not math.isfinite(value):
+            raise ValueError(
+                f"Numeric filter on field '{field}' requires a finite number, got: {value!r}"
             )
         return f"@{field}:[{value} {value}]"
     raise ValueError(
--- a/packages/retrieval-py/betterdb_retrieval/retriever.py
+++ b/packages/retrieval-py/betterdb_retrieval/retriever.py
@@ -38,6 +38,33 @@ _logger = logging.getLogger(__name__)
 
 EmbedFn = Callable[[str], Awaitable[list[float]]]
 
+# Atomic compare-and-set for the shared registry field. REGISTRY_KEY is keyed by
+# name and shared with agent-cache, so a plain HGET -> compare -> HSET/HDEL has a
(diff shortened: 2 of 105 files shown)


packages/retrieval-py/betterdb_retrieval/ft_search.py
```diff
@@ -0,0 +1,45 @@
...
+    if spec is None:
+        raise ValueError(f"Cannot filter on unknown field '{field}'")
+    field_type = spec.get("type")
+    if field_type == "tag":
+        return f"@{field}:{{{escape_tag(str(value))}}}"
+    if field_type == "numeric":
+        if not _is_number(value):
+            raise ValueError(
+                f"Numeric filter on field '{field}' requires a number, got: {type(value).__name__}"
+            )
```
jamby77, line 29: **Numeric filter accepts `nan`/`inf`, producing a malformed FT.SEARCH query**

`_is_number` only excludes `bool`, so `float('nan')`, `float('inf')`, and `float('-inf')` all pass the guard and interpolate raw into the range clause — e.g. `@year:[nan nan]` / `@year:[inf inf]`. That's an invalid query rejected (or mishandled) at runtime by Valkey Search rather than at the SDK boundary. It's the one filter path where a caller value reaches the query string unnormalized (the tag path is correctly escaped via `escape_tag`).

Reject non-finite values up front (requires `import math` at the top of the file):

```suggestion
        if not _is_number(value):
            raise ValueError(
                f"Numeric filter on field '{field}' requires a number, got: {type(value).__name__}"
            )
        if isinstance(value, float) and not math.isfinite(value):
            raise ValueError(
                f"Numeric filter on field '{field}' requires a finite number, got: {value!r}"
            )
```

packages/retrieval-py/betterdb_retrieval/retriever.py
```diff
@@ -0,0 +1,385 @@
...
+            if not is_index_not_found_error(err):
+                raise
+        dims = await self._resolve_dims()
+        schema: RetrievalSchema = {
+            "fields": self._schema["fields"],
+            "vector": {**self._schema["vector"], "dims": dims},
+        }
+        await self._client.execute_command(
+            "FT.CREATE", *build_ft_create_args(self._name, schema, self._capabilities)
+        )
```
jamby77: **`create_index` idempotency breaks under a creation race**

The `FT.INFO` probe → `FT.CREATE` is check-then-act. If another worker creates the index in between (common on multi-worker boot), `FT.CREATE` raises "Index already exists", which isn't caught here — `is_index_not_found_error` only matches not-found. The documented idempotent contract then fails under concurrency. Swallow the already-exists case:

```suggestion
        try:
            await self._client.execute_command(
                "FT.CREATE", *build_ft_create_args(self._name, schema, self._capabilities)
            )
        except Exception as err:
            if "already exists" not in str(err).lower():
                raise
```

Cleaner still would be an `is_index_already_exists_error` helper in the kit, mirroring `is_index_not_found_error`.

packages/retrieval-py/betterdb_retrieval/retriever.py
```diff
@@ -0,0 +1,385 @@
...
+                self._name,
+                existing_type,
+            )
+            return
+        marker = build_retrieval_marker(
+            name=self._name,
+            version=RETRIEVAL_VERSION,
+            started_at=datetime.now(timezone.utc).isoformat(),
+        )
+        await self._client.execute_command("HSET", REGISTRY_KEY, self._name, json.dumps(marker))
```
jamby77: **Discovery register/unregister is a non-atomic read-modify-write on a shared field**

The ownership check is correct in intent, but there's a TOCTOU window between `_read_registry_marker()` (HGET) and the `HSET` here (and between HGET and `HDEL` in `unregister`). Since `REGISTRY_KEY` is explicitly shared with agent-cache, a foreign marker written into that window gets clobbered — the exact case the check is meant to prevent.

Make it atomic: a small `EVAL` Lua script that does HGET → compare `type` → HSET/HDEL in one round trip (or `WATCH`/`MULTI`/`EXEC`). Low urgency since the single-owner path is fine and the worst case is a benign marker overwrite, not data loss.

KIvanow:
@jamby77 all issues should be addressed now. Some were applicable for the TS packages too, so they are ported in a specific commit https://github.com/BetterDB-inc/monitor/pull/269/changes/cf0e147a884c7c988477e9f81ce2816d2b278327

jamby77 (approved):

PR #278, opened 2026-06-24
feat(charts): valkey-search Helm chart for self-host and cloud tenants

KIvanow:
Single chart for both the self-host (marketing) profile and cloud per-tenant provisioning. Ships Valkey with the search module (FT.*), an ACL-hardened named user loaded from an aclfile at boot (survives restarts), optional AOF persistence + maxmemory, and an optional Traefik IngressRouteTCP HostSNI route for public TLS exposure.

The container runs the bundle image via args (not command) so its entrypoint auto-loads the search module.

## Changes
- 

## Checklist
- [ ] Docs added / updated

---

> [!NOTE]
> **Medium Risk**
> New deploy surface for credentials, ACLs, and optional public TLS routing; misconfiguration (GitOps password rotation, external Secret ACLs) could break clients or expose Valkey incorrectly.
> 
> **Overview**
> Adds a new **`charts/valkey-search`** Helm chart that deploys **Valkey with the search module** (`valkey/valkey-bundle`) as a single-replica StatefulSet, ClusterIP Service, and install notes.
> 
> **Self-host defaults** use chart-managed auth: a Secret with `password` and boot-loaded `users.acl` (default user off, optional `-@dangerous` hardening with `+info` for monitoring), stable passwords on upgrade via Helm `lookup`, and whitespace validation on passwords. **Cloud tenants** can point at `auth.existingSecret` and layer **`values-cloud-tenant.yaml`** for persistence, `maxmemory`, resources, and public exposure.
> 
> The workload starts **`valkey-server` via `args`** so the bundle entrypoint still loads the search module; readiness checks **`FT.SEARCH`** registration, not just PING. Optional **Traefik `IngressRouteTCP`** registers a per-tenant **HostSNI** route when `exposure.public` and `exposure.sniRoute` are enabled. Pod rollout annotations hash chart-rendered or external Secrets so credential/ACL changes trigger restarts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3877edacb37353ee7a2f195235b548c4a35aa085. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/charts/valkey-search/templates/NOTES.txt
+++ b/charts/valkey-search/templates/NOTES.txt
@@ -22,7 +22,12 @@ This instance is NOT publicly exposed. Port-forward to reach it locally:
   kubectl port-forward -n {{ .Release.Namespace }} svc/{{ include "valkey-search.fullname" . }} 6379:{{ .Values.service.port }}
 {{- end }}
 
+{{- if .Values.auth.existingSecret }}
+ACLs come from the externally-owned Secret "{{ .Values.auth.existingSecret }}";
+the chart's hardening flag does not control runtime permissions here.
+{{- else }}
 The "default" user is disabled; only "{{ .Values.auth.username }}" can connect.
 {{- if .Values.hardening.disableDangerousCommands }}
 Dangerous commands (FLUSHALL, CONFIG, DEBUG, ...) are blocked; FT.* search and INFO work.
+{{- end }}
 {{- end }}
--- a/charts/valkey-search/templates/NOTES.txt
+++ b/charts/valkey-search/templates/NOTES.txt
@@ -1,0 +1,28 @@
+Valkey + search ({{ .Chart.AppVersion }}) is installing as {{ include "valkey-search.fullname" . }}.
+
+Connect from inside the cluster as user "{{ .Values.auth.username }}":
+
+  Host: {{ include "valkey-search.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local
+  Port: {{ .Values.service.port }}
+
+Fetch the password:
+
+  kubectl get secret -n {{ .Release.Namespace }} {{ include "valkey-search.secretName" . }} \
+    -o jsonpath='{.data.password}' | base64 -d
+
+{{- if and .Values.exposure.public .Values.exposure.sniRoute.enabled }}
+
+Public TLS endpoint (via the shared SNI proxy):
+
+  redis+tls://{{ .Values.auth.username }}:<password>@{{ .Values.exposure.host }}:443
+{{- else }}
+
+This instance is NOT publicly exposed. Port-forward to reach it locally:
+
+  kubectl port-forward -n {{ .Release.Namespace }} svc/{{ include "valkey-search.fullname" . }} 6379:{{ .Values.service.port }}
+{{- end }}
+
+The "default" user is disabled; only "{{ .Values.auth.username }}" can connect.
+{{- if .Values.hardening.disableDangerousCommands }}
+Dangerous commands (FLUSHALL, CONFIG, DEBUG, ...) are blocked; FT.* search works.
+{{- end }}
--- a/charts/valkey-search/templates/NOTES.txt
+++ b/charts/valkey-search/templates/NOTES.txt
@@ -24,5 +24,5 @@ This instance is NOT publicly exposed. Port-forward to reach it locally:
 
 The "default" user is disabled; only "{{ .Values.auth.username }}" can connect.
 {{- if .Values.hardening.disableDangerousCommands }}
-Dangerous commands (FLUSHALL, CONFIG, DEBUG, ...) are blocked; FT.* search works.
+Dangerous commands (FLUSHALL, CONFIG, DEBUG, ...) are blocked; FT.* search and INFO work.
 {{- end }}
--- a/charts/valkey-search/templates/secret.yaml
+++ b/charts/valkey-search/templates/secret.yaml
@@ -20,16 +20,25 @@
 {{- $password = randAlphaNum 32 }}
 {{- end }}
 {{- end }}
+{{- /*
+  A bring-your-own auth.password is written verbatim into the space-delimited
+  aclfile; whitespace would split into stray ACL tokens and CrashLoop the pod.
+  Generated/looked-up passwords are safe, so only guard the explicit value.
+*/}}
+{{- if and .Values.auth.password (regexMatch "\\s" .Values.auth.password) }}
+{{- fail "auth.password must not contain whitespace (it corrupts the space-delimited ACL file)" }}
+{{- end }}
 {{- $username := .Values.auth.username }}
 {{- /*
   users.acl is loaded at boot via --aclfile so the user survives restarts
   (runtime ACL SETUSER would not). default is left OFF so the only way in is
   the named user. -@dangerous strips FLUSHALL/CONFIG/DEBUG/etc while FT.*
-  (search) keeps working.
+  (search) keeps working. INFO lives in @dangerous on valkey-bundle, so it is
+  re-granted (+info) — BetterDB Monitor's metric collection needs it.
 */}}
 {{- $rules := "~* &* +@all" }}
 {{- if .Values.hardening.disableDangerousCommands }}
-{{- $rules = "~* &* +@all -@dangerous" }}
+{{- $rules = "~* &* +@all -@dangerous +info" }}
 {{- end }}
 {{- $acl := printf "user default off\nuser %s on >%s %s\n" $username $password $rules }}
 apiVersion: v1
--- a/charts/valkey-search/templates/secret.yaml
+++ b/charts/valkey-search/templates/secret.yaml
@@ -1,0 +1,45 @@
+{{- if not .Values.auth.existingSecret }}
+{{- /*
+  Chart-managed credential (self-host path). The cloud path sets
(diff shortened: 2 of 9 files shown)


charts/valkey-search/templates/secret.yaml
```diff
@@ -0,0 +1,45 @@
...
+{{- $username := .Values.auth.username }}
+{{- /*
+  users.acl is loaded at boot via --aclfile so the user survives restarts
+  (runtime ACL SETUSER would not). default is left OFF so the only way in is
+  the named user. -@dangerous strips FLUSHALL/CONFIG/DEBUG/etc while FT.*
+  (search) keeps working.
+*/}}
+{{- $rules := "~* &* +@all" }}
+{{- if .Values.hardening.disableDangerousCommands }}
+{{- $rules = "~* &* +@all -@dangerous" }}
```
jamby77: **`-@dangerous` blocks `INFO`.** Verified on valkey-bundle 9.1: `INFO` is in `@dangerous` (alongside `DEBUG`/`FLUSHALL`), so `+@all -@dangerous` denies it. The BetterDB Monitor's metric collection — and common library health/version checks — then get `NOPERM` on every instance provisioned with the default hardening. `CONFIG`/`SLOWLOG`/`CLIENT` are *not* in `@dangerous`, so only `INFO` regresses; easy to grant back:

```suggestion
{{- $rules = "~* &* +@all -@dangerous +info" }}
```

Re-grant any other `@dangerous` command a client needs (check with `ACL CAT dangerous`).

charts/valkey-search/templates/secret.yaml
```diff
@@ -0,0 +1,45 @@
...
+{{- end }}
+{{- $acl := printf "user default off\nuser %s on >%s %s\n" $username $password $rules }}
```
jamby77, line 44: **A bring-your-own `auth.password` with whitespace/special chars corrupts the aclfile → CrashLoop.** ACL rules are space-delimited, so `auth.password: "p@ss w0rd"` renders `user betterdb on >p@ss w0rd ~* &* +@all` — `w0rd…` becomes stray ACL tokens, the aclfile fails to load, and Valkey CrashLoopBackOffs. Generated `randAlphaNum` passwords are safe, but the documented BYO value isn't validated. A guard would catch it early, e.g. `{{- if regexMatch "\\s" $password }}{{- fail "auth.password must not contain whitespace" }}{{- end }}`.

charts/valkey-search/templates/secret.yaml
```diff
@@ -0,0 +1,45 @@
...
+    3. A freshly generated 32-char password on first install.
+*/}}
+{{- $secretName := printf "%s-auth" (include "valkey-search.fullname" .) }}
+{{- $existing := (lookup "v1" "Secret" .Release.Namespace $secretName) }}
+{{- $password := .Values.auth.password }}
+{{- if not $password }}
+{{- if and $existing $existing.data (index $existing.data "password") }}
+{{- $password = index $existing.data "password" | b64dec }}
+{{- else }}
+{{- $password = randAlphaNum 32 }}
```
jamby77, line 20: **The `lookup`-based password reuse silently rotates the password under GitOps.** The "never rotates out from under connected clients" guarantee depends on `lookup` reading the live Secret. Argo CD / Flux render with `helm template` (no cluster access), so `lookup` returns empty → `randAlphaNum 32` mints a **new** password on every sync, rotating the Secret + ACL and breaking connected clients. Worth documenting that GitOps installs must set `auth.existingSecret` (or a fixed `auth.password`), since the reuse path can't work without a live cluster read.

charts/valkey-search/values-cloud-tenant.yaml
```diff
@@ -0,0 +1,32 @@
...
+  # Entitlement pre-creates this Secret (keys: password, users.acl) with a strong
+  # password, so credential ownership stays out of Helm entirely.
+  existingSecret: valkey-acme-auth
+
+persistence:
+  enabled: true
+  size: 2Gi # ~2.5x maxmemory: holds RDB + AOF + rewrite headroom
+
+valkey:
+  maxmemory: 768mb # ~75% of the 1Gi limit
```
jamby77: **`maxmemory` at ~75% of the pod memory limit risks OOMKill during AOF/RDB rewrite.** With `maxmemory: 768mb`, `limits.memory: 1Gi`, and `appendonly: true`, a `BGREWRITEAOF`/RDB-save fork's copy-on-write pages plus AOF rewrite buffers under write load can push RSS past 1Gi → OOMKilled mid-rewrite. ~256Mi of headroom is tight for a forking persistence model; 50–60% of the limit is the usual guidance once persistence is on.

charts/valkey-search/templates/statefulset.yaml
```diff
@@ -0,0 +1,124 @@
...
+                - -c
+                - 'valkey-cli --user {{ .Values.auth.username }} --pass "$VALKEY_PASSWORD" ping'
+            initialDelaySeconds: 10
+            periodSeconds: 10
+          readinessProbe:
+            exec:
+              command:
+                - sh
+                - -c
+                - 'valkey-cli --user {{ .Values.auth.username }} --pass "$VALKEY_PASSWORD" ping'
```
jamby77: **Readiness only `ping`s — it never verifies the search module actually loaded.** If the module fails to load but `valkey-server` stays up (FT.* unknown, not a crash), the pod reports Ready while the chart's entire purpose is broken. A readiness probe that runs `FT._LIST` (or `COMMAND DOCS FT.SEARCH`) would catch a silent module-load failure. The `:latest` double-load case is already covered since it crashes, but a non-crashing miss isn't.

charts/valkey-search/templates/statefulset.yaml
```diff
@@ -0,0 +1,124 @@
...
+    matchLabels:
+      {{- include "valkey-search.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "valkey-search.selectorLabels" . | nindent 8 }}
+      annotations:
+        # Roll the pod when the credential/ACL changes so a rotated password
+        # or ACL rule takes effect (the file is mounted, loaded at boot).
+        checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
```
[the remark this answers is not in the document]
KIvanow: Fixed in b8be535: when `existingSecret` is set, `checksum/secret` now hashes the live Secret's `.data` via `lookup`, so pods roll on rotation.

charts/valkey-search/templates/secret.yaml
KIvanow: Good catch, fixed in b8be535 — ACL is now `+@all -@dangerous +info`.

charts/valkey-search/templates/secret.yaml
KIvanow: Fixed in b8be535: added a `fail` guard rejecting whitespace in `auth.password` at template time.

charts/valkey-search/templates/secret.yaml
KIvanow: Documented in b8be535 (`values.yaml`): GitOps installs must set `existingSecret` or a fixed `auth.password`.

charts/valkey-search/values-cloud-tenant.yaml
KIvanow: Fixed in b8be535: dropped to `512mb` (~50% of the limit).

charts/valkey-search/templates/statefulset.yaml
KIvanow: Fixed in b8be535: readiness now runs `COMMAND INFO FT.SEARCH | grep -q` to confirm the module loaded.

charts/valkey-search/templates/secret.yaml
```diff
@@ -0,0 +1,54 @@
...
```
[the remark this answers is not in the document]
KIvanow: Documented in b8be535 (`values.yaml`): GitOps installs must set `existingSecret` or a fixed `auth.password` since the lookup-reuse path needs a live cluster read.

charts/valkey-search/templates/NOTES.txt
```diff
@@ -0,0 +1,28 @@
...
+
+This instance is NOT publicly exposed. Port-forward to reach it locally:
+
+  kubectl port-forward -n {{ .Release.Namespace }} svc/{{ include "valkey-search.fullname" . }} 6379:{{ .Values.service.port }}
+{{- end }}
+
+The "default" user is disabled; only "{{ .Values.auth.username }}" can connect.
+{{- if .Values.hardening.disableDangerousCommands }}
+Dangerous commands (FLUSHALL, CONFIG, DEBUG, ...) are blocked; FT.* search and INFO work.
+{{- end }}
```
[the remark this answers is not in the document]
KIvanow: Fixed in 36b8592: under `existingSecret` the NOTES now show an external-ACL note instead of claiming the hardening flag controls runtime permissions.

jamby77 (approved):
Reviewed: clean Helm chart for Valkey + search. Defensive overall (FT.SEARCH readiness gate, stable passwords via lookup, args-not-command preserves module load). One minor: guard the lookup(...).data on the pod checksum annotation when auth.existingSecret points at a not-yet-created Secret (mirror the safe $existing pattern from secret.yaml) — non-blocking, fails loudly at render time. Approving.

PR #280, opened 2026-06-25
fix(mcp): pass section param through to info endpoint

AruneshDwivedi:
The get_info MCP tool accepts a section filter (server, clients, memory, stats, replication, keyspace) but the underlying API endpoint at /mcp/instance/:id/info never received it. All sections were always returned regardless of the requested section.

Adds @Query('section') to the McpController.getInfo handler and passes the parsed array to getInfoParsed(). The MCP tool now includes the section as a query param in the API call so the server filters before responding.

### DIFF
--- a/apps/api/src/mcp/mcp.controller.ts
+++ b/apps/api/src/mcp/mcp.controller.ts
@@ -56,10 +56,11 @@ export class McpController {
   }
 
   @Get('instance/:id/info')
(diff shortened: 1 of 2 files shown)


AruneshDwivedi:

KIvanow:
Thanks for the contribution! Clean fix — good catch that the section param was never being passed through. Nice and minimal.

KIvanow (approved):
LGTM

Issue #276, opened 2026-06-24
memory forget: duplicate-pending guard is racy (no DB-level uniqueness)


KIvanow:
Post-release follow-up from #275 (Phase 13c). Non-blocking, intentional for the initial release.

`MemoryProposalService.rejectIfDuplicatePending` does a list → JS check → create with no transactional guarantee (`proprietary/memory-proposals/memory-proposal.service.ts:233`). Two concurrent `proposeForget` calls with the same target can both pass the pre-check and both insert, since there's no DB unique index — the `isUniqueViolation` catch in `proposeForget` never fires for the SQL adapters.

- **Impact:** low — worst case is a duplicate pending row; approving one leaves the other to be rejected/expired. No data-integrity or deletion risk.
- **Why deferred:** JSON-derived unique indexes have had flaky-CI history, and the scope target (scope + sorted tags) doesn't index cleanly.
- **Possible fix:** a partial/expression unique index on `(connection_id, store_name, target_discriminator)` filtered to `status='pending'`, or move the check inside a transaction.

PR #283, opened 2026-06-27
Fail stale applying memory proposals

### LINKED ISSUE #277: memory forget: stuck 'applying' proposals never expire or recover
Post-release follow-up from #275 (Phase 13c). Non-blocking, intentional for the initial release.

The expiry sweep only touches `status='pending'` (`expireMemoryProposalsBefore`), and the apply path intentionally leaves a visible `applying` row if the process crashes between claiming the proposal and finalizing it (`memory-apply.service.ts`). There's currently no reaper or surfacing for these — a crashed mid-apply proposal stays `applying` indefinitely and is only found by manual inspection.

- **Impact:** low and rare (requires a crash in the narrow window after the claim). By design this is the safe failure mode — the row is never falsely marked `applied`.
- **Possible fix:** a sweep that flags `applying` rows older than a threshold (e.g. → `failed` with a `stale_apply` reason, or a dedicated `needs_review` surface), plus a metric/log so operators can see stuck applies.

caioribeiroclw-pixel:
## Summary
- add a storage-level sweep for memory proposals stuck in `applying` after the apply grace window
- mark stale applies as `failed` with a `stale_apply` applied_result and audit event
- cover the memory and SQLite adapters plus service-level stale/applying behavior

## Why
Issue #277 notes that a crash after `approved -> applying` can leave a visible applying row forever. This keeps that safe failure mode during the crash window, but lets the normal expiration cron surface old stuck applies as failed instead of requiring manual inspection.

## Checks
- `SKIP_DOCKER_SETUP=true pnpm test -- --runTestsByPath src/storage/adapters/__tests__/memory-proposals.spec.ts src/storage/adapters/__tests__/memory-proposals-sqlite.spec.ts` ✅
- `git diff --check` ✅

I also tried the service spec path, but this checkout currently fails before running that suite because `../../proprietary/memory-proposals/memory-apply.dispatcher.ts` cannot resolve `@betterdb/agent-memory`; leaving the service-level tests in this PR so CI with the normal workspace setup can exercise them.

Closes #277

### DIFF
--- a/apps/api/src/common/interfaces/storage-port.interface.ts
+++ b/apps/api/src/common/interfaces/storage-port.interface.ts
@@ -480,7 +480,9 @@ export interface StoragePort {
     samples: Omit<StoredCommandStatsSample, 'id' | 'connectionId'>[],
     connectionId: string,
   ): Promise<number>;
-  getCommandStatsHistory(options: CommandStatsHistoryQueryOptions): Promise<StoredCommandStatsSample[]>;
+  getCommandStatsHistory(
+    options: CommandStatsHistoryQueryOptions,
+  ): Promise<StoredCommandStatsSample[]>;
   pruneOldCommandStatsSamples(cutoffTimestamp: number, connectionId?: string): Promise<number>;
 
   // Vector Index Snapshot Methods
@@ -588,4 +590,5 @@ export interface StoragePort {
   ): Promise<StoredMemoryProposalAudit>;
   getMemoryProposalAudit(proposalId: string): Promise<StoredMemoryProposalAudit[]>;
   expireMemoryProposalsBefore(now: number): Promise<StoredMemoryProposal[]>;
+  failStaleApplyingMemoryProposalsBefore(now: number): Promise<StoredMemoryProposal[]>;
 }
--- a/apps/api/src/storage/adapters/__tests__/memory-proposals-sqlite.spec.ts
+++ b/apps/api/src/storage/adapters/__tests__/memory-proposals-sqlite.spec.ts
@@ -148,4 +148,38 @@ describe('Memory proposal storage (SQLite)', () => {
     expect((await storage.getMemoryProposal(fresh.id))?.status).toBe('pending');
     expect(await storage.expireMemoryProposalsBefore(500)).toEqual([]);
   });
+
+  it('fails stale applying proposals, returning the rows', async () => {
+    const stale = build({ proposed_at: 1, expires_at: 10_000 });
+    const fresh = build({
+      proposed_at: 1,
+      expires_at: 10_000,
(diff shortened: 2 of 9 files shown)


KIvanow:
Thanks for the contribution! Clean fix for a real edge case — the stale-apply sweep is well-scoped, all three adapters are consistent, and the test coverage is solid. Appreciate you closing out #277.

@caioribeiroclw-pixel Before we can merge, you'll need to sign our CLA. Just post a comment on this PR with exactly:

```
```

Details here: [CLA Document](https://github.com/BetterDB-inc/monitor/blob/master/CLA.md)

KIvanow (approved):
LGTM

jamby77:
@caioribeiroclw-pixel — this is approved and ready, but it is blocked on one
thing only: the CLA is unsigned, so `CLAAssistant` is the sole failing check.

To unblock it, comment exactly this on this PR:

```
```

That is the only step. Nobody else can sign on your behalf, which is why this
has been sitting since June despite KIvanow's approval on 07-07.

## Why I am asking now

I built the same fix for #277 in #416 before noticing this PR existed. That was
my miss — I planned from the issue text and did not check for an open PR. Rather
than have your approved work superseded, the plan is to land this first and
strip the overlapping half out of #416, which keeps its own separate work on
#276 (the duplicate-pending race) and a set of migration fixes.

So this is not a courtesy ping — #416 is waiting on it.

## One thing worth a look when you get to it

The two implementations time staleness differently. This PR uses `reviewed_at`;
mine added an `applying_at` column stamped at the `approved -> applying` claim.

Yours is simpler and needs no migration, and the two are equivalent today
because approve and apply happen inside one request. The reason I went the other
way is that they stop being equivalent the moment that changes — if approval is
ever decoupled from apply (a queue, a scheduled apply), a sweep measured from
`reviewed_at` starts failing work that is still running.

Not a blocker, and not a reason to change this PR. Worth recording in the code
or the issue so the assumption is visible if anyone touches that path later.

Also note this branch last ran CI in June. It reports mergeable against current
master, but a rebase before merge would be worth it — a few things in that area
have moved since, including migration ordering on existing databases.

jamby77:
Heads up: #416 was merged as-is, which includes the stale-`applying` sweep this PR implements, so this work has been superseded. Sorry — the CLA never cleared and we chose not to hold the #276 duplicate-pending fix behind it any longer.

Recording the one substantive difference for anyone reading later: this PR times staleness from `reviewed_at`, #416 from an `applying_at` column stamped at the `approved -> applying` claim. Equivalent while approve and apply happen in one request; if that is ever decoupled, `reviewed_at` would start failing work that is still running.

Thanks for the contribution, and sorry it landed this way.

PR #287, opened 2026-06-30
feat(retrieval): query-time temporal resolver (Issue 2, slim)

jamby77:
Part of the LongMemEval recall→QA epic (item 205255004). **Stack 4/4** — base `feature/longmemeval-fact-consolidation`.

`resolveTemporal`: as-of/'before X' filter (with empty-result safety) + newest-first recency ordering, behind the `temporal` lever in the reader-context path; recall unchanged. Scoped down because Issue 3's `reconcile` already supersedes and the assembler already orders chronologically. 4 tests.

QA lift target (+8pt temporal-reasoning) pending a real Tier-2 eval.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the LongMemEval eval harness behind an optional lever; production retrieval is unaffected unless separately wired.
> 
> **Overview**
> Adds an opt-in **`temporal`** lever on the LongMemEval eval runner that filters the rerank **pool** before reader contexts are built, without changing recall (still scored on the unmodified top-k).
> 
> **`resolveTemporal`** drops chunks whose `date` tag is after `record.question_date`, keeps undated chunks, preserves incoming rerank order (no recency re-sort), and falls back to the full pool if the filter would remove everything. When the lever is off, behavior is unchanged.
> 
> The QA path now builds contexts from a **`readerPool`** (temporal-filtered when enabled) for both assemble and non-assemble modes; the non-assemble branch takes the first **k** from that pool instead of the recall **`hits`** slice. Temporal work is recorded in the cost report as latency-only (no embed/LLM calls). Unit and integration tests cover filter semantics and lever wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b7fe6e74f73bf095d562ba86ed91697626927cc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/retrieval/eval/longmemeval/runner.ts
+++ b/packages/retrieval/eval/longmemeval/runner.ts
@@ -4,6 +4,7 @@ import { chunkRecord, recordIsHit } from './adapter';
 import { createHybridRerank } from './rerank';
 import { assembleContexts } from './assemble';
 import type { AssembleOptions } from './assemble';
+import { resolveTemporal } from './temporal';
 import { createCostReport } from './levers';
 import type { CostReport, LeverCostEntry, LeverName } from './levers';
 import type { ChunkMode, Embedder, Judge, LmeRecord, Reader, Store } from './types';
@@ -85,6 +86,7 @@ export async function runEval(config: RunConfig): Promise<EvalSummary> {
   const levers = config.levers ?? [];
   const costReport = config.costReport ?? createCostReport();
   const assembleOn = levers.includes('assemble');
+  const temporalOn = levers.includes('temporal');
   const qaRun = reader !== null && judge !== null;
   const useRerank = rerankPool > k;
   const fetchK = useRerank ? rerankPool : k;
@@ -168,13 +170,22 @@ export async function runEval(config: RunConfig): Promise<EvalSummary> {
         // `pool` into the reader's contexts; dedup / MMR / chronological grouping
         // apply only via the opt-in assembleOptions (env-wired in run.ts). Recall
         // above is still measured on the unmodified rerank top-k.
+        let readerPool = pool;
+        if (temporalOn) {
+          const startedAt = Date.now();
+          readerPool = resolveTemporal(pool, { asOf: record.question_date });
+          costReport.record('temporal', { latencyMs: Date.now() - startedAt });
+        }
+
         let contexts: string[];
         if (assembleOn) {
           const startedAt = Date.now();
-          contexts = assembleContexts(pool, k, config.assembleOptions ?? {});
+          contexts = assembleContexts(readerPool, k, config.assembleOptions ?? {});
           costReport.record('assemble', { latencyMs: Date.now() - startedAt });
         } else {
-          contexts = hits.map((h) => (h.fields.date ? `[${h.fields.date}] ${h.text}` : h.text));
+          contexts = readerPool
+            .slice(0, k)
+            .map((h) => (h.fields.date ? `[${h.fields.date}] ${h.text}` : h.text));
         }
         const question =
           record.question_date !== undefined && record.question_date !== ''
--- a/packages/retrieval/eval/longmemeval/temporal.ts
+++ b/packages/retrieval/eval/longmemeval/temporal.ts
@@ -1,0 +1,23 @@
+import type { QueryHit } from '../../src/index';
+
+export interface TemporalOptions {
+  asOf?: string;
+}
+
+function applyAsOf(hits: QueryHit[], asOf: string | undefined): QueryHit[] {
+  if (asOf === undefined || asOf === '') {
+    return hits;
+  }
+  const kept = hits.filter((hit) => {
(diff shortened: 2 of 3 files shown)


KIvanow:
**LongMemEval `_m` ablation result: assemble+temporal regress QA hard.**

Full 500, k=10, pool=30, same harness as the baseline. Levers on (assemble+temporal, no facts): recall@10 92.8% (unchanged), **QA 28.2% vs the 63.8% baseline (-35.6 pts)**. Per-type QA collapses everywhere, worst on multi-session (13.5%) and temporal-reasoning (21.1%).

Recall is identical because it is measured on the untouched rerank top-k, which proves the gold evidence is retrieved into the pool. The assemble/temporal stage then loses it before the reader. Root causes:

1. Ordering contract conflict: the runner pipes pool through `resolveTemporal` before `assembleContexts`. `resolveTemporal` re-sorts the whole pool by date desc, but `assembleContexts` derives MMR relevance from list position (assumes rerank order). So MMR ends up ranking by recency, not relevance.
2. MMR (lambda 0.5) drops corroborating evidence as near-duplicates. For multi-evidence questions this removes the gold chunk. MMR diversification suits result presentation, not QA evidence assembly.
3. Dedup at 0.9 containment (smaller-set denominator) discards short fact chunks subsumed by longer ones.
4. Final chronological re-sort feeds the reader date-ordered rather than relevance-first.

Suggestions: run each lever in isolation on `_m` to attribute the loss (the all-on number conflates them); if keeping assemble, derive relevance from the rerank score rather than post-temporal list position, drop or soften MMR/dedup for evidence selection, and preserve relevance-first ordering into the reader. The temporal as-of filter (drop future-dated chunks) is fine and cheap; the harm is its full re-sort feeding assemble.

Separately, the `facts` lever is impractical on `_m`: it fires one sequential gpt-5.4 chat call per haystack session, and `_m` averages ~475 sessions per record (~237k sequential calls for the full run). It was likely validated on `oracle`, where haystacks are only the evidence sessions. It needs bounded-concurrency extraction (the per-session calls are independent; only reconcile needs ordering) before it can run on `_m`.

jamby77:
Reworked to address the regression (pushed):

- **temporal** (12121bf): `resolveTemporal` is now an as-of filter only — dropped the newest-first re-sort that fed a recency-ordered pool into the assembler.
- **assemble** (8265156): default is now relevance-first and non-destructive (rerank-ordered top-k + date prefix). Dedup, MMR, and chronological grouping are behind opt-in flags, out of the default evidence path.

Net: assemble+temporal on should now sit at ≈ baseline rather than dropping/reordering evidence. Could you re-run `_m` to confirm? And agreed on isolating each lever — the flags let us A/B dedup/MMR/grouping individually.

Separately, the `facts` lever's per-session sequential extraction is untouched here; it still needs bounded-concurrency before it's practical on `_m` (tracked separately).

jamby77:
Follow-up on the `facts` scalability note: bounded-concurrency extraction is now up in #291 (stacked on this chain). Per-session extraction runs via an order-preserving `mapWithConcurrency` (default 8, `LONGMEMEVAL_FACTS_CONCURRENCY`); reconcile stays sequential and curated output is unchanged — so `facts` is now runnable on `_m`.

KIvanow (approved):
Reviewed the full PR and re-ran LongMemEval-M (full 500, k=10, pool=30, same reader/judge/embedder + live valkey-search).

**Code is clean and safe to merge.** The as-of filter is correct: it drops chunks dated after `question_date`, keeps undated chunks, falls back to the original pool if the filter would empty it, and preserves rerank order. Recall stays measured on the unmodified top-k (`pool.slice(0, k)`), so recall is invariant to the lever. With the lever off, `readerPool.slice(0, k)` is identical to the previous `hits` path, so there is no baseline regression. Tests cover the edge cases plus zero-cost and no-recall-regression, which is the right contract.

**It also resolves the earlier regression.** The prior version re-sorted survivors newest-first, which fed a recency-ordered pool into the assembler's position-based relevance and collapsed QA (we saw 28.2% on the combined levers). Scoping this down to an as-of filter only, with ordering left to rerank/assemble, brings QA back to baseline: 62.4% vs 63.8% (within noise), recall unchanged at 92.8%.

**One honest caveat: it does not improve QA on this benchmark.** LongMemEval questions are asked at the end of each timeline, so almost nothing is dated after `question_date` and the filter has little to remove (measured cost was +12ms with effectively no pool change). The +8pt temporal-reasoning target in the description is unmet in practice. That is a property of the benchmark structure, not a code defect. The real QA bottleneck is downstream reasoning (multi-session and temporal-reasoning lag despite 90%+ recall), which a pool filter cannot address.

Net: this is defensible, well-tested, flag-gated scaffolding that fixes the regression and adds no risk, so merging is fine. Just flagging that the temporal QA lift will need to come from the evidence/reasoning layer, not the as-of filter.

PR #289, opened 2026-07-01
feat(retrieval): cross-encoder reranker lever (Issue 5)

jamby77:
Part of the LongMemEval recall→QA epic (item 205255004). **Stack 5/5** — base `feature/longmemeval-temporal-resolver`.

`createCrossEncoderRerank` implements the existing `RerankFn` seam, reordering the over-fetch pool by a batch scorer (one scoring call per query). Behind the `rerank-cross` lever: when enabled and `rerankPool>k`, the pool is reordered by the cross-encoder instead of the hybrid reranker; hybrid stays the default. Mock (offline overlap) + OpenAI scorers; `parseScores` degrades to zeros on malformed output. 5 tests.

Unlike `facts`, this is practical on `_m` (one scoring call per query, ~500 for a full run). QA lift target (net gain vs the lexical hybrid at equal k) pending a real Tier-2 eval.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the LongMemEval eval harness and tests; production `Retriever` behavior is unchanged unless callers adopt the new rerank wiring separately.
> 
> **Overview**
> Adds a **`rerank-cross`** LongMemEval ablation path that swaps the hybrid over-fetch reranker for a batch **cross-encoder** scorer wired through the existing `RerankFn` seam.
> 
> New `cross-encoder` module defines `createCrossEncoderRerank`, a mock token-overlap scorer, and an OpenAI chat scorer that **batches** passages (size 20), **flattens/truncates** text for prompts, and parses JSON relevance arrays via **`parseScores`** (malformed replies become all zeros so incumbent hybrid/KNN order is preserved). `run.ts` injects mock vs OpenAI scorer when the lever is on; `runner.ts` picks cross-encoder vs hybrid rerank, **records LLM cost** per batch, and **throws** if `rerank-cross` is enabled without a scorer or without `rerankPool > k` so ablations cannot silently measure the wrong thing.
> 
> Vitest coverage exercises rerank ordering, score parsing, batching/truncation, misconfiguration guards, and a small harness integration run.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3605a8f42b2ec9222c9bae13eb83c150feafa4d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/retrieval/eval/longmemeval/cross-encoder.ts
+++ b/packages/retrieval/eval/longmemeval/cross-encoder.ts
@@ -2,6 +2,11 @@ import type { QueryHit, RerankFn } from '../../src/index';
 import { chat } from './reader';
 
 const SCORER_MODEL = process.env.LONGMEMEVAL_RERANK_MODEL ?? 'gpt-5.4';
+// Chunks run up to ~6000 tokens (adapter.ts); scoring a whole 30–50 doc pool in
+// one prompt can blow past the context limit and 400 the entire run. Flatten
+// and cap each passage, and score the pool in bounded batches instead.
+const PASSAGE_CHAR_BUDGET = 2000;
+export const SCORER_BATCH_SIZE = 20;
 
 // Batch-scores (query, document) pairs — a cross-encoder / reranker-model seam.
 // One call scores the whole over-fetch pool for a query.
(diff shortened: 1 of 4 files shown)


jamby77:
Review follow-ups pushed (2a2307c):

- **Fail fast on a silent no-op**: `LONGMEMEVAL_RERANK_CROSS=1` with `rerankPool <= k` never invoked the scorer and reported a pure-baseline result labeled as the lever. `runEval` now throws with guidance instead of measuring nothing.
- **Prompt budget**: passages are flattened to one line and capped at 2000 chars; the pool is scored in batches of 20 — a 30–50 doc pool of ~6k-token chunks could previously overflow the context and 400 the whole run.
- **Order-safe degradation**: a partial or non-numeric score reply now degrades to all-zeros for that batch, so the stable sort keeps the incumbent hybrid/KNN order instead of silently demoting trailing passages.
- Cost report counts one llm call per scorer batch.

jamby77:
Both findings fixed in 3605a8f: the guard now fires on the lever itself, not on the scorer's presence — `rerank-cross` in levers without a `crossEncoderScorer` seam throws (it silently fell back to the hybrid reranker / plain top-k while the summary reported the lever), and the pool-size check is unchanged behind it. Both misconfigurations are covered by tests.

KIvanow (commented):
Net-negative: tr QA -8.3 to -13.5 (buries date-anchored evidence), ku -5.2, 918 extra gpt-5.4 calls, +47min
Should be closed

jamby77:
Closing per review — net-negative on the LongMemEval-M ablation (temporal-reasoning QA -8.3 to -13.5, +918 gpt-5.4 calls / +47min). Branch kept.

PR #294, opened 2026-07-01
feat(anomaly): detect stalled BGSAVE/AOF persistence forks

KIvanow:
## Summary
Adds a persistence-stall detector to the anomaly-detection module. Valkey issue [valkey-io/valkey#2322](https://github.com/valkey-io/valkey/issues/2322) asks for visibility when a background save gets wedged. Monitor now flags a BGSAVE / AOF-rewrite fork child that stops making progress, runs past a time ceiling, or whose last save status flips to error, so operators find out before the fork silently blocks writes or exhausts memory.

## Changes
- Added `MetricType.PERSISTENCE_CHILD` and `AnomalyPattern.PERSISTENCE_STALL` (`types.ts`).
- New state-based detector in `anomaly.service.ts` that tracks RDB and AOF fork children per connection from the INFO persistence section:
  - CRITICAL when key progress freezes for `MONITOR_PERSISTENCE_STALL_SEC` (default 60s).
  - CRITICAL when elapsed time exceeds `MONITOR_PERSISTENCE_CRIT_SEC` (default 600s).
  - CRITICAL on an `ok` to `err` last-save/last-rewrite status transition.
  - WARNING when elapsed exceeds `MONITOR_PERSISTENCE_WARN_SEC` (default 120s) while still advancing.
  - Fires once per episode and clears state when the child finishes or the connection is removed. AOF has no per-key counter, so it relies on the time-based ceilings only.
- Added a `PERSISTENCE_STALL` correlation rule in `correlator.ts` with diagnosis and remediation guidance (memory headroom / COW growth, server log, `BGSAVE CANCEL`, disk space and I/O).

## Checklist
- [ ] Docs added / updated
- [ ] [Roborev](https://www.roborev.io/) review passed - run `roborev review --branch` or `/roborev-review-branch` in Claude Code *(internal)*

Notes: 45/45 anomaly-detection tests pass, `tsc --noEmit` clean. Thresholds are env-configurable with safe fallbacks (the ConfigService mock returns non-numeric values, so parsing guards against NaN). No dedicated webhook event in this PR (deferred).

---

> [!NOTE]
> **Medium Risk**
> New monitor path on every anomaly poll reads persistence INFO and maintains per-connection state; mis-tuned thresholds or edge-case logic could cause false positives/negatives on production saves, though behavior is covered by tests and defaults are conservative.
> 
> **Overview**
> Adds **state-based monitoring** for wedged or failing RDB/AOF fork children so operators get alerted before persistence problems block writes or blow up memory.
> 
> **Anomaly pipeline:** Introduces `MetricType.PERSISTENCE_CHILD` and `AnomalyPattern.PERSISTENCE_STALL`. `AnomalyService` polls INFO persistence per connection, tracks RDB and AOF episodes (progress, elapsed time, last status), and emits **WARNING/CRITICAL** events for long runs, frozen key progress (RDB), time ceilings, and error status—with once-per-episode dedupe, re-baseline when a new child starts between polls, and guards for N/N flush tail and missing key totals. A correlator rule surfaces diagnosis and remediation for **PERSISTENCE_STALL**.
> 
> **Config & docs:** Three env vars (`MONITOR_PERSISTENCE_STALL_SEC`, `WARN`, `CRIT`) are validated in `env.schema.ts` (defaults 60/120/600). `docs/anomaly-detection.md` documents the pattern, metric, and variables.
> 
> **Tests:** Large `anomaly.service.spec` suite covers stall, warn, error latching, episode boundaries, and cleanup; `ConfigService` mock returns numeric threshold defaults.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1ea9773396c4a0b6a861b872e354578e46864323. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proprietary/anomaly-detection/anomaly.service.ts
+++ b/proprietary/anomaly-detection/anomaly.service.ts
@@ -624,9 +624,13 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
     // through the RDB flush/fsync/rename tail, during which processed is frozen at N/N —
     // on a large save over a slow disk that tail can exceed persistenceStallSec and would
     // otherwise trip a false "appears stuck (processed N/N keys)". A genuine hang in that
-    // tail is still caught by the elapsed-time ceiling (tooLong). When total is unknown
-    // (older servers without current_save_keys_total) keep the elapsed-only behavior.
-    const progressIncomplete = signals.total === null || signals.processed! < signals.total;
+    // tail is still caught by the elapsed-time ceiling (tooLong).
+    //
+    // We can only assert "keys remain" when the total is known. If current_save_keys_total
+    // is absent (processed reported without a total) we can't tell the completion tail from a
+    // real stall, so we skip frozen-progress detection entirely and rely on the elapsed-time
+    // thresholds — consistent with not raising a CRITICAL we can't substantiate.
+    const progressIncomplete = signals.total !== null && signals.processed! < signals.total;
     const frozenStall =
       signals.processed !== null &&
       progressIncomplete &&
--- a/proprietary/anomaly-detection/anomaly.service.ts
+++ b/proprietary/anomaly-detection/anomaly.service.ts
@@ -30,6 +30,7 @@ interface PersistenceChildTrack {
   startedAt: number;
   lastProcessed: number;
   lastAdvanceTs: number;
+  lastElapsedSec: number;
   warnedLong: boolean;
   reportedStall: boolean;
 }
@@ -571,6 +572,7 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
         startedAt: timestamp,
         lastProcessed: signals.processed ?? 0,
         lastAdvanceTs: timestamp,
+        lastElapsedSec: elapsedSec,
         warnedLong: false,
         reportedStall: false,
       };
@@ -578,6 +580,29 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
       else state.aof = track;
       return;
     }
+
+    // A running child's elapsed time and processed-key count are both
+    // monotonic within a single episode and reset when a new fork starts.
+    // Episode boundaries are otherwise inferred only from observing an idle
+    // (in_progress = 0) sample, which is missed when a new child begins
+    // between polls (tight save cadence, slow interval, or a failed poll).
+    // Without this guard the prior track would be reused — misreporting the
+    // fresh child as a stalled episode (stale lastAdvanceTs) or suppressing
+    // its alerts (carried-over reportedStall/warnedLong). Detect the restart
+    // via a regression in either signal and re-baseline, mirroring the
+    // first-observation branch above.
+    const elapsedRegressed = elapsedSec < track.lastElapsedSec;
+    const processedRegressed = signals.processed !== null && signals.processed < track.lastProcessed;
+    if (elapsedRegressed || processedRegressed) {
+      track.startedAt = timestamp;
+      track.lastProcessed = signals.processed ?? 0;
+      track.lastAdvanceTs = timestamp;
+      track.lastElapsedSec = elapsedSec;
+      track.warnedLong = false;
+      track.reportedStall = false;
+      return;
+    }
+    track.lastElapsedSec = elapsedSec;
 
     // Advance tracking (RDB only exposes processed-keys progress).
     if (signals.processed !== null && signals.processed > track.lastProcessed) {
--- a/proprietary/anomaly-detection/anomaly.service.ts
+++ b/proprietary/anomaly-detection/anomaly.service.ts
@@ -38,8 +38,9 @@ interface PersistenceChildTrack {
 interface ConnectionPersistenceState {
   rdb?: PersistenceChildTrack;
   aof?: PersistenceChildTrack;
-  rdbLastStatus?: string;
-  aofLastStatus?: string;
+  // Latch so a persisting error status fires once, re-armed by a later ok.
+  rdbErrorReported?: boolean;
+  aofErrorReported?: boolean;
 }
 
 @Injectable()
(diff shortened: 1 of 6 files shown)


proprietary/anomaly-detection/anomaly.service.ts
```diff
@@ -443,12 +473,207 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
...
+    state: ConnectionPersistenceState,
+    ctx: ConnectionContext,
+    timestamp: number,
+  ): Promise<void> {
+    // Completed-status transition to error (e.g. failed BGSAVE — the case #2322
+    // users disable stop-writes-on-bgsave-error around).
+    const status = signals.lastStatus;
+    if (status !== undefined && status !== '') {
+      const prevStatus = kind === 'rdb' ? state.rdbLastStatus : state.aofLastStatus;
+      if (prevStatus !== undefined && prevStatus !== status && status === 'err') {
```
jamby77: **A server already in `err` status is never reported.**

The `error` anomaly only fires on an observed transition:

```ts
if (prevStatus !== undefined && prevStatus !== status && status === 'err') {
```

`prevStatus` comes from the in-memory `lastPersistenceState`, which starts empty and is dropped in `onConnectionRemoved`. Consequences:

- If `rdb_last_bgsave_status` / `aof_last_bgrewrite_status` is already `err` when monitoring begins (monitor start, API restart, connection add/re-add), the first poll swallows the `err` as baseline and the guard can never pass while the status stays `err`.
- Repeated failures with no intervening successful save also never re-fire — only an `ok` sample re-arms the check.

That's precisely the `stop-writes-on-bgsave-error` scenario the comment above cites (#2322): the instance can be rejecting writes while the anomaly feed stays silent, and nothing else in the codebase reads these status fields. The spec only covers the ok→err transition with an explicit `ok` baseline poll first.

Suggestion: make this level-triggered rather than edge-triggered — fire when `status === 'err'` and there's no unresolved persistence-error anomaly for this connection (or use a dedup/cooldown), so a pre-existing failure state is caught on the first poll.

proprietary/anomaly-detection/anomaly.service.ts
```diff
@@ -64,6 +85,14 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
     super(connectionRegistry);
     this.correlator = new Correlator(this.correlationIntervalMs);
     this.metricExtractors = this.initializeMetricExtractors();
+
+    const positiveNumber = (value: unknown, fallback: number): number => {
+      const n = Number(value);
+      return Number.isFinite(n) && n > 0 ? n : fallback;
+    };
+    this.persistenceStallSec = positiveNumber(this.configService.get('MONITOR_PERSISTENCE_STALL_SEC'), 60);
```
jamby77: Minor: the three `MONITOR_PERSISTENCE_*_SEC` vars are parsed with a local `positiveNumber` fallback here, while the sibling `MONITOR_*` thresholds (`MONITOR_RECENT_OOM_WINDOW_MS`, `MONITOR_MEMORY_PCT_THRESHOLD`, `MONITOR_REPLICATION_LAG_BYTES`, …) are declared in the Zod schema in `apps/api/src/config/env.schema.ts`. As written, a typo like `MONITOR_PERSISTENCE_STALL_SEC=6O` silently falls back to 60 instead of failing startup validation, and the defaults live in a different place than every other monitor threshold. Worth adding the three vars to the schema and reading them plainly.

jamby77 (commented):
Reviewed the detector at the current head — the re-baseline fix for between-poll fork restarts looks sound (regression in either elapsed or processed re-arms the track cleanly). Two things bugbot hasn't flagged, one correctness, one minor.

proprietary/anomaly-detection/anomaly.service.ts
```diff
@@ -443,12 +472,227 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
...
+      // with different thresholds and messages. Prefer the frozen-progress
+      // reason when both trip (a stuck child is the more actionable signal).
+      const reason = frozenStall ? 'stall' : 'exceeded';
+      await this.addAnomaly(this.buildPersistenceEvent(kind, reason, elapsedSec, signals, timestamp, ctx), ctx);
+      return;
+    }
+
+    if (!track.reportedStall && !track.warnedLong && elapsedSec >= this.persistenceWarnSec) {
+      track.warnedLong = true;
+      await this.addAnomaly(this.buildPersistenceEvent(kind, 'long', elapsedSec, signals, timestamp, ctx), ctx);
```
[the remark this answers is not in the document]
KIvanow: Intended behavior, not a bug. The "running long" WARNING is a pure duration signal: it fires when elapsed crosses MONITOR_PERSISTENCE_WARN_SEC and its message reports elapsed time, not advancement. Instantaneous key progress is handled separately by the stall CRITICAL (frozen for MONITOR_PERSISTENCE_STALL_SEC), which latches reportedStall and suppresses this WARNING once it fires.

So the two thresholds are deliberately orthogonal: a brief pause (frozen under 60s but already past the 120s warn threshold) still deserves the "this save has run 2+ minutes" heads-up, while a sustained freeze escalates to CRITICAL on its own. Gating the WARNING on advancing progress would swallow that legitimate long-duration signal during a transient pause.

jamby77 (approved):
Approving — solid, well-tested detector.

One non-blocking correctness nit not yet raised in the thread: the frozen-stall check (`frozenStall`) has no `processed < total` guard, so a large RDB whose keys are all serialized but still fflush/fsync/rename-ing (in_progress stays 1) can trip a false `CRITICAL … appears stuck (processed N/N keys)` once the tail exceeds MONITOR_PERSISTENCE_STALL_SEC. A `signals.total === null || signals.processed < signals.total` guard closes it. Fine to address in a follow-up.

PR #296, opened 2026-07-02
feat(retrieval): preference lever for LongMemEval

jamby77:
Part of the LongMemEval recall→QA gap epic (Issue 7 — preference extraction + preference-aware rubric). Stacked on #295. This is the last Stage-1 lever.

## What

Adds the `preference` lever, three pieces behind `LONGMEMEVAL_PREFERENCE=1`:

1. **Preference extraction** — a preference-focused extractor seam (mock cue-regex for offline runs, OpenAI prompt for real ones) run through the same Phase-4 `consolidateRecordFacts` machinery (supersession, multi-source provenance, bounded concurrency). Extracted preferences are indexed as `pref_`-prefixed chunks carrying their source session ids/dates.
2. **Importance boost** — pure `promotePreferenceHits`: on recommendation-shaped questions, up to `LONGMEMEVAL_PREF_PROMOTE` (default 2) preference chunks outside the top-k window are stably promoted into its tail slots. Nothing is dropped and the relative order of everything else is preserved.
3. **Preference-aware rubric** — `createOpenAIPreferenceJudge` carries the exact rubric `diag-preference.ts` validates (grades whether the answer *honors* the preferences, not whether it restates gold); `createPreferenceAwareJudge` routes only recommendation-shaped questions to it, everything else stays on the generic judge. `verdictIsCorrect` is extracted into `judge.ts` and shared, so the parse can't drift between judges.

## Question routing without label leakage

Both the boost and the rubric key off `isPreferenceQuestion`, a lexical heuristic over the question text (recommend/suggest/advice/tips/"what should I"). The dataset's `question_type` label is never consulted, so retrieval and grading see nothing a production system wouldn't.

## Recall note (reviewers: this one differs from assemble/temporal/graph)

Unlike the reader-side-only levers, promotion runs **before** the top-k slice, so it intentionally affects measured recall — the epic's acceptance bar for this issue is single-session-preference recall ≥85%. Blast radius is bounded: non-recommendation-shaped questions keep the pool byte-for-byte, and when the pool equals k (rerank off) promotion is a structural no-op.

## Tests

12 new tests (heuristic pos/neg, promotion promote/preserve-all/cap/no-op, mock extractor, verdict parse, judge routing with spy judges, runEval integration incl. cost + no-recall-regression on the fixture). Full retrieval suite 196/196.

[...]

PR #297, opened 2026-07-03
feat(retrieval): event-time fact dates for LongMemEval

jamby77:
Part of the LongMemEval epic; stacked on #296. Directly addresses the open question on the facts lever's temporal behavior (KIvanow's in-flight temporal-facts run): a fact's date was always the *session* date it was mentioned on, never when it became true — "we moved last Christmas" said on 2026-01-01 was stamped 2026-01-01. That skews supersession (`isNewer`) and the temporal as-of filter for temporal-reasoning questions.

## What

- `parseFacts` accepts an optional per-fact `date`, but only normalized `YYYY-MM-DD` — date comparisons are lexicographic, so free-text dates ("last Tuesday") are dropped rather than corrupting `isNewer`/as-of.
- The OpenAI fact extractor prompt now includes the session date as context and asks for a fact `date` ONLY when the session states when the fact became true, resolved to `YYYY-MM-DD` (omit when unsure — conservative by design).
- Absent event dates fall back to the session date exactly as before, so extractors that never emit dates (including the mock) produce byte-identical behavior.

## Tests

2 new tests (parseFacts keeps well-formed event dates / drops free-text; consolidateRecordFacts stamps the chunk with the event date over the session date). Full suite 208/208.

If the temporal-facts run shows distillation losing dates, this is the lever to re-run it with; if temporal already holds, event dates only tighten supersession.

---

> [!NOTE]
> **Medium Risk**
> Changes LongMemEval fact dating and LLM extraction prompts, which affects temporal supersession and eval recall; backward-compatible when extractors omit dates.
> 
> **Overview**
> LongMemEval fact extraction can now attach an **event date** (when a fact became true) instead of always using the session mention date, so supersession (`isNewer`) and temporal as-of filtering align with what the user actually said.
> 
> `parseFacts` accepts an optional per-fact `date` but **only** keeps `YYYY-MM-DD`; free-text values are dropped so lexicographic comparisons stay valid. The OpenAI fact extractor prompt asks for that optional `date`, prepends `Session date: …` to the user payload for resolving relative phrases, and the extractor callback takes session `meta` (including that date). When the model omits a date, behavior is unchanged via the existing session-date fallback in consolidation.
> 
> Two tests cover strict date parsing and chunk `fields.date` preferring the extractor event date over the haystack session date.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 44a2faa9054031ed8ba92fb8b2d5cc02e5f28dbc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/retrieval/eval/longmemeval/facts.ts
+++ b/packages/retrieval/eval/longmemeval/facts.ts
@@ -144,7 +144,16 @@ export function parseFacts(raw: string): Fact[] {
     const subject = record.subject;
     const statement = record.statement;
     if (typeof subject === 'string' && typeof statement === 'string') {
-      facts.push({ subject, statement, tombstone: record.tombstone === true });
+      const fact: Fact = { subject, statement, tombstone: record.tombstone === true };
+      // Event time: when the fact became true, distinct from the session date it
+      // was mentioned on. Only a normalized YYYY-MM-DD is kept — supersession
+      // (isNewer) and the temporal as-of filter compare dates lexicographically,
+      // so free-text dates would corrupt both. Absent/dropped dates fall back to
+      // the session date in consolidateRecordFacts.
+      if (typeof record.date === 'string' && /^\d{4}-\d{2}-\d{2}$/.test(record.date)) {
+        fact.date = record.date;
+      }
+      facts.push(fact);
     }
   }
   return facts;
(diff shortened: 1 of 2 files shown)


KIvanow:
Closing it for now, as we prioritise other things and this doesn't move the needle sufficiently

PR #300, opened 2026-07-03
feat(retrieval): LongMemEval question-type filter + stratified slice

KIvanow:
## Summary

Two eval-harness knobs so a subset of the LongMemEval set can be run in isolation instead of the whole dataset. Both are harness-only (under `packages/retrieval/eval/longmemeval/`); no library code is touched.

## Changes

- **`LONGMEMEVAL_TYPE`** (run.ts + dataset.ts): comma-separated `question_type` allow-list, e.g. `LONGMEMEVAL_TYPE=temporal-reasoning,multi-session`. Unset = all types. The flat `LONGMEMEVAL_LIMIT` now counts only KEPT (filtered) records, so N of a given type run cleanly rather than N raw records that may all be filtered out.
- **`LONGMEMEVAL_PER_TYPE`** (run.ts + dataset.ts): stratified slice keeping N records of EACH `question_type` (a balanced, type-stratified sample for a paired A/B). Needed because `_m`/`_s` are grouped by type on disk, so a flat limit only samples the first type. In this mode the run total is `perType x typeCount`, and the runner's per-record cap is raised to match.
- Run banner now reports the active type filter and stratified limit (e.g. `dataset : ... (limit 1/type x6=6, type=...)`).

Usage:
```
LONGMEMEVAL_TYPE=single-session-user LONGMEMEVAL_LIMIT=50 pnpm -F @betterdb/retrieval eval:longmemeval
LONGMEMEVAL_PER_TYPE=20 pnpm -F @betterdb/retrieval eval:longmemeval
```

## Checklist

- [x] Harness-only (`eval/`), no `src/` changes
- [x] `pnpm --filter @betterdb/retrieval typecheck` clean
- [x] Prettier applied
- [x] Smoke-run verified: type filter and stratified slice both honored (banner + per-type recall)

---

> [!NOTE]
> **Low Risk**
> Eval harness and test-only changes under `eval/longmemeval`; no production retrieval library behavior is modified.
> 
> **Overview**
> Adds **LongMemEval harness** controls to run subsets and balanced samples without scanning multi‑GB files to EOF.
> 
> **`LONGMEMEVAL_TYPE`** filters records by comma-separated `question_type`; flat **`LONGMEMEVAL_LIMIT`** now counts only kept records. **`LONGMEMEVAL_PER_TYPE`** keeps up to N records per type (for datasets grouped by type on disk). Shared **`parseTypeList`** keeps the runner’s `runLimit` (`perType × typeCount`) aligned with **`loadRecords`** filtering and stratified **early-stop** via `expectedTypeCount`.
> 
> The run banner shows stratified limits (e.g. `1/type x6=6`) and active type filters. Tests cover stratified early-stop on streamed JSON and `parseTypeList` semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9469a69087b99a4a8f338afdbabb4cf2f96f3037. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/retrieval/eval/longmemeval/run.ts
+++ b/packages/retrieval/eval/longmemeval/run.ts
@@ -37,12 +37,19 @@ async function main(): Promise<void> {
   const perType = envInt('LONGMEMEVAL_PER_TYPE', 0);
   // LongMemEval has 6 question_types. In stratified mode the total = perType x
   // (number of selected types), so the runner's per-record `limit` cap must be
-  // raised to that total or it would truncate the slice.
+  // raised to that total or it would truncate the slice. Count DISTINCT types
+  // (deduped, mirroring loadRecords' Set-based allow-list) so a repeated type
+  // does not inflate the cap; an empty/whitespace-only filter selects all
+  // types, so fall back to the full count (never zero, which would stop the
+  // run at once while loadRecords still emitted a stratified sample).
   const LME_TYPE_COUNT = 6;
-  const typeCount =
-    questionType !== undefined && questionType !== ''
-      ? questionType.split(',').filter((t) => t.trim() !== '').length
-      : LME_TYPE_COUNT;
+  const selectedTypes = new Set(
+    (questionType ?? '')
+      .split(',')
+      .map((t) => t.trim())
+      .filter((t) => t !== ''),
+  );
+  const typeCount = selectedTypes.size > 0 ? selectedTypes.size : LME_TYPE_COUNT;
   const runLimit = perType > 0 ? perType * typeCount : limit;
 
   const cachePath = join(dirname(fileURLToPath(import.meta.url)), '.cache', 'embeddings.json');
@@ -77,7 +84,11 @@ async function main(): Promise<void> {
   const records = loadRecords(dataPath, limit, questionType, perType);
   const source = sourceLabel(dataPath);
 
-  const tier = qa ? 'Tier 2 (retrieval + QA)' : store.isReal || embedder.dims === 1536 ? 'Tier 1 (real recall)' : 'Tier 0 (offline)';
+  const tier = qa
+    ? 'Tier 2 (retrieval + QA)'
+    : store.isReal || embedder.dims === 1536
+      ? 'Tier 1 (real recall)'
+      : 'Tier 0 (offline)';
 
   console.log('='.repeat(64));
   console.log('LongMemEval retrieval harness — @betterdb/retrieval');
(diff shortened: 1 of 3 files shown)


packages/retrieval/eval/longmemeval/run.ts
```diff
@@ -26,6 +26,31 @@ async function main(): Promise<void> {
...
+  // does not inflate the cap; an empty/whitespace-only filter selects all
+  // types, so fall back to the full count (never zero, which would stop the
+  // run at once while loadRecords still emitted a stratified sample).
+  const LME_TYPE_COUNT = 6;
+  const selectedTypes = new Set(
+    (questionType ?? '')
+      .split(',')
+      .map((t) => t.trim())
+      .filter((t) => t !== ''),
+  );
```
jamby77: This type-list parse is copy-pasted verbatim from `dataset.ts` (the `new Set((...).split(',').map(trim).filter(!=''))`), and the early-stop math depends on the two staying identical: run.ts derives `typeCount` → `runLimit` here, while dataset.ts derives `expectedTypes` → `stratifyDone` from the same parse. If one copy's semantics drift (lowercasing, a dedup tweak), `runLimit` and `expectedTypes` silently disagree → either a truncated stratified slice or a full multi-GB scan.

Worth extracting a shared `parseTypeList(env?: string): Set<string>` in dataset.ts and importing it in both places. Non-blocking.

jamby77 (approved):
Approving — clean, well-tested, harness-only (no src/ changes), and all three prior Bugbot comments are addressed in the current head (Set dedup + LME_TYPE_COUNT fallback + the expectedTypeCount early-stop, with a test proving it stops before the malformed tail).

One cheap follow-up left inline: the type-list parse is duplicated between run.ts and dataset.ts and the early-stop math relies on them staying identical — a shared parseTypeList() would remove that coupling. Also minor: the >6-types case is capped by the hardcoded LME_TYPE_COUNT=6 (fine for real LongMemEval), and a narrowed `cap = perType ?? 0` local drops the two `as number` casts.

packages/retrieval/eval/longmemeval/run.ts
KIvanow: Done in c0c1f1de. Extracted a shared `parseTypeList(value?): Set<string>` in `dataset.ts` as the single source of truth and imported it in both `run.ts` (limit-cap sizing) and `loadRecords` (filter + early-stop), so the two derivations can no longer drift. Also narrowed the per-type cap to a `const cap = perType ?? 0` local, dropping the two `perType as number` casts you flagged. Left `LME_TYPE_COUNT = 6` hardcoded per your note. Added a focused test locking the parse semantics (trim / drop-blanks / dedupe / empty=all).

PR #302, opened 2026-07-06
feat(anomaly): data-loss guard for empty-primary full resync (valkey#579)

KIvanow:
## Summary
Detects the silent data-loss scenario from [valkey/valkey#579](https://github.com/valkey-io/valkey/issues/579): a primary without persistence restarts empty, replicas see a new `master_replid`, full-resync, and wipe their own copies. The monitor now alerts the moment the primary comes back empty (Rule A), confirms when a replica has been wiped (Rule B), dispatches a new Pro-tier `data.loss.detected` webhook, and shows a critical banner with a remediation runbook on the Anomaly dashboard.

## Changes
- `proprietary/anomaly-detection`: new `MetricType.DATASET_KEYS` + per-connection replication snapshot (`replid`, offset, uptime, keyspace key count) in `AnomalyService.pollConnection()`; Rule A (primary restarted empty with restart evidence: replid change / uptime reset / offset regression) and Rule B (replica dataset collapsed ≥90% after replid change) emit CRITICAL anomalies. Same-replid FLUSHALL is deliberately not flagged.
- `packages/shared` + `proprietary/webhook-pro`: new `data.loss.detected` webhook event (Pro tier) with `dispatchDataLossDetected()` payload (`kind`, previous/current keys, replids, role, instance).
- `apps/web`: new `DataLossAlertBanner` (critical alert + REPLICAOF NO ONE runbook, resolve action) on the Anomaly dashboard; `dataset_keys` metric label; webhook event label; `resolveAnomalyEvent` API helper.
- Event ids use `randomUUID()` so they persist on the Postgres storage adapter (uuid column).

## Checklist
- [x] Unit / integration tests added — 8 new jest cases (rules fire, restart-with-data / failover-with-data / FLUSHALL don't; 42/42 in suite) + 6 vitest cases for the banner (209/209 web suite). Live e2e repro of the issue #579 docker scenario verified Rule A within one poll, Rule B after resync, webhook delivery, and banner data.
- [ ] Docs added / updated

---

> [!NOTE]
> **High Risk**
> Changes anomaly detection and resolution paths for replication/data-loss incidents; incorrect rules or resolution semantics could miss real data loss or let critical banners dismiss without durable state.
> 
> **Overview**
> Adds **critical data-loss detection** for the empty-primary / full-resync wipe scenario (valkey#579): `AnomalyService` tracks replication snapshots and emits `dataset_keys` CRITICAL anomalies (primary restarted empty vs replica wiped), with guards for loading/FLUSHALL and a new Pro **`data.loss.detected`** webhook.
> 
> The **Anomaly dashboard** gets a `DataLossAlertBanner` fed by `activeOnly` anomaly queries (no 24h window) plus **`resolveAnomalyEvent`**; dismiss only succeeds after the server persists resolution.
> 
> **Resolution behavior** is reworked: `resolveAnomaly` / `resolveGroup` are async and storage-first for durable events; memory-only anomalies use a `persisted` flag; storage `resolveAnomaly` is **idempotent** (true if already resolved) across memory/SQLite/Postgres.
> 
> API/client tweaks: `activeOnly` on anomaly events, async resolve endpoints, and webhook form label for the new event type.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 620991b7dea4fc1722d501436d80a8b99f3a3913. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proprietary/anomaly-detection/anomaly.service.ts
+++ b/proprietary/anomaly-detection/anomaly.service.ts
@@ -757,7 +757,26 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
         limit,
         connectionId,
       });
-      return stored.map(s => this.storedToAnomalyEvent(s));
+      const storedEvents = stored.map(s => this.storedToAnomalyEvent(s));
+
+      // Union with in-memory unresolved events not yet in storage: a persist failure in
+      // addAnomaly() still leaves the incident in the cache (and still fires the Pro
+      // webhook), so the banner must surface it rather than wait for a later poll to make
+      // it durable. Dedupe by id (storage wins), apply the same filters.
+      const seen = new Set(storedEvents.map(e => e.id));
+      const inMemory = this.recentAnomalies.filter(
+        e =>
+          !e.resolved &&
+          !seen.has(e.id) &&
+          (!connectionId || e.connectionId === connectionId) &&
+          (!metricType || e.metricType === metricType) &&
+          (!severity || e.severity === severity) &&
+          (!endTime || e.timestamp <= endTime),
+      );
+
+      return [...storedEvents, ...inMemory]
+        .sort((a, b) => b.timestamp - a.timestamp)
+        .slice(0, limit);
     }
 
     const cacheThreshold = Date.now() - this.cacheTtlMs;
@@ -1006,17 +1025,26 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
     const group = this.recentGroups.find(g => g.correlationId === correlationId);
     if (!group) return false;
 
-    // Mark all anomalies in the group as resolved
+    // Storage is the source of truth (same as resolveAnomaly): only flip the cached
+    // copy for events whose resolution is durable, and report success only if EVERY
+    // event in the group persisted — otherwise a client could dismiss the group while
+    // later storage-backed polls still return some members unresolved.
     const resolvedAt = Date.now();
+    let allPersisted = true;
     for (const anomaly of group.anomalies) {
-      anomaly.resolved = true;
+      let persisted = false;
       try {
-        await this.storage.resolveAnomaly(anomaly.id, resolvedAt);
+        persisted = await this.storage.resolveAnomaly(anomaly.id, resolvedAt);
       } catch (err) {
         this.logger.error(`Failed to persist resolution for anomaly ${anomaly.id}:`, err);
       }
+      if (persisted) {
+        anomaly.resolved = true;
+      } else {
+        allPersisted = false;
+      }
     }
-    return true;
+    return allPersisted;
   }
 
   clearResolved(): number {
--- a/proprietary/anomaly-detection/anomaly.service.ts
+++ b/proprietary/anomaly-detection/anomaly.service.ts
@@ -741,7 +741,25 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
     metricType?: MetricType,
     limit = 100,
     connectionId?: string,
+    activeOnly = false,
   ): Promise<AnomalyEvent[]> {
+    // Active-incident feed (e.g. the data-loss banner): return every UNRESOLVED
+    // event of any age. Must query durable storage — the in-memory cache is
+    // capped and lost on restart, and a lingering open incident can be older
+    // than any time window — with no startTime floor so old-but-open incidents
+    // are never filtered out.
+    if (activeOnly) {
+      const stored = await this.storage.getAnomalyEvents({
+        endTime,
+        severity: severity as string,
+        metricType: metricType as string,
+        resolved: false,
+        limit,
+        connectionId,
+      });
+      return stored.map(s => this.storedToAnomalyEvent(s));
+    }
+
     const cacheThreshold = Date.now() - this.cacheTtlMs;
 
     if (!startTime || startTime >= cacheThreshold) {
--- a/proprietary/anomaly-detection/anomaly.service.ts
+++ b/proprietary/anomaly-detection/anomaly.service.ts
@@ -942,20 +942,29 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
   }
 
   async resolveAnomaly(anomalyId: string): Promise<boolean> {
-    const anomaly = this.recentAnomalies.find(a => a.id === anomalyId);
-    if (anomaly) {
-      anomaly.resolved = true;
-    }
-
-    // Also persist the resolution so it survives restarts and storage-backed queries
+    // Storage is the source of truth for later (storage-backed) polls, so
+    // persist first and only report success once the resolution is durable.
+    // Reporting success on an in-memory-only flip would let a client dismiss a
+    // banner that subsequent polls still return as unresolved.
     let persisted = false;
     try {
       persisted = await this.storage.resolveAnomaly(anomalyId, Date.now());
     } catch (err) {
       this.logger.error(`Failed to persist resolution for anomaly ${anomalyId}:`, err);
+      return false;
     }
 
-    return Boolean(anomaly) || persisted;
+    if (!persisted) {
+      return false;
+    }
+
+    // Keep the cached copy in sync with the durable store.
+    const anomaly = this.recentAnomalies.find(a => a.id === anomalyId);
+    if (anomaly) {
+      anomaly.resolved = true;
+    }
+
+    return true;
   }
 
   async resolveGroup(correlationId: string): Promise<boolean> {
--- a/proprietary/anomaly-detection/anomaly.service.ts
+++ b/proprietary/anomaly-detection/anomaly.service.ts
@@ -421,8 +421,14 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
           connectedSlaves: this.parseNumber(info.connected_slaves) ?? 0,
         };
 
+        // While the server is still loading its dataset from disk after a
+        // restart (RDB/AOF), the keyspace reports zero until the load finishes.
+        // Skip data-loss detection and keep the prior snapshot so a normal
+        // restart with persistence does not look like an empty-primary wipe.
+        const isLoading = info.loading === '1' || info.async_loading === '1';
+
         const prev = this.prevReplSnapshot.get(ctx.connectionId);
-        if (prev) {
+        if (prev && !isLoading) {
           let dataLossKind: 'primary_restarted_empty' | 'replica_wiped' | null = null;
           let message = '';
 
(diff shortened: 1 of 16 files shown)


proprietary/anomaly-detection/anomaly.service.ts
```diff
@@ -217,6 +227,17 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
     return isNaN(parsed) ? null : parsed;
   }
 
+  /** Sums `keys=` across all `db<N>` entries of the INFO keyspace section. */
+  private sumKeyspaceKeys(info: Record<string, string>): number {
+    let total = 0;
+    for (const [key, value] of Object.entries(info)) {
+      if (!/^db\d+$/.test(key)) continue;
+      const match = /keys=(\d+)/.exec(value);
```
jamby77: `sumKeyspaceKeys` never returns nonzero in production, so Rule A/B can't fire.

It regexes `keys=(\d+)` out of each `db<N>` value expecting a raw string, but the real INFO parser stores each db as an object (`{keys, expires, avg_ttl}` — see `info.parser.ts` `parseKvLine` and the consumer at `metrics.service.ts:199-201`). `convertInfoToRecord` then stringifies it to `"[object Object]"`, so the regex misses → `total` stays 0 every poll → `prev.totalKeys > 0` is never true.

Tests pass only because the mock injects `db0: 'keys=150,...'` as a **string**, a shape the parser never emits.

Fix: read `.keys` off the typed keyspace object (mirror `metrics.service.ts`) and switch the test mock to the real object shape.

KIvanow: Dug into this and the premise is actually inverted — `sumKeyspaceKeys` does fire in production. The parser does **not** store keyspace dbs as objects: `InfoParser.parse` (info.parser.ts) writes each section value as the raw substring after the colon, and `parseKvLine` is never applied to the keyspace section. So `info.keyspace.db0` is the string `"keys=150,expires=5,avg_ttl=0"`, and the regex matches. Verified against the real parser:

```
InfoParser.parse("# Keyspace\r\ndb0:keys=150,expires=5,avg_ttl=0\r\n").keyspace.db0
// => "keys=150,expires=5,avg_ttl=0"  (typeof "string")
```

`migration.service.ts:113` relies on this same string shape (`val.match(/keys=(\d+)/)`) and would throw at runtime if `val` were an object — so it is definitively a string in prod. And the existing test mock (`db0: 'keys=150,...'`) is faithful, not fictional.

The `metrics.service.ts:199` reference is actually the mis-shaped one: its `typeof val === 'object' && 'keys' in val` check is `false` against the real string, so its `keyspaceSize` is silently always 0. The root cause is a type-vs-runtime mismatch: the `KeyspaceInfo` type declares an object but the parser emits strings. (Worth a separate fix for metrics.service — flagging it.)

Addressed the underlying fragility in 182d7b4e: `sumKeyspaceKeys` now reads the typed `keyspace` section off the INFO response (before the lossy `String()` flatten that *would* produce `"[object Object]"`) and handles **both** the string and object shapes, so it is correct today and cannot silently break if the parser is later aligned with the type. Added tests for both shapes.

jamby77: You're right and I was wrong here — apologies. The anomaly service uses `getInfo` (→ `InfoParser.parse`, which keeps each section value as a raw string), not the typed `getInfoParsed`, so `keyspace.db0` is `"keys=150,..."` and the regex matches; `sumKeyspaceKeys` was firing correctly in prod. My proof cited `metrics.service.ts:199`, which is actually the mis-shaped consumer (silently 0 against the real string) — I inverted it. Ironically this contradicts what I'd verified on #304 (that `InfoParser.parse` keeps strings) and I failed to carry it back here.

`182d7b4e` is a good hardening regardless — reading the typed keyspace before the lossy flatten and handling both shapes future-proofs it, and the separate `metrics.service` `keyspaceSize === 0` bug is a real catch. Withdrawing this as a blocker.

proprietary/anomaly-detection/anomaly.service.ts
```diff
@@ -822,25 +976,47 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
...
       anomaly.resolved = true;
-      return true;
     }
-    return false;
+
+    return true;
   }
 
-  resolveGroup(correlationId: string): boolean {
+  async resolveGroup(correlationId: string): Promise<boolean> {
```
jamby77, line 1372: The storage-first fix landed on `resolveAnomaly` but not here — `resolveGroup` is now inconsistent with its sibling. It sets `anomaly.resolved = true` **before** the storage call, swallows every per-item error in the `try/catch`, and unconditionally `return true`.

So if storage is briefly down, `POST /anomaly/groups/:id/resolve` returns `{success:true}`, the UI hides the group, but the next storage-backed poll returns them still unresolved (banner reappears) and the in-memory copies are wrongly marked resolved — the exact durability gap `resolveAnomaly` was just fixed to avoid.

Mirror the `resolveAnomaly` shape: persist first, only flip the cached copy on success, and return false if any member fails to persist (or aggregate the results).

KIvanow: Should be addressed now

proprietary/anomaly-detection/anomaly.service.ts
```diff
@@ -386,6 +424,104 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
...
+                });
+            }
+          }
+        }
+
+        // Don't record the transient empty snapshot taken mid-load; otherwise
+        // the next poll (keys restored) would compare against zero.
+        if (!isLoading) {
+          this.prevReplSnapshot.set(ctx.connectionId, snapshot);
+        }
```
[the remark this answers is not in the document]
KIvanow: Acknowledged — leaving this as-is for v1.

Detecting a data loss that happened while the monitor was down (or on the very first poll of an already-empty instance) requires a durable pre-loss baseline: persisting the per-connection replication snapshot (replid/offset/totalKeys/uptime) and restoring it on the first poll. That is a new per-connection storage record across the memory/sqlite/postgres adapters — a schema change larger than what this PR is scoped for.

The in-memory baseline covers the primary case (monitor running across the loss event), which is the common one. We will revisit persisting the baseline in a follow-up if usage/interest shows the restart-window gap matters in practice.

proprietary/anomaly-detection/anomaly.service.ts
```diff
@@ -822,25 +941,38 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
...
+
+    // Also persist the resolution so it survives restarts and storage-backed queries
+    let persisted = false;
+    try {
+      persisted = await this.storage.resolveAnomaly(anomalyId, Date.now());
+    } catch (err) {
+      this.logger.error(`Failed to persist resolution for anomaly ${anomalyId}:`, err);
+    }
+
+    return Boolean(anomaly) || persisted;
```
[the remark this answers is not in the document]
KIvanow: This was fixed after the reviewed commit. `resolveAnomaly` is now persist-first: it calls `storage.resolveAnomaly` first and returns `false` on a thrown error or a `false` result, only flipping the in-memory copy after a durable success — so the banner can no longer dismiss an event that storage still returns as unresolved. See commit `fb7cc7c` ("only report resolve success when persistence is durable"). Marking as resolved.

proprietary/anomaly-detection/anomaly.service.ts
```diff
@@ -1158,25 +1331,56 @@ export class AnomalyService extends MultiConnectionPoller implements OnModuleIni
...
+
+    // Keep the cached copy in sync with the durable store.
     const anomaly = this.recentAnomalies.find(a => a.id === anomalyId);
     if (anomaly) {
       anomaly.resolved = true;
-      return true;
     }
-    return false;
+
+    return true;
```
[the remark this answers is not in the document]
KIvanow: Accurate mechanics, but leaving as-is for v1.

This only occurs when `saveAnomalyEvent` is actively failing — a storage-outage state in which no anomaly of any type can be persisted or durably resolved, not a data-loss-specific defect. It is also in tension with the earlier "Banner misses unpersisted incidents" finding, which asked us to surface in-memory incidents: showing them is what makes the storage-down case visible at all. The "vanishes on API restart" part is the same in-memory-only limitation already tracked for v1.

The proper fix is to retry persistence (re-attempt the write on later polls / when storage recovers) so the incident becomes durable and therefore dismissable — a follow-up alongside the durable replication-baseline work, not a blocker for this PR.

jamby77:
Resolve regression on Postgres: `anomaly_events.id` is `UUID PRIMARY KEY`, but failover/promotion/cluster/persistence/dup-primary events use string ids (`${connectionId}-failover-${ts}` etc.). Those inserts throw and are swallowed in `addAnomaly`, so they never persist. The new persist-first `resolveAnomaly`/`resolveGroup` then returns `false` for them → those anomaly types become undismissable in the UI on Postgres (prod).

Fix: fall back to the in-memory flip when storage reports no row (safe — a never-stored event can't be resurfaced by a storage-backed poll). Prefer this over giving everything `randomUUID()`, since the deterministic ids drive the idempotent `ON CONFLICT` upsert on sqlite/memory.

Nit: new `sumKeyspaceKeys(infoResponse: any)` violates the no-`any` rule.

KIvanow:
> Resolve regression on Postgres: `anomaly_events.id` is `UUID PRIMARY KEY`, but failover/promotion/cluster/persistence/dup-primary events use string ids (`${connectionId}-failover-${ts}` etc.). Those inserts throw and are swallowed in `addAnomaly`, so they never persist. The new persist-first `resolveAnomaly`/`resolveGroup` then returns `false` for them → those anomaly types become undismissable in the UI on Postgres (prod).
> 
> Fix: fall back to the in-memory flip when storage reports no row (safe — a never-stored event can't be resurfaced by a storage-backed poll). Prefer this over giving everything `randomUUID()`, since the deterministic ids drive the idempotent `ON CONFLICT` upsert on sqlite/memory.
> 
> Nit: new `sumKeyspaceKeys(infoResponse: any)` violates the no-`any` rule.

@jamby77 this should be fixed now. When requesting changes, please use the `request changes` review status, so it is easier to find among the GH spam

jamby77 (approved):
Approved. Solid data-loss guard — persisted-flag durability model is sound, loading guard and FLUSHALL discrimination are correct, strong test coverage. Non-blocking: a few one-line `if`s (sumKeyspaceKeys + the 3 adapters + resolveGroup guard) should be braced per CLAUDE.md; Postgres resolveAnomaly idempotency path is untested (describe.each covers Memory+Sqlite only). CI red is the unrelated pre-existing cache-proposals-sqlite failure.

PR #305, opened 2026-07-07
feat(anomaly): detect duplicate primaries in a shard (split-brain)

KIvanow:
## Summary
Adds topology-level anomaly detection for the fault described in [valkey-io/valkey#2261](https://github.com/valkey-io/valkey/issues/2261): a shard ending up with **two primaries** (split-brain). We can't fix the server-side consensus bug, but we can alert operators the moment a cluster enters the invalid state.

From any single node's `CLUSTER NODES` view, the unambiguous symptom is a hash-slot range claimed by more than one `master`-flagged node. The master with the **lower `configEpoch`** is the stale/phantom primary — exactly the node Valkey's own proposed fix demotes to a replica — so we name it in the alert.

## Changes
- **New pure detector** `proprietary/anomaly-detection/duplicate-primary-detector.ts` — finds overlapping-slot primary pairs and identifies the phantom (lower `configEpoch`). Fully unit-tested in isolation.
- **Wired into `AnomalyService.pollConnection`**, inside the existing cluster-mode block, emitting a `CRITICAL` `CLUSTER_TOPOLOGY` anomaly. Alerts are deduped per conflict signature and re-arm once the conflict resolves.
- Added `MetricType.CLUSTER_TOPOLOGY` and `AnomalyPattern.SPLIT_BRAIN`.
- Surfaced `configEpoch` on `DiscoveredNode` for future UI highlighting of the phantom primary.
- Guarded so a failed `CLUSTER NODES` call never breaks the poll loop (debug-logged, like the adjacent cluster-info block).

## Checklist
- [x] Unit / integration tests added — pure-detector spec + `AnomalyService` poll cases (emit, healthy, dedupe, resolve-and-recur, error-safe). 63 tests pass; typecheck clean.
- [ ] Docs added / updated

---

> [!NOTE]
> **Medium Risk**
> New cluster-mode poll path affects alerting on production Valkey clusters; logic is guarded for failovers and poll failures but misclassification could cause alert noise or missed incidents.
> 
> **Overview**
> Adds **topology-level split-brain alerting** when two live primaries claim overlapping hash slots (Valkey #2261), using `CLUSTER NODES` on each cluster poll.
> 
> A new pure **`duplicate-primary-detector`** finds overlapping-slot primary pairs, treats the lower **`configEpoch`** node as the stale primary, and skips transient flags (`fail`, `handshake`, etc.) so normal failovers do not false-positive. **`AnomalyService`** emits **CRITICAL** **`CLUSTER_TOPOLOGY`** events with per-conflict dedupe, re-alert after heal/recur, and clears dedupe when `getClusterNodes` fails so missed heals cannot silence alerts.
> 
> Also adds **`MetricType.CLUSTER_TOPOLOGY`**, **`AnomalyPattern.SPLIT_BRAIN`**, **`configEpoch`** on **`DiscoveredNode`**, broadens **`agent-cache`** tsconfig test excludes, and adds unit/integration tests for the detector and poll wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 24caaf53e5db0b17f1a454166fea9d2162e5a460. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proprietary/anomaly-detection/duplicate-primary-detector.ts
+++ b/proprietary/anomaly-detection/duplicate-primary-detector.ts
@@ -51,16 +51,35 @@ export function firstOverlappingRange(
 }
 
 /**
+ * Transient/unhealthy CLUSTER NODES flags. A node carrying any of these is not a
+ * stable live primary, so it must not count toward a slot-overlap conflict:
+ * during a normal failover the old primary briefly shows as `master,fail` (or
+ * `master,fail?`) while still listing its slots, and the just-promoted replica
+ * lists the same slots as `master`. Counting both as live would fire a false
+ * split-brain CRITICAL on healthy recovery. `handshake`/`noaddr` nodes are not
+ * fully known to the gossip layer yet.
+ */
+const UNHEALTHY_PRIMARY_FLAGS = ['fail', 'fail?', 'handshake', 'noaddr'];
+
+/** A node is a live primary only if it's master-flagged, owns slots, and is not transient/unhealthy. */
+function isLivePrimary(node: ClusterNode): boolean {
+  return (
+    node.flags.includes('master') &&
+    node.slots.length > 0 &&
+    !UNHEALTHY_PRIMARY_FLAGS.some((flag) => node.flags.includes(flag))
+  );
+}
+
+/**
  * Finds every pair of primaries that claim overlapping slots. Replicas (nodes
- * not flagged `master`) and masters that own no slots are ignored, so a healthy
- * primary + its replicas never trips detection.
+ * not flagged `master`), masters that own no slots, and masters in a transient
+ * or unhealthy state (`fail`/`fail?`/`handshake`/`noaddr`) are ignored, so a
+ * healthy primary + its replicas — and a normal failover — never trip detection.
  */
 export function detectDuplicatePrimaries(
   nodes: ClusterNode[],
 ): DuplicatePrimaryConflict[] {
-  const masters = nodes.filter(
-    (n) => n.flags.includes('master') && n.slots.length > 0,
-  );
+  const masters = nodes.filter(isLivePrimary);
 
   const conflicts: DuplicatePrimaryConflict[] = [];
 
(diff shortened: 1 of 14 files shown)


proprietary/anomaly-detection/duplicate-primary-detector.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Finds every pair of primaries that claim overlapping slots. Replicas (nodes
+ * not flagged `master`) and masters that own no slots are ignored, so a healthy
+ * primary + its replicas never trips detection.
+ */
+export function detectDuplicatePrimaries(
+  nodes: ClusterNode[],
+): DuplicatePrimaryConflict[] {
+  const masters = nodes.filter(
+    (n) => n.flags.includes('master') && n.slots.length > 0,
```
jamby77: This filter matches `fail`/`handshake` nodes as live masters, which can fire a false CRITICAL during a normal failover: from one node's `CLUSTER NODES` view the old primary transiently shows as `master,fail` still listing its slots while the just-promoted replica already lists the same slots as `master` — both pass `flags.includes('master')`, overlap is found, and split-brain alerts on healthy recovery.

Suggest excluding transient/unhealthy states, e.g. also require the node isn't flagged `fail`/`handshake`/`noaddr` before treating it as a live primary.

jamby77 (approved):
Approving — cleanest of the anomaly detectors: well-separated pure module, correct data-shape assumptions, both prior Bugbot comments already addressed.

One robustness nit left inline (non-blocking): the master filter can count a `master,fail` node during a normal failover and fire a false CRITICAL split-brain. A one-line flag exclusion closes it. Minor follow-ups: equal-configEpoch ties name an arbitrary phantom (soften the message), a couple of single-line conditionals vs CLAUDE.md, and `as any` in the detector spec.

proprietary/anomaly-detection/duplicate-primary-detector.ts
KIvanow: Nit addressed and resolved :)

PR #310, opened 2026-07-09
feat(anomaly): detect stuck orphaned replicas after primary loss (valkey#2090)

KIvanow:
## Summary

Adds monitoring for the stuck-cluster state in [valkey-io/valkey#2090](https://github.com/valkey-io/valkey/issues/2090) — an open, ~14-month-old cluster bug the maintainers haven't fixed (it's blocked on an unresolved design question). After a primary is lost and replaced by a fresh node (classic k8s pod restart: node returns at the same address with a new node/shard id), a surviving replica keeps replicating the **old** primary and never adopts the replacement, staying orphaned until an operator intervenes.

We can't fix it server-side, but we can **detect** it and recommend the maintainers' own remediation.

## What it does

- **`stuck-replica-detector.ts`** — pure detector over a single `CLUSTER NODES` view: flags any replica whose `master` resolves to a dead (`fail`/`fail?`/`noaddr`) or absent primary. `CLUSTER NODES` doesn't expose shard ids (only `nodes.conf`/`CLUSTER SHARDS` do), so detection uses the always-present `flags` + `master` fields.
- **`anomaly.service.ts`** — wires it into the cluster poll next to the existing duplicate-primary (#2261) detector. Emits a `WARNING` on `CLUSTER_TOPOLOGY` **only after the orphaned pair persists past a 30s grace window**, so a normal failover's brief orphaned window never trips it. Dedupes per `(replica, primary)` pair and clears state on recovery so a resolved-then-recurring case re-alerts. The alert message recommends `CLUSTER REPLICATE <new-primary-id>`.

## Scope

Detection only. One-click `CLUSTER REPLICATE` remediation is intentionally a **separate follow-up** (it's resync-triggering, so it needs its own guardrails).

## Tests

- 10 pure-detector cases, with fixtures built directly from the issue's own `CLUSTER NODES` dumps (healthy shard, the #2090 orphaned replica, `fail?`/`noaddr`, unknown primary, legacy `slave` vs newer `replica` flag, multiple orphans).
- 7 service cases: no-alert-within-grace, no-alert-on-transient-failover, fire-after-grace, dedupe, resolve/recur, healthy cluster, poll failure.
- Full anomaly suite: 168 passing; typecheck clean.

## Follow-ups (from the implementation plan)
- `CLUSTER SHARDS`-based shard-id divergence as a corroborating signal.
- One-click guarded `CLUSTER REPLICATE` remediation.

---

> [!NOTE]
> **Medium Risk**
> Changes cluster anomaly alerting during failovers; false positives are mitigated by persistence gating but mis-timed alerts could still add operator noise on flaky topology polls.
> 
> **Overview**
> Adds **cluster topology monitoring** for replicas that stay pointed at a dead or missing primary after failover (valkey#2090), alongside the existing duplicate-primary detector.
> 
> A new **`stuck-replica-detector`** scans `CLUSTER NODES` and flags replicas whose `master` is failed/unreachable or absent. **`AnomalyService`** runs this on each cluster poll, waits **30s** before alerting so normal failovers do not noise, **dedupes** per `(replica, primary)` pair, clears state on recovery or failed polls, and emits a **`CLUSTER_TOPOLOGY` WARNING** that suggests `CLUSTER REPLICATE <new-primary-id>`. Connection teardown clears the new per-connection grace/dedupe maps.
> 
> Unit and service tests cover the #2090 fixture, grace/dedupe/recur behavior, and healthy clusters.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e67ccc7f034546efab5aca5f882f528b0a477dc1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proprietary/anomaly-detection/__tests__/anomaly.service.spec.ts
+++ b/proprietary/anomaly-detection/__tests__/anomaly.service.spec.ts
@@ -1797,4 +1797,126 @@ describe('AnomalyService', () => {
       expect(events).toHaveLength(2);
     });
   });
+
+  // ─── Stuck-replica detection (valkey-io/valkey#2090) ───────────────────────
+  describe('stuck replica detection', () => {
+    const clusterInfoResponse = {
+      server: { role: 'master' },
+      clients: { connected_clients: '10', blocked_clients: '0' },
+      memory: { used_memory: '1000000', allocator_frag_ratio: '1.1' },
+      stats: {
+        instantaneous_ops_per_sec: '100',
+        instantaneous_input_kbps: '50',
+        instantaneous_output_kbps: '30',
+        evicted_keys: '0',
+        keyspace_misses: '5',
+        rejected_connections: '0',
+        acl_access_denied_auth: '0',
+        cluster_enabled: '1',
+      },
+    };
+
+    let now: number;
+
+    beforeEach(() => {
+      (dbClient.getInfoParsed as jest.Mock).mockResolvedValue(clusterInfoResponse);
+      dbClient.getClusterInfo = jest.fn().mockResolvedValue({ cluster_state: 'ok' });
+      now = 1_700_000_000_000;
+      jest.spyOn(Date, 'now').mockImplementation(() => now);
+    });
+
+    afterEach(() => {
+      (Date.now as jest.Mock).mockRestore();
+    });
+
+    const healthyNodes = [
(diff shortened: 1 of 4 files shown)


jamby77 (approved):
Approved. Solid stuck-replica detector (valkey#2090), well-tested from the issue's real CLUSTER NODES dumps, cleanly wired next to the duplicate-primary detector; grace/dedupe/recover logic is correct. Non-blocking notes: (1) the 30s persistence gate is fixed while the detector's own header calls for '≥ cluster node timeout' — clusters with node-timeout >30s could false-positive on a slow failover; consider deriving it from node-timeout. (2) A few one-line `if`s should be braced per CLAUDE.md.

PR #262, opened 2026-06-19
feat(agent-cache/ai): cache streaming LLM responses via wrapStream

amitkojha05:
## Summary

The agent-cache Vercel AI SDK adapter currently implements `wrapGenerate` for non-streaming calls. Streaming responses are not cached, requiring callers to manually accumulate streamed content and invoke `store()` themselves.

This PR adds support for `wrapStream`.

On a cache miss, the upstream stream is tee'd:
- One branch flows to the caller unchanged.
- The second branch accumulates `text-delta` chunks and stores the final response via `cache.llm.store()` on `finish`.

On a cache hit:
- `doStream` is skipped entirely.
- The cached response is replayed as a single `text-delta` event followed by a `finish` event.
- The finish event includes:

```ts
providerMetadata: {
  agentCache: {
    hit: true,
  },
}
```

Tool-call streams are not cached, matching existing `wrapGenerate` behavior.

Additional guarantees:
- Store failures never affect the caller-facing stream.
- `check()` failures fall through to the upstream provider.
- Fully backward compatible; existing `wrapGenerate` consumers are unaffected.

### `adapters/ai.ts`
- Added `synthesizeStreamFromCached()`
- Added `accumulateAndStore()`
- Added `wrapStream` to the middleware return object
- No changes to existing `wrapGenerate` behavior

### `adapters/__tests__/ai.test.ts`
Added 5 new tests:
1. Cache miss → cache hit flow
2. `providerMetadata` included on cache hits
3. Tool-call stream bypass
4. Upstream error propagation
5. Store failure does not break caller stream

### `CHANGELOG.md`
- Added v0.7.0 entry

### `README.md`
- Updated Vercel AI SDK documentation section

## Testing

```bash
pnpm --filter @betterdb/agent-cache test
pnpm --filter @betterdb/agent-cache build
```

PR #311, opened 2026-07-11
feat(agent-cache): add CachedChatModel for tool-aware LangChain caching

amitkojha05:
## Summary

Adds `CachedChatModel`, a `BaseChatModel` wrapper for LangChain that caches at the model level instead of through LangChain's `BaseCache`, so bound tool schemas (`.bindTools(...)`), `tool_choice`, and the wrapped model's sampling parameters are all included in the cache key. This closes the tool-schema-drift limitation documented under `[0.7.0]` in this package's changelog:

> LangChain's `BaseCache` interface exposes only `(prompt, llm_string)`, so tool definitions are structurally unreachable.

`BetterDBLlmCache` is unchanged, so existing cache keys remain fully compatible. `CachedChatModel` is introduced as a new opt-in export for applications that bind tools.

---

### `src/adapters/langchain.ts`

Introduces:

- `CachedChatModel`
- `CachedChatModelOptions`
- `CachedChatModelCallOptions`

Also adds helper utilities for:

- `convertMessage`
- `convertTools`
- `tokensOf`
- `buildToolCallBlocks`
- `rebuildGeneration`

Overrides both:

- `_generate()`
- `_streamResponseChunks()`

so that both `.invoke()` and `.stream()` participate in cache lookup and storage.

---

### `src/adapters/__tests__/langchain.test.ts`

Adds **13 tests** covering:

- Cache hit with identical tools
- Cache miss when bound tools differ
- No-tools behavior matching `BetterDBLlmCache`
- Tool-call round-trip through `storeMultipart`
- Token extraction from `usage_metadata`
- **Tool-schema drift regression** (different Zod schemas now produce different cache keys)
- **`tool_choice` participates in the cache key**
- **Sampling parameters (`temperature`, `max_tokens`, etc.) participate in the cache key**
- Streaming cache miss passes chunks through while storing the aggregated response
- Streaming cache hit replays the cached response without invoking the model
- Fail-open behavior for cache read errors
- Fail-open behavior for cache write errors

---

### `README.md`

Adds a new section:

> **Tool-schema-aware caching (`CachedChatModel`)**

Also updates the Known Limitations section to clarify:

- `CachedChatModel` is the recommended solution for tool-schema-aware caching.
- `BetterDBLlmCache` intentionally retains the limitations imposed by LangChain's `BaseCache`.

Documents two additional limitations specific to `CachedChatModel`:

- Cache hits do not report token usage.
- Callback traces appear as separate root runs.

---

### `CHANGELOG.md`

Adds an `[Unreleased]` entry following the existing release workflow used in the #262 → #306 sequence.

---

### `examples/langchain/index.ts`

Adds a new `askWithTools` example demonstrating:

- Same prompt
- Different bound tool sets
- Second invocation correctly resulting in a cache miss

---

### `package.json`

Adds:

```json
"zod": "^3.25.0"
```

as a **devDependency**.

This dependency is used only by the tool-schema-drift regression tests and introduces **no runtime or peer dependency changes**.

---

### `bindTools`

`BaseChatModel.bindTools()` does not provide a default implementation, and the dependency version (`@langchain/core@^1.1.35`) no longer includes `Runnable.bind()` (superseded by `withConfig()`).

`CachedChatModel.bindTools()` is implemented using:

```ts
this.withConfig({ tools, ...kwargs })
```

This matches the implementation used internally by `ChatOpenAI.bindTools()` (verified against the compiled `@langchain/openai` source).

The method:

- returns a new runnable,
- does **not** mutate the existing instance,
- follows LangChain conventions.

---

### Tool Schema Normalization

Tool schemas are normalized using LangChain's `convertToOpenAITool()` before contributing to the cache key.

This is necessary because raw Zod schemas (`tool()` / `DynamicStructuredTool`) do **not** serialize deterministically with `JSON.stringify()`. In Zod v3, object shapes are stored lazily as functions under `_def.shape`, meaning two structurally different schemas with the same tool name could otherwise produce identical hashes.

Using `convertToOpenAITool()`:

- converts Zod schemas into canonical JSON Schema,
- leaves already-OpenAI-formatted tools unchanged,
- preserves existing usage,
- strips `$schema` before hashing so a Zod v3 → v4 upgrade does not invalidate cache entries.

---

### Streaming

`_streamResponseChunks()` is overridden because the default `BaseChatModel` implementation falls back to `_generate()` if streaming is not implemented, which would silently degrade `.stream()` into a single generated response.

Behavior:

**Cache miss**

- Upstream chunks stream through unchanged.
- The aggregated response is stored after completion.

**Cache hit**

- Cached output is replayed as a single chunk.

This matches the behavior of `wrapStream` used by the Vercel AI SDK adapter (#262).

---

### Cache-Key Inputs Beyond Messages and Tools

The cache key also incorporates:

#### Call options

- `tool_choice`

#### Wrapped model invocation parameters

- `temperature`
- `top_p`
- `max_tokens`
- `seed`
- `stop`

These are obtained via `invocationParams()`.

As a result, two otherwise identical requests using different sampling parameters do **not** share cache entries.

This behavior is consistent with the existing:

- `openai-chat`
- `anthropic`
- `ai`

adapters.

---

## Checklist

- [x] Documentation added / updated
- [ ] Roborev review passed (`roborev review --branch` or `/roborev-review-branch` in Claude Code)
- [ ] Competitive analysis completed / discussed (internal)
- [ ] Blog post discussed (internal)

### DIFF
--- a/packages/agent-cache/CHANGELOG.md
+++ b/packages/agent-cache/CHANGELOG.md
@@ -5,6 +5,31 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **`CachedChatModel` — tool-schema-aware LangChain caching.** A `BaseChatModel`
+  wrapper (`@betterdb/agent-cache/langchain`) that caches at the model level
+  instead of through LangChain's `BaseCache`, so bound tool schemas
+  (`.bindTools(...)`), `tool_choice`, and the wrapped model's sampling params are
+  all included in the cache key. Wrap any chat model:
+  `new CachedChatModel({ model, cache })`. Tool schemas are normalized with
+  LangChain's `convertToOpenAITool`, so Zod-based tools key on canonical JSON
+  Schema rather than on Zod internals. Streaming is supported: on a miss the
+  upstream chunks pass through untouched and the aggregate is stored on
+  completion; on a hit the stored response replays as a single chunk. Tool-call
+  responses are stored via `storeMultipart` and rebuilt on hit.
+
+  Resolves the tool-schema-drift limitation documented under v0.7.0 for callers
+  who opt into the wrapper; the existing `BetterDBLlmCache` (`cache=`) path is
+  unchanged and retains the `(prompt, llm_string)` limitation by design.
+
+### Behavior
+
+- No cache keys change for existing `BetterDBLlmCache` users. `CachedChatModel`
+  is a separate, opt-in integration.
+
 ## [0.11.1] - 2026-07-09
 
 ### Fixed
--- a/packages/agent-cache/README.md
+++ b/packages/agent-cache/README.md
@@ -278,6 +278,8 @@ const effectiveness = await cache.toolEffectiveness();
 
 ### LangChain
 
+**Zero-config caching** — plug into LangChain's native `cache` option:
+
 ```typescript
 import { ChatOpenAI } from '@langchain/openai';
 import { BetterDBLlmCache } from '@betterdb/agent-cache/langchain';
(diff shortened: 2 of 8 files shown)


amitkojha05:
@KIvanow ready for review whenever you have a moment.

Two things not covered in the description:
- `.github/workflows/api-tests.yml` doesn't actually run `packages/agent-cache`'s `tsc`/vitest (same gap as #306) — I've run both locally and they're clean on this branch. Happy to add a workflow job for this in a separate PR if useful.
- The new `zod` devDep is test-only (regression test for the schema-drift fix), not a runtime/peer dep.

PR #316, opened 2026-07-14
feat(logs): magnitude sort (top-N worst offenders) for slow/command log

KIvanow:
## Summary

Adds an optional `sortBy: 'recent' | 'magnitude'` to the SLOWLOG and COMMANDLOG entry queries. `magnitude` returns the **worst offenders first** (top-N by `duration` — microseconds for `slow`, bytes for `large-request`/`large-reply`) instead of only newest-first.

This closes the interim gap behind [valkey-io/valkey#3895](https://github.com/valkey-io/valkey/issues/3895) — a request for a `magnitude` retention mode so that bursts of mildly-slow commands can't evict the genuine outliers from Valkey's small, recency-based in-memory log. We already sidestep that limitation by **persisting every SLOWLOG/COMMANDLOG entry durably** rather than relying on the ~128-entry ring buffer, so the outliers are already retained — this PR just lets callers **rank by them**. Reuses the existing `idx_slowlog_duration` / `idx_commandlog_duration` indexes, so there is **no migration**.

Default behaviour is unchanged (`recent`).

## Changes
- `common/interfaces/storage-port.interface.ts`: `CommandLogSortBy` type + `sortBy?` on `SlowLogQueryOptions` and `CommandLogQueryOptions`.
- SQL adapters (sqlite + postgres) and the memory adapter: order by `duration DESC, timestamp DESC` when `sortBy === 'magnitude'` — across the slowlog repositories and the commandlog getter methods.
- `slowlog-analytics` / `commandlog-analytics` controllers: `sortBy` query param (validated, defaults to `recent`).
- New spec `log-magnitude-sort.spec.ts`: magnitude ordering, top-N via `limit`, `type` filter, and the recency default — run against both `MemoryAdapter` and `SqliteAdapter`.

## Checklist
- [ ] Docs added / updated

---

> [!NOTE]
> **Low Risk**
> Additive query parameter and read-path ORDER BY only; invalid values fall back to `recent` with no migration or write-path changes.
> 
> **Overview**
> Adds optional **`sortBy`** (`recent` | `magnitude`) on **slowlog-analytics** and **commandlog-analytics** `entries` APIs, wired through `SlowLogQueryOptions` / `CommandLogQueryOptions` and all storage paths (memory, SQLite, Postgres).
> 
> **`recent`** keeps today’s behavior (newest first). **`magnitude`** orders by `duration DESC` (then `timestamp DESC`) so **`limit`** can return top-N worst offenders from persisted logs. For command logs, `duration` is µs for `slow` and bytes for large-request/reply — Swagger notes that mixing types under `magnitude` compares unlike units unless a single `type` is set.
> 
> New adapter tests (`log-magnitude-sort.spec.ts`) assert default recency, magnitude ordering, and limit + type filter; no schema migration (existing duration indexes).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea6427e0b0b150ef5bd62b52d32693f53e614335. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/api/src/commandlog-analytics/commandlog-analytics.controller.ts
+++ b/apps/api/src/commandlog-analytics/commandlog-analytics.controller.ts
@@ -20,6 +20,7 @@ export class CommandLogAnalyticsController {
   @ApiQuery({ name: 'minDuration', required: false, description: 'Minimum duration in microseconds' })
   @ApiQuery({ name: 'limit', required: false, description: 'Maximum number of entries to return' })
   @ApiQuery({ name: 'offset', required: false, description: 'Offset for pagination' })
+  @ApiQuery({ name: 'sortBy', required: false, enum: ['recent', 'magnitude'], description: "Order: 'recent' (newest first, default) or 'magnitude' (worst offenders / top-N by duration or size)" })
   async getStoredCommandLog(
     @ConnectionId() connectionId?: string,
     @Query('startTime') startTime?: string,
@@ -30,6 +31,7 @@ export class CommandLogAnalyticsController {
     @Query('minDuration') minDuration?: string,
     @Query('limit') limit?: string,
     @Query('offset') offset?: string,
+    @Query('sortBy') sortBy?: string,
   ): Promise<StoredCommandLogEntry[]> {
     return this.commandLogAnalyticsService.getStoredCommandLog({
       startTime: startTime ? parseInt(startTime, 10) : undefined,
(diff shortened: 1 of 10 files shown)


jamby77 (approved):
LGTM. Additive, safe default, well-tested across Memory + Sqlite.

Issue #329, opened 2026-07-22
MCP update


jamby77:
Review our MCP server for missing functionalities based on recent PRs and everything we've added since the last integration of commands in it.

Suggest new tools for it. We'll discuss them. Then implement them

---

## Proposal: tools to add

The server exposes ~50 tools but has no coverage for several recently shipped features (AI observability, forecasting, inference latency, vector search, key analytics). Proposed additions, all read-only:

### AI Observability (biggest gap)
- `ai_list_instances` — discovered AI cache/memory/retrieval instances + latest sample
- `ai_instance_history` — time-series for one instance
- `list_ai_traces` / `get_ai_trace` — recent AI traces + a trace's span waterfall
- `correlate_ai_trace` — join a trace's cache/memory spans with live Valkey state to explain a hit/miss (the differentiator)

### Analytics (companions to existing tools)
- `get_forecast` — projected time-to-limit for memory/keys
- `get_latency_regressions` — detected regressions (like `get_anomalies`)
- `get_largest_keys` — top keys by memory (complements `get_hot_keys`)

### Vector search (audience-gated; do together)
- `get_inference_latency` — per-index FT.SEARCH p50/p95/p99 + SLA breach
- `get_vector_indexes` — docs, memory, indexing failures, % indexed

**Suggested order:** AI Observability, then Analytics, then Vector search.

PR #330, opened 2026-07-22
feat(mcp): AI observability MCP tools

jamby77:
Adds five read-only MCP tools exposing the AI observability feature to agents:

- `ai_list_instances` — discover AI component instances (semantic caches, agent caches, memory stores, retrieval pipelines) with liveness and latest metrics
- `ai_instance_history` — stored metrics time-series for one AI instance
- `list_ai_traces` — recent OpenTelemetry AI traces
- `get_ai_trace` — full span waterfall for one trace
- `correlate_ai_trace` — join a trace's cache/memory spans with live Valkey state

Server side: new `McpAiController` with `/mcp/*` GET routes delegating to `AiObservabilityService` and `TraceCorrelationService` (now exported from `AiObservabilityModule`), plus a shared `mapMcpError` helper (501 for capability errors, logged 500 otherwise). Client side: five `server.tool` registrations in the MCP package, with sibling cross-references to `cache_list` / `memory_stores`.

Includes a standalone prettier-only formatting commit for the MCP server index so the feature diff stays reviewable.

Part of #329.

https://claude.ai/code/session_01BeJC3cWSEXFGD7MLMXaCiH

---

> [!NOTE]
> **Medium Risk**
> New authenticated MCP surface on observability data with bounded query params; vector-search error mapping changes could alter HTTP status for missing capabilities (501 vs string-matched 501).
> 
> **Overview**
> Exposes **AI observability** to MCP agents via a new **`McpAiController`** (agent-token guarded `/mcp` routes) and five matching **`server.tool`** registrations: instance discovery, per-instance metrics history (downsampled to 200 points), trace listing, span detail, and **trace–Valkey correlation**.
> 
> Shared plumbing includes **`CapabilityUnavailableError`** → **501** through **`mapMcpError`**, a reusable **`downsampleSeries`** helper, and stricter **`safeLimit`** (empty strings, fractional ceil, per-route max). **`VectorSearchService`** / **`VectorSearchController`** adopt the same error and downsampling patterns instead of ad hoc logic. **`TraceCorrelationService`** is exported from **`AiObservabilityModule`** for MCP wiring.
> 
> Controller behavior is covered by **`mcp-ai.controller.spec.ts`**. The MCP package index also includes a broad formatting-only reshuffle alongside the new AI tools.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d81c5902e3fa0a64a7532ad6693a4d9342c34cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/api/src/mcp/ai/mcp-ai.controller.ts
+++ b/apps/api/src/mcp/ai/mcp-ai.controller.ts
@@ -1,0 +1,40 @@
+import { Controller, Get, Logger, Param, Query, UseGuards } from '@nestjs/common';
+import { AgentTokenGuard } from '../../common/guards/agent-token.guard';
+import { ValidateInstanceIdPipe, mapMcpError, safeParseInt } from '../mcp-helpers';
+import { AiObservabilityService } from '../../ai-observability/ai-observability.service';
+import { TraceCorrelationService } from '../../ai-observability/trace-correlation.service';
+
+@Controller('mcp')
+@UseGuards(AgentTokenGuard)
+export class McpAiController {
+  private readonly logger = new Logger(McpAiController.name);
+
+  constructor(
+    private readonly aiObservability: AiObservabilityService,
+    private readonly traceCorrelation: TraceCorrelationService,
+  ) {}
+
+  @Get('instance/:id/ai/instances')
+  async getInstances(@Param('id', ValidateInstanceIdPipe) id: string) {
+    try {
+      const instances = await this.aiObservability.getInstances(id);
+      return { instances };
+    } catch (error) {
+      throw mapMcpError(this.logger, error, 'Failed to list AI instances');
+    }
+  }
+
+  @Get('instance/:id/ai/instances/:field/history')
+  async getHistory(
+    @Param('id', ValidateInstanceIdPipe) id: string,
+    @Param('field') field: string,
+    @Query('hours') hours?: string,
+  ) {
+    try {
+      const samples = await this.aiObservability.getHistory(id, field, safeParseInt(hours, 24));
+      return { samples };
+    } catch (error) {
+      throw mapMcpError(this.logger, error, 'Failed to get AI instance history');
+    }
+  }
+}
--- a/apps/api/src/mcp/ai/mcp-ai.controller.ts
+++ b/apps/api/src/mcp/ai/mcp-ai.controller.ts
@@ -5,20 +5,8 @@ import { ValidateInstanceIdPipe, mapMcpError, safeLimit } from '../mcp-helpers';
 const MAX_HISTORY_HOURS = 168;
 const MAX_TRACE_LIMIT = 1000;
 const MAX_HISTORY_POINTS = 200;
-
-function downsample<T>(samples: T[], maxPoints: number): T[] {
-  if (samples.length <= maxPoints) {
-    return samples;
-  }
-  const step = samples.length / maxPoints;
-  const sampled: T[] = [];
-  for (let i = 0; i < maxPoints - 1; i++) {
-    sampled.push(samples[Math.floor(i * step)]);
-  }
-  sampled.push(samples[samples.length - 1]);
-  return sampled;
-}
 import { AiObservabilityService } from '../../ai-observability/ai-observability.service';
+import { downsampleSeries } from '../../common/utils/downsample';
 import { TraceCorrelationService } from '../../ai-observability/trace-correlation.service';
 
 @Controller('mcp')
@@ -53,7 +41,7 @@ export class McpAiController {
         field,
         safeLimit(hours, 24, MAX_HISTORY_HOURS),
       );
-      return { samples: downsample(samples, MAX_HISTORY_POINTS) };
+      return { samples: downsampleSeries(samples, MAX_HISTORY_POINTS) };
     } catch (error) {
       throw mapMcpError(this.logger, error, 'Failed to get AI instance history');
     }
--- a/apps/api/src/mcp/ai/mcp-ai.controller.ts
+++ b/apps/api/src/mcp/ai/mcp-ai.controller.ts
@@ -1,9 +1,23 @@
 import { Controller, Get, Logger, Param, Query, UseGuards } from '@nestjs/common';
 import { AgentTokenGuard } from '../../common/guards/agent-token.guard';
-import { ValidateInstanceIdPipe, clampedParseInt, mapMcpError } from '../mcp-helpers';
+import { ValidateInstanceIdPipe, mapMcpError, safeLimit } from '../mcp-helpers';
 
 const MAX_HISTORY_HOURS = 168;
 const MAX_TRACE_LIMIT = 1000;
+const MAX_HISTORY_POINTS = 200;
+
+function downsample<T>(samples: T[], maxPoints: number): T[] {
+  if (samples.length <= maxPoints) {
+    return samples;
+  }
+  const step = samples.length / maxPoints;
+  const sampled: T[] = [];
+  for (let i = 0; i < maxPoints - 1; i++) {
+    sampled.push(samples[Math.floor(i * step)]);
+  }
+  sampled.push(samples[samples.length - 1]);
+  return sampled;
+}
 import { AiObservabilityService } from '../../ai-observability/ai-observability.service';
 import { TraceCorrelationService } from '../../ai-observability/trace-correlation.service';
 
@@ -37,9 +51,9 @@ export class McpAiController {
       const samples = await this.aiObservability.getHistory(
         id,
         field,
-        clampedParseInt(hours, 24, MAX_HISTORY_HOURS),
+        safeLimit(hours, 24, MAX_HISTORY_HOURS),
       );
-      return { samples };
+      return { samples: downsample(samples, MAX_HISTORY_POINTS) };
     } catch (error) {
       throw mapMcpError(this.logger, error, 'Failed to get AI instance history');
     }
@@ -54,9 +68,9 @@ export class McpAiController {
     try {
       const hasService = service !== undefined && service !== '';
       const traces = await this.aiObservability.getTraces(
-        clampedParseInt(hours, 1, MAX_HISTORY_HOURS),
+        safeLimit(hours, 1, MAX_HISTORY_HOURS),
         hasService === true ? service : undefined,
-        clampedParseInt(limit, 100, MAX_TRACE_LIMIT),
+        safeLimit(limit, 100, MAX_TRACE_LIMIT),
       );
       return { traces };
     } catch (error) {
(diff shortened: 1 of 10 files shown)


jamby77:
bugbot run

jamby77:
bugbot run

apps/api/src/mcp/mcp-helpers.ts
```diff
@@ -39,3 +62,15 @@ export function msToSeconds(value: string | undefined): number | undefined {
   }
   return Math.floor(ms / 1000);
 }
+
+export function mapMcpError(logger: Logger, error: unknown, fallback: string): HttpException {
+  if (error instanceof HttpException) {
+    return error;
+  }
+  const message = error instanceof Error ? error.message : fallback;
+  if (message.includes('not available') || message.includes('not supported')) {
```
jamby77: Two problems with this classification:

1. Any error whose message merely *contains* "not available"/"not supported" becomes a 501 carrying the raw internal message — e.g. a transient "database connection not available" from the driver. Agents treat 501 as "capability permanently missing" and stop retrying, and since `logger.error` only runs on the 500 fallback below, these are never logged server-side. A typed error (e.g. `CapabilityUnavailableError`, like `InferenceLatencyValidationError` already does for the 400 class) checked with `instanceof` would make this robust.

2. This near-duplicates the private `mapError` in `vector-search.controller.ts` (~line 180), which this PR doesn't delete — and the two already diverge (that one returns `${fallback}: ${msg}`, this one the bare message). If the shared helper is the way forward, the vector-search copy should be folded into it.

apps/api/src/mcp/mcp-helpers.ts
```diff
@@ -31,6 +41,19 @@ export function safeLimit(value: string | undefined, defaultValue: number): numb
   return Math.max(1, Math.min(safeParseInt(value, defaultValue), MAX_LIMIT));
 }
 
+/** Parse a positive query param, clamping to [1, max]; non-positive or invalid values fall back to the default. */
+export function clampedParseInt(
```
jamby77: Two clamping issues:

1. The tool schemas declare `hours: z.number()` (not `.int()`), so `hours: 0.5` is schema-legal and serialized as `"0.5"` — `parseInt('0.5')` → `0` → `<= 0` → silently replaced with the default. An agent asking for 30 minutes of history gets 24 hours (~5,760 samples) mislabeled as the requested window. Either `.int().min(1).max(168)` in the schemas or `Math.ceil(Number(value))` here.

2. This helper now coexists with `safeLimit` above with divergent semantics: `limit=-5` yields **1** result via `safeLimit` (`get_latency_regressions`, `get_largest_keys` in the later layers) but **100** via `clampedParseInt` (`list_ai_traces`). One parameterized helper (`safeLimit` with an optional `max`) would give sibling endpoints consistent behavior — and encoding the ranges in the zod schemas would give agents the constraints they currently only see in prose.

apps/api/src/mcp/ai/mcp-ai.controller.ts
```diff
@@ -0,0 +1,89 @@
...
+  async getHistory(
+    @Param('id', ValidateInstanceIdPipe) id: string,
+    @Param('field') field: string,
+    @Query('hours') hours?: string,
+  ) {
+    try {
+      const samples = await this.aiObservability.getHistory(
+        id,
+        field,
+        clampedParseInt(hours, 24, MAX_HISTORY_HOURS),
```
jamby77: This clamps the *window* but not the *row count*: `getHistory` computes `limit: Math.max(10_000, expectedPoints)`, so `hours=168` at the default 15s poll interval returns ~40,420 samples — which the MCP tool then `JSON.stringify(data, null, 2)`'s straight into the agent's context. That's megabytes of tokens for one tool result.

The REST path has the same raw output, but a browser parses it into a chart; an LLM pays for every row. `VectorSearchService.getSnapshots` already demonstrates the fix (downsample to ~200 points); alternatively cap rows here like `getTraces` does with `MAX_TRACE_LIMIT`.

jamby77 (commented):
Reviewed the whole #330/#331/#332 stack; posting findings per layer. This layer: error mapping and param clamping in the shared helpers, plus payload size on the history tool.

jamby77 (commented):
One more from the stack review, on the tool-handler pattern.

packages/mcp/src/index.ts
```diff
@@ -1412,16 +1765,115 @@ server.tool(
...
+);
+
+server.tool(
+  'ai_list_instances',
+  'List AI component instances (semantic caches, agent caches, agent memory stores, retrieval pipelines) auto-discovered on the connected Valkey/Redis instance, with liveness and the latest stored metrics sample. This is the superset discovery view across all AI components — use cache_list for cache-specific live stats and memory_stores for memory-store details. Use the returned instance field value as the field parameter of ai_instance_history.',
+  {
+    instanceId: z.string().optional().describe('Optional instance ID override'),
+  },
+  async ({ instanceId }) =>
+    withTelemetry('ai_list_instances', async () => {
```
jamby77, line 1792: Non-blocking, but flagging it while the file is open: this handler body (`resolveInstanceId` → `apiFetch` → `isLicenseError` check → `JSON.stringify` content wrapper) is repeated verbatim in all 10 new tools across the stack, bringing the file to ~34 copies of the same block. The pattern predates this stack (24 copies), but at this size a `jsonApiTool(name, path)` helper returning the standard `ToolResult` would collapse each handler to one or two lines — and it becomes the single place to fix things like the 404 handling `get_largest_keys` had to hand-roll in the next layer. Any change to the license-error contract currently requires ~34 coordinated edits, and one missed site returns raw `__licenseError` JSON to the agent.

apps/api/src/mcp/mcp-helpers.ts
jamby77: Fixed in 9b0c95c2: substring matching is gone — new typed `CapabilityUnavailableError` (apps/api common/errors) is mapped to 501 via `instanceof`; the Search-module throw in VectorSearchService now uses it. Also folded vector-search.controller's private `mapError` into the shared `mapMcpError` (both rethrow HttpException and log the 500 fallback).

apps/api/src/mcp/mcp-helpers.ts
jamby77: Fixed in 9b0c95c2: `clampedParseInt` removed — one `safeLimit(value, default, max)` everywhere, parsing via `Number` + `Math.ceil` so `hours=0.5` becomes 1 instead of silently the default. The zod schemas now declare `.int().min(1).max(168|1000)`, so agents see the ranges and fractions are rejected at the tool boundary.

apps/api/src/mcp/ai/mcp-ai.controller.ts
jamby77: Fixed in 9b0c95c2: the MCP history route now downsamples to ≤200 evenly spaced points (always keeping the newest sample), mirroring `getSnapshots`. REST path unchanged.

packages/mcp/src/index.ts
jamby77: Agreed — deferring this one as a follow-up refactor so the file doesn't end up with two handler styles mid-review; when done it should cover all ~34 handlers and become the single home for the 404/licensing handling. Tracked.

jamby77:
bugbot run

apps/api/src/vector-search/vector-search.service.ts
```diff
@@ -216,7 +246,11 @@ export class VectorSearchService extends MultiConnectionPoller implements OnModu
         const remaining = cappedLimit - allKeys.length;
         const useCursor = i === prefixIdx ? currentCursor : '0';
         const [nc, keys] = await rawClient.scan(
-          useCursor, 'MATCH', `${prefixes[i]}*`, 'COUNT', remaining,
```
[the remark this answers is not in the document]
jamby77: Fixed in 1ea4fd39: sampleKeys' key-browsing throw now uses CapabilityUnavailableError, restoring the 501.

apps/api/src/mcp/mcp-helpers.ts
```diff
@@ -26,9 +37,20 @@ export function safeParseInt(value: string | undefined, defaultValue?: number):
...
+  max = MAX_LIMIT,
+): number {
+  if (value === undefined) {
+    return defaultValue;
+  }
+  const parsed = Number(value);
+  if (Number.isFinite(parsed) === false) {
+    return defaultValue;
+  }
+  return Math.max(1, Math.min(Math.ceil(parsed), max));
```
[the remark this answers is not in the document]
jamby77: Fixed in 1ea4fd39: safeLimit treats empty/whitespace values as absent and returns the caller default (test added).

apps/api/src/mcp/mcp-helpers.ts
```diff
@@ -39,3 +61,14 @@ export function msToSeconds(value: string | undefined): number | undefined {
...
+export function mapMcpError(logger: Logger, error: unknown, fallback: string): HttpException {
+  if (error instanceof HttpException) {
+    return error;
+  }
+  if (error instanceof CapabilityUnavailableError) {
+    return new HttpException(error.message, HttpStatus.NOT_IMPLEMENTED);
+  }
+  logger.error(fallback, error instanceof Error ? error.stack : String(error));
+  return new HttpException(fallback, HttpStatus.INTERNAL_SERVER_ERROR);
+}
```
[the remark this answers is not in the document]
jamby77: Fixed in 1ea4fd39: mapMcpError's 500 now carries `${fallback}: ${message}` again (still logged server-side), restoring the previous REST detail for MCP and vector-search alike.

jamby77:
bugbot run

jamby77:
bugbot run

apps/api/src/mcp/mcp-helpers.ts
```diff
@@ -26,9 +37,20 @@ export function safeParseInt(value: string | undefined, defaultValue?: number):
...
+  defaultValue: number,
+  max = MAX_LIMIT,
+): number {
+  if (value === undefined || value.trim() === '') {
+    return defaultValue;
+  }
+  const parsed = Number(value);
+  if (Number.isFinite(parsed) === false) {
+    return defaultValue;
+  }
```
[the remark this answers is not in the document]
jamby77: Fixed in b6ac2c99: safeLimit now clamps every return path into [1, max], including the default fallback, matching its doc comment.

apps/api/src/mcp/ai/mcp-ai.controller.ts
```diff
@@ -0,0 +1,103 @@
...
+    return samples;
+  }
+  const step = samples.length / maxPoints;
+  const sampled: T[] = [];
+  for (let i = 0; i < maxPoints - 1; i++) {
+    sampled.push(samples[Math.floor(i * step)]);
+  }
+  sampled.push(samples[samples.length - 1]);
+  return sampled;
+}
```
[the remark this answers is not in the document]
jamby77: Fixed in b6ac2c99: extracted a shared downsampleSeries() (common/utils/downsample.ts) used by both the MCP history route and VectorSearchService.getSnapshots — one algorithm (floor-spaced, always keeps the newest sample), no divergent copies.

jamby77:
bugbot run

KIvanow (approved):

PR #332, opened 2026-07-22
feat(mcp): vector search MCP tools

### LINKED ISSUE #329: MCP update

jamby77:
Adds the two vector-search MCP tools completing the MCP update:

- `get_vector_indexes` — per-index health (docs, memory, indexing failures, % indexed); clear 501 when the Search module is absent
- `get_inference_latency` — FT.SEARCH p50/p95/p99 profile per index, merged with per-index SLA breach status when BetterDB Pro inference SLA monitoring is active (`sla: null` otherwise)

Server side: two new routes on `McpAnalyticsController`, plus the plan's only new feature logic — `getSlaStatus(connectionId)` on the pro inference-latency service (license-checked via `Feature.INFERENCE_SLA`, reads the existing per-connection SLA breach state) with a new shared `InferenceSlaIndexStatus` type. Covered by 7 unit tests including the breach→recovery transition.

Stacked on #331. Closes #329.

https://claude.ai/code/session_01BeJC3cWSEXFGD7MLMXaCiH

---

> [!NOTE]
> **Medium Risk**
> Touches Pro inference SLA evaluation and alerting consistency (shared `isBreachActive`), plus new agent-authenticated MCP routes; behavior is heavily unit-tested but threshold/stale rules affect live breach reporting.
> 
> **Overview**
> Adds **vector search** and **FT.SEARCH inference latency** to the MCP agent surface: new tools `get_vector_indexes` and `get_inference_latency` call two routes on `McpAnalyticsController`, wired through `VectorSearchModule` and `InferenceLatencyModule`.
> 
> **`GET …/vector-indexes`** lists index names and expands each to health via `getIndexInfo`, tolerating partial failures (warn + partial list) but failing if every lookup fails; missing Search module maps to **501** via `CapabilityUnavailableError`.
> 
> **`GET …/inference-latency`** returns the latency profile plus optional **`sla`**: merged from Pro’s new **`getSlaStatus(connectionId)`** when `INFERENCE_LATENCY_PRO_SERVICE` is present (otherwise `sla: null`), with **`windowMs`** validated as a positive integer.
> 
> Pro side exposes per-index SLA status (`InferenceSlaIndexStatus` on the shared interface) with license gating on `Feature.INFERENCE_SLA`. SLA state/evaluation gains **`isBreachActive`**, stale expiry after 15 minutes without samples, and re-evaluation against the **current** threshold so Prometheus carry-forward and MCP status stay aligned; **`evaluateSla`** tracks `lastP99Us` / `lastEvaluatedAt` and adjusts debounce when thresholds change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d304969047178b0864359d2a5e04cea0f85eb6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/api/src/mcp/mcp-analytics.controller.ts
+++ b/apps/api/src/mcp/mcp-analytics.controller.ts
@@ -60,13 +60,24 @@ export class McpAnalyticsController {
           return this.vectorSearchService.getIndexInfo(id, name);
         }),
       );
-      const indexes = settled
-        .filter((result): result is PromiseFulfilledResult<VectorIndexInfo> => {
-          return result.status === 'fulfilled';
-        })
-        .map((result) => {
-          return result.value;
+      const indexes: VectorIndexInfo[] = [];
+      const failedNames: string[] = [];
+      settled.forEach((result, index) => {
+        if (result.status === 'fulfilled') {
+          indexes.push(result.value);
+          return;
+        }
+        failedNames.push(names[index]);
+      });
+      if (names.length > 0 && indexes.length === 0) {
+        const failures = settled.filter((result): result is PromiseRejectedResult => {
+          return result.status === 'rejected';
         });
+        throw failures[0].reason;
+      }
+      if (failedNames.length > 0) {
+        this.logger.warn(`Failed to get info for vector indexes: ${failedNames.join(', ')}`);
+      }
       return { indexes };
     } catch (error) {
       throw mapMcpError(this.logger, error, 'Failed to get vector indexes');
--- a/apps/api/src/mcp/mcp-analytics.controller.ts
+++ b/apps/api/src/mcp/mcp-analytics.controller.ts
@@ -1,20 +1,42 @@
-import { Controller, Get, Logger, Param, UseGuards } from '@nestjs/common';
-import { MetricKind } from '@betterdb/shared';
+import {
+  Controller,
+  Get,
+  HttpException,
+  HttpStatus,
+  Inject,
+  Logger,
+  Optional,
+  Param,
+  Query,
+  UseGuards,
+} from '@nestjs/common';
+import { INFERENCE_LATENCY_PRO_SERVICE, IInferenceLatencyProService, MetricKind } from '@betterdb/shared';
 import { AgentTokenGuard } from '../common/guards/agent-token.guard';
-import { ValidateInstanceIdPipe, mapMcpError } from './mcp-helpers';
+import { ValidateInstanceIdPipe, mapMcpError, safeParseInt } from './mcp-helpers';
 import { MetricForecastingService } from '../metric-forecasting/metric-forecasting.service';
 import { VectorSearchService } from '../vector-search/vector-search.service';
+import {
+  InferenceLatencyService,
+  InferenceLatencyValidationError,
+} from '../inference-latency/inference-latency.service';
 import { MetricKindValidationPipe } from '../metric-forecasting/pipes/metric-kind-validation.pipe';
 
 @Controller('mcp')
 @UseGuards(AgentTokenGuard)
 export class McpAnalyticsController {
   private readonly logger = new Logger(McpAnalyticsController.name);
+  private readonly inferenceLatencyProService: IInferenceLatencyProService | null;
 
   constructor(
+    private readonly inferenceLatencyService: InferenceLatencyService,
     private readonly vectorSearchService: VectorSearchService,
     private readonly metricForecastingService: MetricForecastingService,
-  ) {}
+    @Optional()
+    @Inject(INFERENCE_LATENCY_PRO_SERVICE)
+    inferenceLatencyProService?: IInferenceLatencyProService,
+  ) {
+    this.inferenceLatencyProService = inferenceLatencyProService ?? null;
+  }
 
   @Get('instance/:id/forecast/:metricKind')
   async getForecast(
@@ -40,6 +62,28 @@ export class McpAnalyticsController {
       return { indexes };
     } catch (error) {
       throw mapMcpError(this.logger, error, 'Failed to get vector indexes');
+    }
+  }
+
+  @Get('instance/:id/inference-latency')
+  async getInferenceLatency(
+    @Param('id', ValidateInstanceIdPipe) id: string,
+    @Query('windowMs') windowMs?: string,
+  ) {
+    try {
+      const profile = await this.inferenceLatencyService.getProfile(id, {
+        windowMs: safeParseInt(windowMs),
+      });
+      const sla =
+        this.inferenceLatencyProService === null
+          ? null
+          : this.inferenceLatencyProService.getSlaStatus(id);
+      return { profile, sla };
+    } catch (error) {
+      if (error instanceof InferenceLatencyValidationError) {
+        throw new HttpException(error.message, HttpStatus.BAD_REQUEST);
+      }
+      throw mapMcpError(this.logger, error, 'Failed to get inference latency profile');
     }
   }
 }
--- a/apps/api/src/mcp/mcp-analytics.controller.ts
+++ b/apps/api/src/mcp/mcp-analytics.controller.ts
@@ -3,6 +3,7 @@ import { MetricKind } from '@betterdb/shared';
 import { AgentTokenGuard } from '../common/guards/agent-token.guard';
 import { ValidateInstanceIdPipe, mapMcpError } from './mcp-helpers';
 import { MetricForecastingService } from '../metric-forecasting/metric-forecasting.service';
+import { VectorSearchService } from '../vector-search/vector-search.service';
 import { MetricKindValidationPipe } from '../metric-forecasting/pipes/metric-kind-validation.pipe';
 
 @Controller('mcp')
@@ -10,7 +11,10 @@ import { MetricKindValidationPipe } from '../metric-forecasting/pipes/metric-kin
 export class McpAnalyticsController {
   private readonly logger = new Logger(McpAnalyticsController.name);
 
-  constructor(private readonly metricForecastingService: MetricForecastingService) {}
+  constructor(
+    private readonly vectorSearchService: VectorSearchService,
+    private readonly metricForecastingService: MetricForecastingService,
+  ) {}
 
   @Get('instance/:id/forecast/:metricKind')
   async getForecast(
@@ -21,6 +25,21 @@ export class McpAnalyticsController {
       return await this.metricForecastingService.getForecast(id, metricKind);
     } catch (error) {
       throw mapMcpError(this.logger, error, 'Failed to get forecast');
+    }
+  }
+
+  @Get('instance/:id/vector-indexes')
+  async getVectorIndexes(@Param('id', ValidateInstanceIdPipe) id: string) {
+    try {
+      const names = await this.vectorSearchService.getIndexList(id);
+      const indexes = await Promise.all(
+        names.map((name) => {
+          return this.vectorSearchService.getIndexInfo(id, name);
+        }),
+      );
+      return { indexes };
+    } catch (error) {
+      throw mapMcpError(this.logger, error, 'Failed to get vector indexes');
     }
   }
 }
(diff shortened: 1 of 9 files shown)


jamby77:
bugbot run

jamby77:
bugbot run

jamby77:
bugbot run

proprietary/inference-latency-pro/inference-latency-pro.service.ts
```diff
@@ -147,4 +152,27 @@ export class InferenceLatencyProService implements IInferenceLatencyProService {
...
+    const statuses: InferenceSlaIndexStatus[] = [];
+    for (const [indexName, entry] of Object.entries(slaConfig)) {
+      if (entry?.enabled !== true) {
+        continue;
+      }
+      const prior = this.slaState.get(`${connectionId}|${indexName}`);
+      statuses.push({
+        indexName,
+        thresholdUs: entry.p99ThresholdUs,
+        breached: prior !== undefined && prior.resolved === false,
```
jamby77: This pairs a *stale* breach flag with the *live* threshold. `evaluateSla` only resolves a breach when a later tick contains a bucket for that index with p99 back under threshold — but an index whose FT.SEARCH traffic stops while breached (or that gets dropped) never produces another bucket, and `onProfileTick`'s carry-forward preserves `resolved: false` indefinitely (cleanup only removes entries whose config is deleted/disabled). Meanwhile `thresholdUs: entry.p99ThresholdUs` is read from current settings.

Concrete case: breach at p99=200µs vs threshold 100; traffic stops; operator raises the threshold to 500 → every `get_inference_latency` call reports `{ thresholdUs: 500, breached: true }` — a breach the displayed threshold would never trigger — until restart. The new spec covers recovery-with-a-bucket but not the quiet-index or threshold-change cases. Options: expire breach state after N bucket-less ticks, re-evaluate against current config in `getSlaStatus`, or at least return `lastEvaluatedAt`/the threshold-at-breach so consumers can tell the flag is stale.

apps/api/src/mcp/mcp-analytics.controller.ts
```diff
@@ -69,4 +98,59 @@ export class McpAnalyticsController {
...
+  }
+
+  @Get('instance/:id/inference-latency')
+  async getInferenceLatency(
+    @Param('id', ValidateInstanceIdPipe) id: string,
+    @Query('windowMs') windowMs?: string,
+  ) {
+    try {
+      const profile = await this.inferenceLatencyService.getProfile(id, {
+        windowMs: safeParseInt(windowMs),
```
jamby77: `windowMs` is silently coerced rather than validated: `safeParseInt` maps garbage to `undefined` (service default) and passes negatives/tiny values through to `resolveProfileWindow`'s `Math.max(1_000, windowMs)`. An agent passing `windowMs: 60` (thinking seconds) gets a 1-second profile with near-zero samples, returned as if the request was honored — "vector search looks idle".

Relatedly, the `InferenceLatencyValidationError` catch below is unreachable: `getProfile` only throws it for `startTime`/`endTime` mismatches, and this endpoint never passes those. The spec's 400 test only passes because the mock throws. Either validate `windowMs` here (reject non-positive with a 400) or drop the dead catch and let the schema constrain it.

jamby77 (commented):
Stack review, #332 layer: a stale-state exposure in the new SLA status query and a dead validation path on the inference-latency endpoint.

apps/api/src/mcp/mcp-analytics.controller.ts
```diff
@@ -69,4 +98,59 @@ export class McpAnalyticsController {
...
+      });
+      const sla =
```
jamby77, line 108: `sla: null` is now the third distinct response shape this stack uses for the same condition ("Pro capability absent"): latency-regressions returns 200 with `{ events: [], note: '...requires BetterDB Pro' }`, key-analytics surfaces a 404 the client rewrites, and this endpoint returns a bare `null` — while the established license path elsewhere is a 402 `__licenseError` payload that packages/mcp already renders with tier + upgrade URL.

Consequences: an agent can't distinguish "no SLA configured" from "feature not installed" from "not licensed" without parsing free text (or nothing at all, here), and the `isLicenseError` branches in the new McpAnalyticsController-backed tools are dead code since these endpoints never emit the 402 shape. Worth converging on one contract — the 402 `__licenseError` route via `LicenseGuard` seems like the natural one, and it's what the same stack already does for key-analytics. Fine as a follow-up across the three tools rather than a blocker here.

jamby77 (commented):
One more from the stack review, on the capability-absent contract.

proprietary/inference-latency-pro/inference-latency-pro.service.ts
jamby77: Fixed in 6b39894a: `SlaState` now records `lastP99Us` on every evaluated tick, and `getSlaStatus` re-evaluates `breached` against the CURRENT threshold (`resolved === false && lastP99Us > thresholdUs`) instead of trusting the frozen flag. Your concrete case (breach at 200µs, threshold raised to 500) now reports `breached: false` — pinned by a new test. `lastP99Us` is also exposed in the status (alongside lastFiredAt) so consumers can judge staleness for the quiet-index case.

apps/api/src/mcp/mcp-analytics.controller.ts
jamby77: Fixed in 6b39894a: `windowMs` is validated before the service call — non-integer or non-positive returns 400 (`windowMs: 60` meaning seconds now fails loudly instead of returning a 1s profile as if honored… well, 60 is a positive integer so that one still clamps to 1000ms server-side, but the tool schema now declares `.int().min(1000)` so agents can't send it). The unreachable `InferenceLatencyValidationError` catch and its mock-based 400 test are gone, replaced with real validation tests.

apps/api/src/mcp/mcp-analytics.controller.ts
jamby77: Mostly converged by the #331 licensing fix: anomalies/latency-regressions now use the standard 402 `__licenseError` path via LicenseGuard (same as key-analytics), and 404 is reserved for bundle-absent/version-skew with one shared client wording — so the graceful-note shape is gone. The remaining `sla: null` is the merged community+pro endpoint, where a 402 would wrongly fail the community profile half; the tool description now spells out null (absent/unlicensed) vs [] (active, nothing configured). Happy to revisit in the follow-up if you want strict convergence there too.

jamby77:
bugbot run

proprietary/inference-latency-pro/inference-latency-pro.service.ts
```diff
@@ -147,4 +152,30 @@ export class InferenceLatencyProService implements IInferenceLatencyProService {
...
+    const settings = this.settingsService.getCachedSettings();
+    const slaConfig = settings.inferenceSlaConfig ?? {};
+    const statuses: InferenceSlaIndexStatus[] = [];
+    for (const [indexName, entry] of Object.entries(slaConfig)) {
+      if (entry?.enabled !== true) {
+        continue;
+      }
+      const prior = this.slaState.get(`${connectionId}|${indexName}`);
+      const breached =
+        prior !== undefined && prior.resolved === false && prior.lastP99Us > entry.p99ThresholdUs;
```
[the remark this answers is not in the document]
jamby77: Fixed in 8704dab8: one shared isBreachActive(prior, currentThreshold, now) rule now drives BOTH the Prometheus quiet-index carry-forward and getSlaStatus, so the two surfaces cannot disagree on threshold changes.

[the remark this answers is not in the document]
jamby77: Fixed in 8704dab8: breach state with no fresh FT.SEARCH samples for >15 min (SLA_STATE_STALE_MS, tracked via lastEvaluatedAt) is treated as expired rather than live, on both surfaces — quiet/dropped indexes no longer report a stuck breach (test added).

jamby77:
bugbot run

jamby77:
bugbot run

jamby77:
bugbot run

proprietary/inference-latency-pro/sla.ts
```diff
@@ -18,10 +20,39 @@ export interface EvaluateSlaResult {
...
+  now: number,
+): boolean {
+  if (prior === undefined || prior.resolved === true) {
+    return false;
+  }
+  if (now - prior.lastEvaluatedAt > SLA_STATE_STALE_MS) {
+    return false;
+  }
+  return prior.lastP99Us > thresholdUs;
+}
```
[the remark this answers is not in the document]
jamby77: Fixed in b9835bc5: isBreachActive no longer consults `resolved` — status is purely last-observed p99 vs current threshold (plus the staleness window), so tightening the ceiling below the last recorded p99 re-flags the breach, symmetric with the raise case. `resolved` keeps driving only webhook re-fire debouncing in evaluateSla. New test covers the tighten-after-recovery case (32/32 green).

jamby77:
bugbot run

packages/shared/src/types/inference-latency.ts
```diff
@@ -33,6 +33,14 @@ export interface InferenceSlaEntry {
 
 export type InferenceSlaConfig = Record<string, InferenceSlaEntry>;
 
+export interface InferenceSlaIndexStatus {
+  indexName: string;
+  thresholdUs: number;
+  breached: boolean;
+  lastFiredAt: number | null;
+  lastP99Us: number | null;
+}
```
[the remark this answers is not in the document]
jamby77: Fixed in 2981c4f6: InferenceSlaIndexStatus now carries lastEvaluatedAt, so an expired-stale entry (breached: false with lastP99Us above threshold) is self-explanatory from the payload.

jamby77:
bugbot run

jamby77:
bugbot run

jamby77:
bugbot run

proprietary/inference-latency-pro/sla.ts
```diff
@@ -33,7 +68,12 @@ export function evaluateSla(params: EvaluateSlaParams): EvaluateSlaResult {
   }
 
   if (!prior) {
-    state.set(key, { lastFiredAt: now, resolved: false });
+    state.set(key, {
+      lastFiredAt: now,
+      resolved: false,
+      lastP99Us: currentP99Us,
+      lastEvaluatedAt: now,
+    });
```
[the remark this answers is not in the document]
jamby77: Fixed in the vector-branch tip: evaluateSla now applies the same re-evaluation rule as isBreachActive to the debounce state — when the current threshold clears the last observation (e.g. a raised ceiling), resolved flips true, so a later sample breaching the NEW threshold takes the re-fire path and dispatches the webhook immediately. Two new tests: fresh-episode fires; unchanged-threshold debounce still holds (34/34 green).

jamby77:
bugbot run

KIvanow (approved):

PR #334, opened 2026-07-23
feat(anomaly): detect client-eviction storms (maxmemory-clients)

KIvanow:
> 🔗 **Upstream Valkey issue:** [valkey-io/valkey#4151](https://github.com/valkey-io/valkey/issues/4151) — Client eviction can disconnect more clients than necessary

## What

Adds a new **client-eviction storm** detector to the anomaly engine. When Valkey/Redis evicts clients to reclaim client-buffer memory (`maxmemory-clients`), those disconnects are silent today. Upstream **[valkey#4151](https://github.com/valkey-io/valkey/issues/4151)** reports the eviction can be *over-aggressive*, disconnecting healthy clients while memory is nowhere near the limit — this makes that visible.

## How

`EVICTED_CLIENTS` detector, modelled on the existing `REJECTED_CONNECTIONS` delta pattern:

- Feeds the **per-poll delta** of the lifetime `INFO stats evicted_clients` counter to a `SpikeDetector`. Any eviction is abnormal, so absolute thresholds are low (warn ≥1, crit ≥10 per poll).
- `max(0, cur - last)` absorbs the counter reset on restart (no false spike).
- Skipped entirely when eviction is disabled (`maxmemory-clients=0`), read once via `CONFIG GET` and cached per connection.
- The alert message surfaces the configured limit and current client-buffer memory (`mem_clients_normal`) so an operator can distinguish a **justified** eviction (real memory pressure) from an **over-aggressive** one (headroom below the limit — the [valkey#4151](https://github.com/valkey-io/valkey/issues/4151) case).
- Emits over webhook + OTLP for free via `addAnomaly`, and shows in the anomaly feed with a **"Client Evictions"** label.

## Test plan

- `npx jest anomaly.service` → **130 passed**. New cases: delta-fires-CRITICAL (after baseline warms), disabled-at-0 (no buffer armed), counter-reset-on-restart (no false fire).
- `tsc` clean on `apps/api` + `apps/web`.

Closes the monitor side of [valkey#4151](https://github.com/valkey-io/valkey/issues/4151).

---

> [!NOTE]
> **Low Risk**
> Additive monitoring on the poll path with conservative defaults when config is unreadable; no auth or data-path changes, covered by new unit tests.
> 
> **Overview**
> Adds **client-eviction storm** monitoring for Valkey `maxmemory-clients` disconnects (valkey#4151), surfacing spikes that were previously invisible in the anomaly feed.
> 
> The engine introduces `MetricType.EVICTED_CLIENTS` and, on each poll, feeds a **per-interval delta** of `INFO stats evicted_clients` into a lazily created `SpikeDetector` (warn ≥1, critical ≥10 per poll), with `max(0, cur − last)` so restarts do not false-alert. Detection runs only when `maxmemory-clients` is not `0` (`CONFIG GET` cached per connection); connection teardown clears the delta/limit maps. Alerts replace the default message with operator context: configured limit, post-eviction `mem_clients_normal` (with a caveat that it understates pre-eviction pressure), and guidance on justified vs over-aggressive eviction.
> 
> The anomaly dashboard maps `evicted_clients` to the **Client Evictions** label. Tests cover spike → CRITICAL, disabled eviction (no buffer), and counter reset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08572bef0f935a0ff2757ec530614fea84a0beef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jamby77 (approved):
Approving — the detector is correct, well-integrated, and mirrors the rejected-connections per-poll-delta pattern faithfully (delta with `max(0, …)` restart protection, lazy buffer, cleanup in `onConnectionRemoved`, addSample-then-detect ordering all check out). One thing I'd like addressed, before merge or as an immediate follow-up:

1. **`maxmemory-clients` is cached forever per connection.** `clientEvictionEnabled()` never refreshes, so if an operator runs `CONFIG SET maxmemory-clients 4gb` after the first poll cached `'0'`, the detector stays permanently disarmed until reconnect — silently missing exactly the storms this feature exists to catch, and contradicting the PR's own "better to arm than to miss" rationale. Same root cause: a transient first-poll `CONFIG GET` failure is cached forever, and `describeClientEviction` prints the stale cached limit in the alert message. The Raft detector already re-reads `cluster-node-timeout` every `RAFT_NODE_TIMEOUT_RECHECK_POLLS` (60) polls — that pattern is a drop-in template here.

Smaller notes:

2. The 30-sample `MetricBuffer.isReady()` warmup means an eviction storm during the first ~30 polls after connect is completely silent — the absolute thresholds never get a chance. Inherited from the rejected-connections pattern, but for a metric whose pitch is "any eviction is abnormal, warn ≥1", a smaller `minSamples` or bypassing the ready-gate for absolute thresholds would fit the intent better.
3. Test gaps: no WARNING-severity case (delta 1–9), no config-read-failure → default-armed path, and the once-only `CONFIG GET` caching behavior is unasserted.
4. New test code has a one-line `for (let i = 0; i < 30; i++) await poll();` and expression-bodied filter callbacks — house style wants braces/multi-line bodies. Also the PR description's test plan says `npx jest`; repo standard is pnpm.

The three tests that are there are well-constructed — the CRITICAL case genuinely verifies the delta (15, not the raw 20) is buffered, and the disabled/counter-reset paths are asserted directly.

PR #335, opened 2026-07-23
feat(otel): unit metadata + decouple OTLP event mirroring from the webhook

KIvanow:
> 🔗 **Upstream Valkey issue:** [valkey-io/valkey#4078](https://github.com/valkey-io/valkey/issues/4078) — OpenTelemetry (OTLP) metrics export

## What

Three OTLP-mirror polish items ([valkey#4078](https://github.com/valkey-io/valkey/issues/4078)) so an **OTLP-only** deployment — collector configured, no webhook — gets complete, well-typed telemetry rather than a subset gated behind the Pro webhook.

## Changes

**1. Unit metadata on mirrored metrics.** `planInstruments` now derives a UCUM `unit` from the prom-client name suffix, and the exporter passes it to `createObservable{Gauge,Counter}`:

| suffix | unit |
|---|---|
| `_bytes` | `By` |
| `_seconds` | `s` |
| `_milliseconds` | `ms` |
| `_ratio` | `1` |
| `_percent` | `%` |
| `_total` / unknown | *(none)* |

The mirrored metric **name is never rewritten** — the unit lives in OTLP metadata, so downstream backends can convert/label. `unit` is omitted from the instrument options entirely when nothing is derivable (an empty-string unit is meaningful noise).

**2. Decouple anomaly OTLP dispatch from the Pro webhook gate.** `addAnomaly` no longer wraps the OTLP emit in `webhookEventsProService.isEnabled()`. OTLP is its own opt-in channel, gated only by the `OTEL_*` env vars inside the dispatcher (which no-ops when disabled). A licensed operator can now ship anomalies to their collector without also configuring a webhook.

**3. Widen event coverage.** `cluster.failover` (PrometheusService) and the `instance.down` / `instance.up` availability edges (HealthService) now mirror to OTLP **on the state transition itself**, independent of webhook config. The availability edges previously only advanced their state when a webhook dispatcher was wired; the transition + OTLP emit now happen regardless, with the webhook dispatch kept as a separate best-effort step.

## Test plan

- `deriveUnit` suffix table + "unit only when derivable" on the exporter (fake meter captures instrument options).
- Anomaly mirror fires with the **Pro webhook disabled** (`isEnabled → false`), via the data-loss Rule A path.
- `prom-otel-bridge`, `otel-metrics-exporter` (**15 passed**) and `anomaly.service` (**128 passed**) suites green; `apps/api` tsc clean.

Addresses [valkey#4078](https://github.com/valkey-io/valkey/issues/4078).

---

> [!NOTE]
> **Medium Risk**
> Changes when operational events are emitted and can increase OTLP traffic for deployments that previously relied on webhook gates; behavior is still bounded by OTEL env config and existing transition logic.
> 
> **Overview**
> Improves **OTLP-only** deployments so telemetry is complete and better typed without depending on Pro webhook configuration.
> 
> **Mirrored metrics** now attach UCUM **unit** metadata derived from Prometheus name suffixes (`_bytes` → `By`, `_seconds` → `s`, etc.); metric names are unchanged and empty units are omitted from OTel instrument options.
> 
> **OTLP event mirroring** is decoupled from the Pro webhook gate: anomaly detection always calls `otelEvents.dispatch` (still no-ops unless `OTEL_*` is set). **Instance up/down** in `HealthService` and **cluster.failover** in `PrometheusService` emit on state transitions regardless of webhook wiring; webhooks remain a separate optional step. Cluster failover state is advanced even when only OTLP fires.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60ca87e52d91255e58a85b2d9777e9fe20331aaa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jamby77 (approved):
Approving — logic traced clean. The health/cluster state-machine restructure preserves edge semantics (one event per transition, no double-fires), OTLP failures can't suppress webhooks (separate try/catch, and `OtelEventDispatcherService.dispatch` is synchronous so the catch is real), and the `@Optional()` injection resolves since `OtelEventsModule` is `@Global()`. Two follow-ups worth doing:

1. **Test gap on the riskiest part.** The HealthService `instance.down`/`instance.up` and PrometheusService `cluster.failover` transitions now advance without webhook wiring, and neither path has coverage — `health.service.spec.ts` never touches `handleInstanceDown`/`handleInstanceUp`, and there's no PrometheusService unit spec at all. A future refactor that reintroduces the webhook guard would silently kill OTLP-only availability events again (exactly the regression class this PR fixes) with nothing going red.

2. **`deriveUnit()` misses `_seconds_total`.** The suffix match bails on `_total` before deriving, so `betterdb_cpu_sys_seconds_total` / `betterdb_cpu_user_seconds_total` get no unit even though they measure seconds. The OTel Prometheus convention derives from the token preceding `_total` — checking `endsWith('_seconds_total')` before the `_total` bail-out (or stripping `_total` first) fixes it. Same family: `_us` (e.g. `betterdb_commandstats_latency_us`) is unmapped, though I see the PR scopes to unambiguous suffixes deliberately.

The bridge/exporter half is well-tested (suffix table, unknown/`_total` cases, exporter omitting the key when underivable, and the anomaly mirror firing with `isEnabled() → false`).

PR #336, opened 2026-07-23
feat(key-analytics): composite multi-dimensional hot/big-key detection

KIvanow:
> 🔗 **Upstream Valkey issue:** [valkey-io/valkey#4189](https://github.com/valkey-io/valkey/issues/4189) — Multi-dimensional hot-key & big-key detection

## What

Adds a **composite ("hot big key") signal** to key analytics ([valkey#4189](https://github.com/valkey-io/valkey/issues/4189)). The existing lists rank each dimension independently — hot keys (by access frequency or recency) and `cardinality` (big by element count / byte length) — so a key that is *only* hot or *only* big is already covered. What none of them surface is the key extreme on **both at once**: a large collection that is also hammered, or a hot key whose value has quietly grown huge. Those are the ones that cause real incidents — bandwidth amplification, single-shard hotspots, O(N) commands on a big value under load.

## How (Phase A — no extra scanning, no migration)

A pure post-processing pass over the per-key data already collected in one scan:

- Take the per-dimension top-N of **hotness** and of **cardinality**; a key landing in **both** is a composite.
- Hotness is LFU access frequency (`OBJECT FREQ`) when available, else idle recency (`OBJECT IDLETIME`) — mirroring the hot-keys collector's fallback, so it works on the default (non-LFU) policy too.
- Rank by the sum of normalized `value / dimension-max` contributions, so the most extreme rise to the top.
- **Memory is context, not a ranking dimension:** the collector only retains a global top-N by LFU / idletime / cardinality, so a "memory top-N" would be computed over a biased subset. `memoryBytes` is reported on each entry as context; cardinality already captures "big" across types (element count for collections, byte length for strings).

The ranker (`composite-key-ranker.ts`) is SDK-free and side-effect-free. Results persist as a new `'composite'` `signalType` **reusing the existing `hot_key_stats` columns** (`signal_type` is unconstrained `TEXT`, so no schema migration). `getCompositeKeys` adds a **freshness guard** so an empty composite result in the newest scan does not keep serving the previous batch under the default `latest` view.

## API

- `GET /key-analytics/composite-keys` (mirrors `hot-keys` / `largest-keys`: `limit`, `startTime`, `endTime`, `latest`, `oldest`).
- `KeyAnalyticsService.getCompositeKeys()` accessor.
- Web `keyAnalyticsApi.getCompositeKeys()` client, ready for a follow-up UI tab (the frontend surface is intentionally a separate PR, as with the Raft panel).

## Test plan

- `composite-key-ranker.spec.ts` (**6**): the both-dimensions gate, idle-recency hotness fallback, score ordering, single-dimension exclusion, non-positive/NaN handling, empty-input / zero-cutoff guards.
- `key-analytics.service.spec.ts` (**3**): the composite freshness guard (fresh / stale-after-empty-scan / time-range).
- `apps/api` and `apps/web` tsc clean.

Implements Phase A of [valkey#4189](https://github.com/valkey-io/valkey/issues/4189).

---

> [!NOTE]
> **Low Risk**
> Additive analytics feature on existing storage; main behavioral risk is read-path freshness logic for empty composite scans, which is covered by tests.
> 
> **Overview**
> Introduces **composite ("hot big key")** detection for key analytics: keys that land in the per-dimension top-N for both **hotness** (LFU or idle recency) and **cardinality**, ranked by normalized scores. Results are stored as `signalType: 'composite'` using existing hot-key storage (no migration).
> 
> **Collection** now derives composites from the same scan `keyDetails` via `rankCompositeKeys`, and **persists hot, largest, and composite rows in a single `saveHotKeys` call** so partial writes cannot confuse the read path.
> 
> **API:** `GET /key-analytics/composite-keys` (same query params as hot/largest keys), `KeyAnalyticsService.getCompositeKeys()` with a **freshness guard** for connection-scoped `latest` queries (empty composite scans no longer serve stale batches), and web `keyAnalyticsApi.getCompositeKeys()`. Shared `HotKeyEntry` types document the new signal.
> 
> Unit tests cover the ranker edge cases (LFU-cold zero, idle fallback) and service behavior (freshness guard, atomic persistence).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ce0208f68fd69ac1cf39f497deae1bab6a033f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jamby77 (commented):
No functional bug found in the new code itself — the ranker logic, freshness-guard semantics, and endpoint mirroring all trace clean — but the PR can't merge as-is:

1. **[BLOCKER] Unmergeable — needs a rebase onto master.** GitHub reports `mergeable: CONFLICTING`. The merge-base predates #331, which rewrote `getLargestKeys` from a pass-through into the memory-re-ranking implementation (`LARGEST_KEYS_FETCH_CAP` + `dedupeByKeyMaxMemory`) at exactly the spot where this PR inserts `getCompositeKeys` (the diff's context lines still show the old body). After rebasing, please re-run the service spec — `getLargestKeys` changed underneath it.

2. **[MINOR] Freshness guard reads aren't a consistent snapshot.** In `getCompositeKeys`, if a collection commits between the composite read and the `newestAny` read, `composite[0].capturedAt < latestCollectionAt` is true even though the new scan produced composites, so that one request returns `[]` despite fresh data. Transient and self-healing next request, but worth either a re-read of composites when the guard trips or a note.

3. **[MINOR] Idle-fallback docstring overstates its invariant.** `hasFrequencySignal` returns false for NaN `freqScore`, so a NaN-freq key with idle data does fall back to idle — the `nan-hot` test only passes because its fixture has `idleSeconds: null` (its "no idle fallback" comment is misleading). And when a scan yields mixed freq/idle candidates, both magnitudes normalize against the same `dimensionMax` (e.g. 255), deflating idle-derived contributions to near zero in the composite score. Neither breaks the top-N gate, but the doc/test comments should match the actual behavior.

4. **[MINOR] `as any` in the new spec.** `key-analytics.service.spec.ts` lines 17-19 cast `storage`/`registry`/`license` with `as any`, against the repo's no-`any` rule; the sibling `key-analytics-largest-keys.spec.ts` on master shows the sanctioned `as unknown as StoragePort` pattern.

5. **[MINOR] Post-#331 design wrinkle worth a note.** After the rebase, "largest keys" ranks by `memoryBytes` while the composite "big" dimension remains cardinality — a hot, memory-huge but low-cardinality value tops largest-keys yet never qualifies as composite. Your rationale for cardinality (memory top-N over the pruned pool is biased) is sound and documented, but the two surfaces now disagree on what "big" means; an explicit note or a Phase-B follow-up issue would help.

Test gaps: (a) the collection path — the composite→`HotKeyEntry` mapping and the single combined atomic `saveHotKeys` call, which is the PR's own atomicity argument — is untested; (b) no adapter-level test that `signalTypes: ['composite']` round-trips through the real SQL `latest` subquery; (c) no controller test for `GET /composite-keys` (consistent with the untested sibling endpoints, so not gating).

Cleared: default `signalTypes: ['lfu','idletime']` prevents composite rows leaking into legacy hot-keys reads; both SQL adapters' `saveHotKeys` are atomic; the shared type widening flows through `storage-port.interface.ts`; no consumer switches exhaustively on `signalType`; `COMPOSITE_TOP_N = KEY_DETAILS_TOP_N` keeps the intersection unbiased over the pruned pool.

KIvanow:
> No functional bug found in the new code itself — the ranker logic, freshness-guard semantics, and endpoint mirroring all trace clean — but the PR can't merge as-is:
> 
> 1. **[BLOCKER] Unmergeable — needs a rebase onto master.** GitHub reports `mergeable: CONFLICTING`. The merge-base predates #331, which rewrote `getLargestKeys` from a pass-through into the memory-re-ranking implementation (`LARGEST_KEYS_FETCH_CAP` + `dedupeByKeyMaxMemory`) at exactly the spot where this PR inserts `getCompositeKeys` (the diff's context lines still show the old body). After rebasing, please re-run the service spec — `getLargestKeys` changed underneath it.
> 2. **[MINOR] Freshness guard reads aren't a consistent snapshot.** In `getCompositeKeys`, if a collection commits between the composite read and the `newestAny` read, `composite[0].capturedAt < latestCollectionAt` is true even though the new scan produced composites, so that one request returns `[]` despite fresh data. Transient and self-healing next request, but worth either a re-read of composites when the guard trips or a note.
> 3. **[MINOR] Idle-fallback docstring overstates its invariant.** `hasFrequencySignal` returns false for NaN `freqScore`, so a NaN-freq key with idle data does fall back to idle — the `nan-hot` test only passes because its fixture has `idleSeconds: null` (its "no idle fallback" comment is misleading). And when a scan yields mixed freq/idle candidates, both magnitudes normalize against the same `dimensionMax` (e.g. 255), deflating idle-derived contributions to near zero in the composite score. Neither breaks the top-N gate, but the doc/test comments should match the actual behavior.
> 4. **[MINOR] `as any` in the new spec.** `key-analytics.service.spec.ts` lines 17-19 cast `storage`/`registry`/`license` with `as any`, against the repo's no-`any` rule; the sibling `key-analytics-largest-keys.spec.ts` on master shows the sanctioned `as unknown as StoragePort` pattern.
> 5. **[MINOR] Post-#331 design wrinkle worth a note.** After the rebase, "largest keys" ranks by `memoryBytes` while the composite "big" dimension remains cardinality — a hot, memory-huge but low-cardinality value tops largest-keys yet never qualifies as composite. Your rationale for cardinality (memory top-N over the pruned pool is biased) is sound and documented, but the two surfaces now disagree on what "big" means; an explicit note or a Phase-B follow-up issue would help.
> 
> Test gaps: (a) the collection path — the composite→`HotKeyEntry` mapping and the single combined atomic `saveHotKeys` call, which is the PR's own atomicity argument — is untested; (b) no adapter-level test that `signalTypes: ['composite']` round-trips through the real SQL `latest` subquery; (c) no controller test for `GET /composite-keys` (consistent with the untested sibling endpoints, so not gating).
> 
> Cleared: default `signalTypes: ['lfu','idletime']` prevents composite rows leaking into legacy hot-keys reads; both SQL adapters' `saveHotKeys` are atomic; the shared type widening flows through `storage-port.interface.ts`; no consumer switches exhaustively on `signalType`; `COMPOSITE_TOP_N = KEY_DETAILS_TOP_N` keeps the intersection unbiased over the pruned pool.

The conflicts were resolved and most of the nits were addressed. @jamby77

jamby77 (approved):
Approving — the blocker is resolved and the fixes check out. I diffed the head `key-analytics.service.ts` against master: post-#331 `getLargestKeys` is intact verbatim (`LARGEST_KEYS_FETCH_CAP`, `dedupeByKeyMaxMemory`, rank renumbering), `getCompositeKeys` is layered cleanly after it, and the only deltas vs master are the PR's intended additions. Nothing dropped or duplicated in the merge, CI green, PR mergeable.

Also verified fixed:
- NaN/idle-fallback docs now match actual behavior, with a new test pinning the non-finite-freq → idle-fallback path.
- `as any` casts replaced with the `as unknown as StoragePort` pattern; no `any` remains.
- The largest-vs-composite "big" disagreement now has an explicit docstring including the Phase-B condition (global memory top-N retention).
- The collection path is now covered: single combined `saveHotKeys` call with all three signal groups, composite→`HotKeyEntry` mapping asserted.

Remaining non-gating follow-ups:
1. The freshness-guard snapshot race (two separate `getHotKeys` reads; a collection committing between them returns `[]` once) was the one nit skipped — still worth a re-read-on-trip or at least a note.
2. Adapter-level round-trip of `signalTypes: ['composite']` through the real SQL `latest` subquery is still mock-only; no controller test for `GET /composite-keys` (consistent with siblings).
3. A few new/edited lines exceed the 100-char Prettier width (ranker lines ~101/141, spec fixtures) — worth a `prettier --write` pass before or after merge.

PR #344, opened 2026-07-27
feat(anomaly): detect replicas stuck in slot migrating/importing state (valkey#1664)

KIvanow:
## Summary

Adds an anomaly detector for **[valkey-io/valkey#1664](https://github.com/valkey-io/valkey/issues/1664)**: after a reshard, some replicas get stuck reporting slots in `migrating`/`importing` state (or wrongly owning slots) in `CLUSTER NODES` — a silent, non-self-healing inconsistency whose only known fix is an operator running `CLUSTER SETSLOT <slot> STABLE`. Monitor connects as a client and surfaces it, the same **detect + advise** play as our #2090 stuck-replica and #2261 duplicate-primary detectors.

Detection is two-layer:
- **Layer 1 (`CLUSTER NODES`):** flag a replica carrying `migratingSlots`/`importingSlots` — a replica must never carry slot-migration state.
- **Layer 2 (`CLUSTER SHARDS` cross-view):** role authority (suppresses the mid-promotion false positive where `CLUSTER NODES` flags are briefly stale), shard attribution, and slot-view divergence (a replica that *owns* slots). Degrades to Layer 1 if `CLUSTER SHARDS` is unsupported.

Because migrating/importing markers are **node-local** — only the queried (`myself`) node's own line carries them — detection **fans out per-node** via `ClusterDiscoveryService`, so a replica polled only through the primary (the common single-connection setup) is still seen. Falls back to the connected-node view when cluster discovery is unavailable.

## Changes
- **`replica-slot-state-detector.ts`** (new) — pure `detectReplicaSlotState(nodes, shards?)` + `replicaSlotSignature()`. Divergence signatures on the full range bounds so a stuck range that grows/shrinks (even keeping its start) re-alerts.
- **`CLUSTER SHARDS` plumbing** — new `ClusterShard` type, `MetricsParser.parseClusterShards()` (handles the RESP2 flat-array and RESP3 map reply shapes), and `getClusterShards()` on both `DatabasePort` implementers.
- **`anomaly.service.ts`** — new `MetricType.REPLICA_SLOT_STATE`; per-node fan-out (`gatherReplicaSlotView`); `CLUSTER NODES` fetched **once per poll** and shared across all three topology detectors, with `CLUSTER SHARDS` fetched concurrently; a shared persistence-gate/dedup helper (`applyTopologyPersistenceGate`) used by both stuck-replica and replica-slot; **auto-resolve on recovery** so the WARNING banner clears; operator addresses stripped of the cluster-bus port; per-slot remediation guidance (divergence gets ownership-appropriate advice, not `SETSLOT STABLE`). A transient `CLUSTER NODES` gap no longer clears the WARNING detectors' dedup state (no duplicate alerts).
- **Web** — `ReplicaSlotStateBanner` (WARNING, with a `CLUSTER SETSLOT ... STABLE` runbook and server-confirmed dismiss) + dashboard label and active-feed wiring.
- **`types.ts`** — `METRICS_HANDLED_OUTSIDE_EXTRACTOR` set as the single source of truth for the buffer-exclusion guard and its test.

## Verification
- API + web typecheck clean.
- **API: 185 unit tests pass** — pure detector, `parseClusterShards`, and service-level: persistence gate, `CLUSTER SHARDS` role-authority suppression, degrade path, **per-node fan-out**, **auto-resolve on recovery**, and **no-duplicate-after-transient-failure**. **Web banner: 7 pass.**
- Reviewed at max effort via the multi-agent `/code-review` workflow; all confirmed red/orange/yellow findings fixed in this branch.

## Follow-ups
- Live-cluster integration (`docker-compose.cluster.yml`): confirm `parseClusterShards` against a real `CLUSTER SHARDS` reply and verify no false positives during an actual reshard.
- The upstream fix belongs in the engine (auto-heal); this detector is the interim operator-facing signal.

## Checklist
- [ ] Docs added / updated

https://claude.ai/code/session_01D6cfkxXM4PMNq1ac6JTjzz

---

> [!NOTE]
> **Medium Risk**
> Changes cluster monitoring poll behavior (per-replica fan-out, shared topology fetch) and connection pooling; incorrect detection could miss or false-alert on reshards, but logic is heavily tested and degrades when `CLUSTER SHARDS` or fan-out is unavailable.
> 
> **Overview**
> Adds **replica slot-state** anomaly detection for [valkey#1664](https://github.com/valkey-io/valkey/issues/1664): replicas stuck reporting `migrating`/`importing` slots or wrongly owning slots in `CLUSTER NODES`, with remediation text (`CLUSTER SETSLOT … STABLE` vs `CLUSTER REPLICATE` for ownership divergence).
> 
> **Detection** uses Layer 1 on `CLUSTER NODES` and optional Layer 2 from **`CLUSTER SHARDS`** (new `ClusterShard` types, `parseClusterShards` for RESP2/RESP3/Map replies, `getClusterShards` on database adapters). Because migration markers are **node-local**, `AnomalyService` **fans out** `CLUSTER NODES` to reachable replicas via `ClusterDiscoveryService` (2s command timeout), merges each node’s self-view, and fetches topology **once per poll** for duplicate-primary, stuck-replica, and slot-state detectors. A shared **`applyTopologyPersistenceGate`** handles 30s grace, dedupe, operator dismiss re-alert, auto-resolve on recovery, and preserves state when replicas are unreachable.
> 
> **`ClusterDiscoveryService`** fixes a **connection leak**: ready-but-stale cached clients are reused (stamp refreshed) instead of replaced; replacing a not-ready client **`quit`s** the old socket.
> 
> **Web:** `ReplicaSlotStateBanner` on the anomaly dashboard (active feed, runbook, server-confirmed dismiss). **`METRICS_HANDLED_OUTSIDE_EXTRACTOR`** centralizes metrics excluded from z-score buffers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e97a498bb5d665e6e9cee45c2e4b0923b7169af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jamby77 (commented):
The detection logic and the test suite (~38 tests across detector, parser, service gate, and banner) are genuinely well built. One finding is a blocker for me, the rest are smaller:

1. **[MAJOR] Per-poll fan-out leaks TCP connections to the monitored cluster.** `gatherReplicaSlotView` goes through `ClusterDiscoveryService.getNodeConnection` (`apps/api/src/cluster/cluster-discovery.service.ts:119-230`), which only returns the cached client while `status === 'ready' && Date.now() - lastHealthCheck < 30_000`. Nothing in the fan-out path refreshes `lastHealthCheck` (only `healthCheckNode` does, and that's driven by the Cluster page). Once the cached connection is >30s old, the ready-but-stale case falls through both early returns, creates a **new** `Valkey` connection, and overwrites the map entry — the old, still-`ready` client is never `quit()`. `cleanupIdleConnections` can't reach it (it's out of the map) and `MAX_CONNECTIONS` doesn't help (map size stays constant). With the default `ANOMALY_POLL_INTERVAL_MS=1000`, steady state is ~1 orphaned ESTABLISHED connection per remote replica every 30s (~120/hr/replica) against the customer's database. This PR is the first continuous caller of that path, so it needs either refresh-on-use or reuse/quit-before-replace in `getNodeConnection`, plus a test exercising repeated polls through the real discovery service.

2. **[MINOR] Auto-resolve can retry forever.** In `applyTopologyPersistenceGate`, if an emitted event is evicted from `recentAnomalies` (capped at 1000) and was never durably persisted (the Postgres-UUID-PK case `resolveAnomaly`'s own comment describes for deterministic string ids), `resolveAnomaly(id)` returns false forever, the `eventIds` mapping is never dropped, and the resolve retries every poll — a wasted storage call per stale signature per poll, plus unbounded map retention. Consider dropping the mapping when the event is absent from both cache and storage.

3. **[MINOR] `flatReplyToMap` and RESP3 `Map` replies.** A JS `Map` satisfies `typeof entry === 'object'` but `Object.entries(map)` returns `[]`, so `Map`-shaped replies would silently yield shards with empty `slots`/`nodes` and `hasShards` would disable Layer 2. Degrade is safe (Layer 1 still runs), but an `entry instanceof Map` branch — or covering it in the live-cluster follow-up you already list — would close it.

4. **[MINOR] Silent degradation on unreachable replicas.** Rejected `Promise.allSettled` results in `gatherReplicaSlotView` are dropped with no service-level log, so a persistently unreachable (but not `fail`-flagged) replica leaves detection silently blind for that node, costing up to the 5s connect timeout per poll. The discovery service logs each unique connect error only once, so there's no ongoing signal the detector is degraded.

5. [MINOR] House-style drift concentrated in the new code: one-line conditionals (`if (!shardMap) continue;`, `if (!Array.isArray(raw)) return [];`, several in the gate/fan-out and `ReplicaSlotStateBanner.tsx`) and expression-bodied filter/find callbacks. Existing code in these files has the same style, so noting rather than gating on it.

Things I checked and cleared: `getNodeConnection` call signature matches; `ClusterModule` exports `ClusterDiscoveryService` so the `@Optional()` injection resolves; both `DatabasePort` implementers got `getClusterShards()`; `metric_type` is TEXT so the new enum value needs no migration; `CLUSTER SHARDS` only fires inside the `cluster_enabled === '1' && !isRaft` branch; signature ids survive `encodeURIComponent`; banner styling matches existing precedent.

jamby77 (commented):
All four substantive findings from the previous review are genuinely fixed, with real covering tests — nice turnaround:

- Connection leak: `getNodeConnection` now reuses any `ready` client unconditionally, refreshes `lastHealthCheck` on use, and quits a replaced client before the map overwrite; the regression test seeds a ready-but-stale entry and asserts same-client reuse. (The quit-before-replace path itself is untested — minor.)
- Auto-resolve give-up: `resolveAnomalyOutcome` returns `'gone'` when the event is absent from both cache and storage and the gate drops the mapping; the test asserts the retry count stops growing.
- RESP3 `Map` replies: `entry instanceof Map` branch added, correctly before the plain-object branch, with a nested-Map test.
- Degraded fan-out is now logged (throttled warn + recovery line, state cleaned up in `onConnectionRemoved`).

However, verifying the fixes surfaced issues in the same code path that I'd want addressed before merge (the first two align with what Bugbot flagged):

1. **[HIGH] Fan-out `CLUSTER NODES` has no command timeout and can freeze ALL anomaly detection.** `client.call('CLUSTER', 'NODES')` on the per-replica connection has no timeout race — only *connect* is bounded (5s); compare `healthCheckNode`, which races its ping against 2s. `MultiConnectionPoller.tick()` holds `polling = true` until every `pollConnection` settles and skips all subsequent ticks while set, so a blackholed replica socket (established but unresponsive peer; iovalkey status stays `ready`, and the new reuse-any-ready path hands that client back forever) hangs one poll promise for TCP-retransmit timescales — suspending anomaly detection for every monitored connection. Fix: race the fan-out call against a timeout the way `healthCheckNode` does.

2. **[MEDIUM] A failed fan-out reads as recovery.** A rejected self-view leaves that node on the base gossip view, which never carries node-local migrating/importing markers — so the finding disappears for that poll, the gate deletes `firstSeen` (the 30s grace restarts), and the emitted WARNING is auto-resolved. A still-stuck replica with flaky connectivity flaps its banner or never alerts — exactly the degraded conditions this detector targets. Degradation is logged now, but never fed into the gate. Contrast: a failed base `CLUSTER NODES` fetch skips the detectors and preserves state; per-node fan-out failure should likewise preserve that node's prior state rather than count as recovery.

3. **[MEDIUM] Dismiss/dedupe check consults only the in-memory ring.** If the tracked event is evicted from the 1000-cap `recentAnomalies` while still unresolved in storage, `existing` is undefined, the mapping is dropped, and a duplicate WARNING is emitted while the old row stays unresolved in the storage-backed feed.

4. [NIT] `(service as any).recentAnomalies` in the new give-up test — repo bans `any`; the discovery spec does the same thing with `as unknown as {...}`.

KIvanow:
> All four substantive findings from the previous review are genuinely fixed, with real covering tests — nice turnaround:
> 
> * Connection leak: `getNodeConnection` now reuses any `ready` client unconditionally, refreshes `lastHealthCheck` on use, and quits a replaced client before the map overwrite; the regression test seeds a ready-but-stale entry and asserts same-client reuse. (The quit-before-replace path itself is untested — minor.)
> * Auto-resolve give-up: `resolveAnomalyOutcome` returns `'gone'` when the event is absent from both cache and storage and the gate drops the mapping; the test asserts the retry count stops growing.
> * RESP3 `Map` replies: `entry instanceof Map` branch added, correctly before the plain-object branch, with a nested-Map test.
> * Degraded fan-out is now logged (throttled warn + recovery line, state cleaned up in `onConnectionRemoved`).
> 
> However, verifying the fixes surfaced issues in the same code path that I'd want addressed before merge (the first two align with what Bugbot flagged):
> 
> 1. **[HIGH] Fan-out `CLUSTER NODES` has no command timeout and can freeze ALL anomaly detection.** `client.call('CLUSTER', 'NODES')` on the per-replica connection has no timeout race — only _connect_ is bounded (5s); compare `healthCheckNode`, which races its ping against 2s. `MultiConnectionPoller.tick()` holds `polling = true` until every `pollConnection` settles and skips all subsequent ticks while set, so a blackholed replica socket (established but unresponsive peer; iovalkey status stays `ready`, and the new reuse-any-ready path hands that client back forever) hangs one poll promise for TCP-retransmit timescales — suspending anomaly detection for every monitored connection. Fix: race the fan-out call against a timeout the way `healthCheckNode` does.
> 2. **[MEDIUM] A failed fan-out reads as recovery.** A rejected self-view leaves that node on the base gossip view, which never carries node-local migrating/importing markers — so the finding disappears for that poll, the gate deletes `firstSeen` (the 30s grace restarts), and the emitted WARNING is auto-resolved. A still-stuck replica with flaky connectivity flaps its banner or never alerts — exactly the degraded conditions this detector targets. Degradation is logged now, but never fed into the gate. Contrast: a failed base `CLUSTER NODES` fetch skips the detectors and preserves state; per-node fan-out failure should likewise preserve that node's prior state rather than count as recovery.
> 3. **[MEDIUM] Dismiss/dedupe check consults only the in-memory ring.** If the tracked event is evicted from the 1000-cap `recentAnomalies` while still unresolved in storage, `existing` is undefined, the mapping is dropped, and a duplicate WARNING is emitted while the old row stays unresolved in the storage-backed feed.
> 4. [NIT] `(service as any).recentAnomalies` in the new give-up test — repo bans `any`; the discovery spec does the same thing with `as unknown as {...}`.

These should be addressed now

jamby77 (approved):
Approving — all three findings from the previous round are genuinely fixed, with covering tests and CI green:

- **Command timeout:** the fan-out `CLUSTER NODES` call is now raced against `FANOUT_COMMAND_TIMEOUT_MS = 2_000` via `raceWithTimeout`, the timer is cleared in a `finally`, and a timeout feeds the unreachable path. Both race outcomes are unit-tested.
- **Gap ≠ recovery:** `gatherReplicaSlotView` returns `unreachableIds` (collected positionally from `allSettled`, order-safe), and preserved signatures keep `firstSeen`, skip auto-resolve, and stay in the active set — the test asserts the WARNING stays open with no `resolveAnomaly` call during the gap. Per-connection state is cleaned up in `onConnectionRemoved`.
- **Storage-aware dedupe:** on a ring-eviction miss the gate now checks `getAnomalyEvents({ connectionId, resolved: false, metricType })`, with storage errors conservatively treated as still-open; cost only on the rare miss.

One note on Bugbot's post-fix finding ("timeout leaves unhandled rejection"): it's a **false positive** — `Promise.race` attaches reactions to every input promise, so the abandoned `client.call` rejecting late is handled (verified with a Node repro; no `unhandledRejection` fires). No change needed there.

Small follow-ups, none gating:

1. **Residual door in the gap≠recovery fix:** replicas filtered out by `DEAD_FLAGS` (`fail?`/`fail`/`noaddr`/`handshake`) are never attempted, so they never enter `unreachableIds` and `preserveSignature` doesn't match them — sustained flakiness that trips gossip PFAIL (`fail?`) still flaps/suppresses the alert through this path. One-line fix: add dead-flagged replicas' ids to `unreachableIds` (at least for transient `fail?`).
2. A timed-out client stays `ready` in the discovery map with the stuck command still queued, so subsequent polls stack pending commands on the blackholed socket until TCP teardown — bounded now (each await is 2s), but `client.disconnect()` on timeout would be cleaner.
3. The `as any` test nit from last round wasn't addressed and the new tests add three more (`(service as any).clusterDiscovery`, `(service as any).recentAnomalies`); line ~3168 in the same spec shows the sanctioned `as unknown as {...}` pattern.
4. Commit 4e97a498's message carries LLM attribution trailers — repo rule says never; please make sure the squash-merge message is clean.
5. Untested minor branches of the new dedupe: resolved-in-storage → re-emit, and the storage-error path.

PR #362, opened 2026-08-06
fix(metrics): parse keyspace/commandstats/errorstats in parseInfoToTyped

### LINKED ISSUE #360: Health summary always reports keyspaceSize 0: summation expects parsed objects but InfoParser stores raw strings
### Summary

`GET`ting the health summary always reports `keyspaceSize: 0`, regardless of how many keys the instance holds. On an instance where `INFO keyspace` reports `db0:keys=568,expires=310,...`, the health summary on the same connection returns `keyspaceSize: 0` while every other field (hit rate, fragmentation, connected clients) tracks live values correctly.

Reproduced on `@betterdb/monitor` 0.32.0; the same code is present in 0.34.0 (latest at time of filing).

### Root cause

The keyspace summation in `getHealthSummary` and the INFO parser disagree about the shape of the keyspace section.

`InfoParser.parse` stores every INFO line as a raw string:

```js
// InfoParser.parse — each line becomes sections[section][key] = <raw string>
i[l][h] = g;   // e.g. keyspace.db0 = "keys=568,expires=310,avg_ttl=7510966104"
```

(`MetricsParser.parseInfoToTyped` is an identity function, so nothing re-shapes it afterwards.)

But `getHealthSummary` only counts a db entry if its value is an **object** with a `keys` property:

```js
let R = 0;
if (g) {
  for (const [r, i] of Object.entries(g)) {
    if (r.startsWith("db") && i && typeof i === "object" && "keys" in i) {
      R += Number(i.keys) || 0;
    }
  }
}
```

Since every value is a string, `typeof i === "object"` is never true, the loop adds nothing, and `keyspaceSize` is structurally always `0` in every deployment.

### Why it matters

A hard `0` here reads as "empty database", which is a plausible and very wrong diagnosis — it's the first summary field an operator (or an LLM agent using the MCP tools, whose `get_health` description recommends it as the first call) reaches for when checking whether a cache is populated.

### Suggested fix

Parse the db line before summing — the codebase already has a suitable helper:

```js
if (r.startsWith("db") && typeof i === "string") {
  const kv = InfoParser.parseKvLine(i, ",");   // { keys: "568", expires: "310", ... }
  R += Number(kv.keys) || 0;
}
```

(or use `DBSIZE`, which the client already wraps as `getDbSize()`).

Additionally, consider reporting `null` rather than `0` when the keyspace section is absent entirely, so "no data" and "empty database" stay distinguishable.

KIvanow:
Fixes #360

## Problem

`MetricsParser.parseInfoToTyped` was an identity cast (`return info as InfoResponse`), so the sections whose types promise parsed objects — `KeyspaceInfo`, `CommandStatsInfo`, `ErrorStatsInfo` — actually held raw `"k=v,k=v"` strings at runtime. `getHealthSummary` only counts a db entry when `typeof val === 'object' && 'keys' in val`, which was never true, so the health summary (and the MCP `get_health` tool built on it) reported `keyspaceSize: 0` on every instance, indistinguishable from an empty database.

## Fix

- `parseInfoToTyped` now genuinely parses the three k=v-line sections via the existing `InfoParser.parseKvLine` helper (whose doc comment already named these as intended use cases). The transform is idempotent: only string values under matching keys (`db\d+`, `cmdstat_*`, `errorstat_*`) are converted; everything else passes through.
- Scalar sections (server/clients/memory/stats/replication/cpu/cluster/persistence) intentionally **stay strings** — their declared types say `string`, and consumers rely on string compares (`cluster_enabled === '1'`, `role === 'master'`, `master_link_status === 'up'`).
- `getHealthSummary` reports `keyspaceSize: null` when the keyspace section is absent, so "no data" and "empty database" stay distinguishable (as suggested in the issue).
- Web `InfoResponse` type updated to the object shape (type-only; no web code reads `info.keyspace`).

## Consumer audit (why this doesn't break anything)

- **Prometheus exporter**: `updateKeyspaceMetricsFromInfo` already handled both string and object shapes; the object branch simply becomes the live path. Verified `betterdb_db_keys` gauges unchanged against a live instance.
- **Web app**: zero consumers of `info.keyspace`; all scalar reads use `parseInt`/`parseFloat` or direct render.
- **MCP `get_info`**: pure JSON passthrough.
- **commandstats/errorstats pollers**: use raw `getInfo`, not `getInfoParsed` — untouched.
- **API contract note (release notes)**: `GET /metrics/info` and `GET /mcp/instance/:id/info` now serialize keyspace/commandstats/errorstats as objects instead of raw strings, and `keyspaceSize` in the health payload is `number | null`.

## Tests

- New `metrics.service.spec.ts` — regression guard for #360: sums across dbs, `0` for present-but-empty keyspace, `null` when absent, other summary fields unchanged, replica lag.
- `metrics.parser.spec.ts` — first coverage for `parseInfoToTyped`: all three sections, scalars untouched, idempotency, malformed-input defaults.
- `unified.adapter.spec.ts` — end-to-end raw INFO string → typed objects through `getInfoParsed`.

## Verification

- 170 unit suites / 2394 tests pass (`SKIP_DOCKER_SETUP=true jest`, e2e excluded; the one failure is a pre-existing environmental issue in `proprietary/licenses` caused by a persisted local license key — passes with a clean data dir).
- `pnpm build:api` and `pnpm --filter web build` pass.
- Live check against a running Valkey 9.1.0: `GET /mcp/instance/:id/health` now returns the real key count (previously always 0), `/metrics/info` serializes keyspace as objects, and `/prometheus/metrics` `betterdb_db_keys` matches.

---

> [!NOTE]
> **Medium Risk**
> Changes serialized API shapes for `/metrics/info` and health `keyspaceSize` semantics, and touches shared INFO parsing used by health, Prometheus, and anomaly detection—well tested but behavior-visible to API consumers.
> 
> **Overview**
> Fixes **#360** by making `MetricsParser.parseInfoToTyped` actually parse Redis INFO sections that were typed as objects but still held raw `k=v` strings at runtime.
> 
> **Parsing:** `keyspace` (`db*`), `commandstats` (`cmdstat_*`), and `errorstats` (`errorstat_*`) are converted via `InfoParser.parseKvLine` into numeric objects; scalar INFO sections stay strings. Malformed lines remain raw strings; parsing is idempotent. Extra keyspace fields (e.g. `subexpiry`) are preserved when numeric.
> 
> **Downstream:** `getHealthSummary` sums typed keyspace objects and returns `keyspaceSize: null` when the keyspace section is missing (vs `0` for empty). Prometheus keyspace gauges rely on typed objects only (string fallback removed). Types and web `InfoResponse` allow `string | object` for those sections.
> 
> **Hardening:** `__proto__` keys are skipped in `InfoParser` and `parseKvSection`; `toNumber` is exported from `commandstats-parser` for shared coercion.
> 
> **Tests:** New coverage for `parseInfoToTyped`, `getHealthSummary`, adapter `getInfoParsed`, and anomaly fixtures aligned with the real parser output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12ee4888746bcdf80742bfe2a68cdc4e99e297eb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/api/src/common/types/metrics.types.ts
+++ b/apps/api/src/common/types/metrics.types.ts
@@ -240,12 +240,19 @@ export interface ModulesInfo {
   [key: string]: unknown;
 }
 
+export interface KeyspaceDbInfo {
+  keys: number;
+  expires: number;
+  avg_ttl: number;
+  // Additional numeric fields some server versions emit (e.g. subexpiry on
+  // Redis 7.4+) are preserved by the parser.
+  [field: string]: number;
+}
+
 export interface KeyspaceInfo {
-  [dbKey: string]: {
-    keys: number;
-    expires: number;
-    avg_ttl: number;
-  };
+  // A string value means the line was not a parseable db entry (non-db* key
+  // or malformed) and was passed through raw.
+  [dbKey: string]: KeyspaceDbInfo | string;
 }
 
 export interface ClusterInfo {
@@ -254,19 +261,23 @@ export interface ClusterInfo {
 }
 
 export interface CommandStatsInfo {
-  [commandKey: string]: {
-    calls: number;
-    usec: number;
-    usec_per_call: number;
-    rejected_calls?: number;
-    failed_calls?: number;
-  };
+  // A string value means the line was not a parseable cmdstat entry and was
+  // passed through raw.
+  [commandKey: string]:
+    | {
+        calls: number;
+        usec: number;
+        usec_per_call: number;
+        rejected_calls?: number;
+        failed_calls?: number;
+      }
+    | string;
 }
 
 export interface ErrorStatsInfo {
-  [errorKey: string]: {
-    count: number;
-  };
+  // A string value means the line was not a parseable errorstat entry and was
+  // passed through raw.
+  [errorKey: string]: { count: number } | string;
 }
 
 export interface LatencyStatsInfo {
--- a/apps/api/src/database/adapters/unified.adapter.spec.ts
+++ b/apps/api/src/database/adapters/unified.adapter.spec.ts
@@ -54,3 +54,20 @@ describe('UnifiedDatabaseAdapter.getInfo — Redis 6 / KeyDB compatibility', () 
     expect(info).toHaveBeenCalledWith();
   });
 });
+
+describe('UnifiedDatabaseAdapter.getInfoParsed', () => {
+  it('parses raw INFO keyspace lines into typed objects (issue #360)', async () => {
+    const { adapter } = makeAdapter(
+      () =>
+        `# Stats\r\nkeyspace_hits:42\r\n\r\n# Keyspace\r\ndb0:keys=568,expires=310,avg_ttl=7510966104\r\n`,
+    );
+
+    const result = await adapter.getInfoParsed();
+
+    expect(result.keyspace).toEqual({
+      db0: { keys: 568, expires: 310, avg_ttl: 7510966104 },
+    });
+    // Scalar sections stay strings.
+    expect(result.stats?.keyspace_hits).toBe('42');
+  });
+});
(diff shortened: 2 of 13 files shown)


KIvanow:
@bugbot review

KIvanow:
@bugbot review

PR #372, opened 2026-08-10
fix(anomaly): skip underflowed memory samples in memory-overhead detector

### LINKED ISSUE #369: Guard memory-overhead detector against used_memory_dataset underflow
## Summary

Harden the existing `memory-overhead` detector against a known Valkey reporting artifact where `used_memory_dataset` underflows toward 2^64 and `used_memory_dataset_perc` goes absurd (well above 100 percent). This is not a new alert source; it is a **guard to suppress false positives**. Upstream valkey-io/valkey#1373 documents the artifact and maintainers describe it as transient and self-resolving on the next sample, so we should recognize and ignore it rather than alerting on it.

## Upstream context

- valkey-io/valkey#1373 "[BUG] Incorrectly reported used_memory_dataset and used_memory_dataset_perc".
- Cause: memory accounting is sampled non-atomically across threads (the lazyfree thread updates overhead concurrently), so momentarily `used_memory_overhead > used_memory`. Dataset memory is computed as `used_memory - used_memory_overhead`, so it underflows to a value near UINT64_MAX and the percentage explodes.
- Most visible on replicas. Maintainers consider it transient by design; the next INFO call reads a consistent snapshot.

## Problem for us

Our `memory-overhead` detector reasons about the split between dataset and non-dataset memory. If it consumes a sample taken mid-underflow, it can (a) emit a bogus "huge non-dataset overhead" anomaly, or (b) compute nonsense bytes-per-key. Both are false positives that erode trust.

## What we can observe

- INFO Memory: `used_memory`, `used_memory_overhead`, `used_memory_dataset`, `used_memory_dataset_perc`.
- MEMORY STATS: `dataset.bytes`, `keys.bytes-per-key`.

## Guard logic

Before the `memory-overhead` detector uses a sample, treat it as invalid and skip it when any of:

- `used_memory_overhead > used_memory` (the direct cause), or
- `used_memory_dataset` is implausibly large (near UINT64_MAX, e.g. greater than `used_memory`), or
- `used_memory_dataset_perc > 100`.

On a skipped sample: do not update baselines from it, do not emit an event, and (optionally) count consecutive skips so a genuinely persistent inconsistency across many polls can be surfaced as a low-severity "memory accounting looks inconsistent" note rather than a false overhead spike.

## Acceptance criteria

- `memory-overhead` detector ignores underflowed samples per the rules above.
- Unit tests: a crafted underflow sample (skipped, no event, baseline untouched), a normal sample (processed as today), and a run of consecutive underflows (optional low-severity note).
- No change to behavior on healthy instances.

Fit score from the Page 6 triage: 4/10 (MAYBE, guard rather than feature).

jamby77:
## Summary

Hardens the `memory-overhead` detector against the valkey-io/valkey#1373 reporting artifact: INFO memory accounting is sampled non-atomically across threads, so a sample can momentarily report `used_memory_overhead > used_memory`, wrapping `used_memory_dataset` toward 2^64 and pushing `used_memory_dataset_perc` far past 100%. Feeding such a sample into the detector produced bogus "huge non-dataset overhead" anomalies, most visibly on replicas.

## Changes

- Skip a sample when any of: `used_memory_overhead > used_memory`, `used_memory_dataset > used_memory`, or `used_memory_dataset_perc > 100`. A skipped sample emits nothing and — critically — does not touch hysteresis state, so a mid-glitch "low" reading cannot re-arm an already-acknowledged alert.
- Track consecutive skipped samples; a run of 10+ polls surfaces once as a low-severity INFO note ("memory accounting looks inconsistent", pointing at valkey#1373) rather than a false overhead spike. Upstream describes the artifact as gone by the next INFO read, so a run that long is genuinely worth a look.
- Service maps the new `info` finding level to `AnomalySeverity.INFO` (already rendered by the web dashboard) and logs it as an advisory, not an anomaly.
- Test fixture now models physically consistent INFO samples (`used_memory = dataset + overhead`); the old defaults were accidentally inconsistent and would themselves trip the guard.

## Test plan

- 7 new unit tests: each skip condition in isolation, normal-sample passthrough, hysteresis protection across a glitch, the persistent-run INFO note (emit-retry + acknowledge contract), and streak reset on a healthy sample.
- Full anomaly-detection suite: 505/505 green; `tsc --noEmit` clean.

Closes #369

---

> [!NOTE]
> **Low Risk**
> Detector-only change with conservative skip rules and INFO-level escalation for persistent inconsistency; no auth, persistence, or API surface changes.
> 
> **Overview**
> Hardens the **memory-overhead** detector against the valkey#1373 INFO artifact: non-atomic memory accounting can briefly report `used_memory_overhead > used_memory`, wrapping `used_memory_dataset` and inflating `used_memory_dataset_perc`, which previously produced false “huge overhead” anomalies (often on replicas).
> 
> **Skipped samples** when overhead exceeds total memory, dataset exceeds total memory, or dataset percent is above 100. Those polls emit nothing and **do not touch hysteresis**, so a glitch cannot re-arm an already-acknowledged warning.
> 
> After **10 consecutive** invalid polls, a single **INFO** advisory notes persistent inconsistency (with valkey#1373 context). `anomaly.service` maps that level to `AnomalySeverity.INFO` and logs it as an advisory rather than a warn-level anomaly.
> 
> Test fixtures now default to physically consistent INFO (`used_memory = dataset + overhead`); seven new unit tests cover skip conditions, hysteresis across glitches, the INFO note contract, and streak reset on a healthy sample.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbd434ce89e6682687f7fb90c1459d7a0b33620a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proprietary/anomaly-detection/__tests__/memory-overhead-detector.spec.ts
+++ b/proprietary/anomaly-detection/__tests__/memory-overhead-detector.spec.ts
@@ -1,4 +1,5 @@
 import {
+  INVALID_SAMPLE_NOTE_STREAK,
   MemoryOverheadInput,
   OVERHEAD_CRIT_FRACTION,
   OVERHEAD_WARN_FRACTION,
@@ -36,11 +37,15 @@ describe('evaluateMemoryOverhead', () => {
    */
   function input(over: Partial<MemoryOverheadInput> = {}): MemoryOverheadInput {
     const overheadBytes = over.usedMemoryOverhead ?? startup + 100 * MB;
+    const datasetBytes = over.usedMemoryDataset ?? 250 * MB;
     return {
-      usedMemory: 300 * MB,
+      // A consistent INFO sample reports used_memory = dataset + overhead;
+      // tests craft an INconsistent one by overriding usedMemory directly.
+      usedMemory: datasetBytes + overheadBytes,
       usedMemoryOverhead: overheadBytes,
       usedMemoryStartup: startup,
-      usedMemoryDataset: 250 * MB,
+      usedMemoryDataset: datasetBytes,
+      usedMemoryDatasetPerc: null,
       maxmemory: MAXMEMORY,
       maxmemoryPolicy: 'allkeys-lru',
       components: components({ clientsNormal: overheadBytes - startup }),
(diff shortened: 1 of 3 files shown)


jamby77:
@bugbot review

KIvanow (approved):
Overall - LGTM. A few nits for future PRs:

- Diff noise: roughly a third of the anomaly.service.ts churn is unrelated Prettier reformatting that could be split out to keep the fix focused.
- Event shape: the INFO advisory reuses the overhead event with anomalyType: SPIKE and value/threshold: 0, which reads a little oddly - a dedicated advisory type would be more honest.
- Guard placement: the underflow guard runs before the pure detector's maxmemory <= 0 early-return, which is moot today since the service gates on maxmemory first, but it's a latent coupling worth a comment.
- Test gap: the new info → AnomalySeverity.INFO mapping and the logger.log advisory branch in the service aren't covered - the detector is, but that wiring isn't.
- Scope note: the canonical valkey#1373 case is a maxmemory:0 replica, which the service already short-circuits before the detector, so the real-world blast radius here is the narrower maxmemory > 0 instances that hit the same sampling race -worth setting expectations accordingly.

None of these block merge or need to be explicitly fixed (it's just a nice to have)

PR #377, opened 2026-08-11
fix(docker): harden image (0 crit/high) and halve size

KIvanow:
## Summary
Harden and shrink the Docker image. Rebased onto latest `master`, so this now sits **on top of** the Dependabot dependency fixes (#364/#366) and touches **only the two Dockerfiles** — no dependency/lockfile changes here.

Result (grype, no-ai image): **0 Critical, 0 High** (7 Medium/Low remain — OS packages with no fix + suite-locked mediums). Image **1.34 GB → 739 MB**.

## Changes (Dockerfile + Dockerfile.prod)
- **RedisShake built from source** on Go 1.26 (upstream prebuilt ships EOL Go 1.21.13 stdlib); embedded `golang.org/x/text` bumped 0.14.0 → 0.39.0 — clears the Go CVE wall
- **Drop unused bundled npm CLI and GNU tar** from the runtime image
- **Healthcheck via `node`** instead of the wget package (removes wget CVEs); uses the correct `/api/health` path; also removes the prod-stage `apk del wget` that had been breaking the check
- **`COPY --chown`** instead of `chown -R /app` — removes a duplicated node_modules layer (~600 MB), then a non-recursive chown restores `/app` writability for the runtime user (RedisShake data dir, sqlite/license)
- **Base image** `node:25-alpine` → `node:26-alpine` (25 is non-LTS with no patch line)

Verified locally: builds on node 26 (native modules recompile), container boots, `/api/health` returns real JSON 200, docker healthcheck healthy, static UI serves, redis-shake runs, `/app` writable.

## Checklist
- [ ] Unit / integration tests added _(n/a — Docker only)_
- [ ] Docs added / updated

---

> [!NOTE]
> **Medium Risk**
> Touches production image build, runtime ownership, and the RedisShake migration binary (now a from-source build with a patched dependency). Failures here can break container boots, healthchecks, or migrations.
> 
> **Overview**
> Hardens and shrinks both `Dockerfile` and `Dockerfile.prod`: base image moves to **`node:26-alpine`**, unused runtime packages are dropped, and image size roughly halves by replacing recursive `chown -R /app` with `COPY --chown` plus a non-recursive `/app`/`/app/data` chown.
> 
> **RedisShake** is no longer downloaded as a prebuilt release. It is built from pinned `v4.6.1` source on **Go 1.26**, with `golang.org/x/text` bumped to clear EOL-stdlib CVEs, then copied into the runtime image.
> 
> Also removes the bundled npm CLI (CVE surface), avoids installing full wget/GNU tar, and switches the healthcheck to busybox `wget` against `127.0.0.1:${PORT:-3001}/api/health`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1680e4355243464e6c894831d364c12f5fedb7e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,14 +66,38 @@ ENV VITE_REGISTRATION_URL=$VITE_REGISTRATION_URL
 RUN pnpm --filter "api..." --filter "web..." build
 
 # ============================================
+# RedisShake Build Stage
+# ============================================
+# Build the migration binary from source with a current Go toolchain so the
+# embedded Go standard library is patched. The official prebuilt releases are
+# compiled with Go 1.21.13 (EOL), whose stdlib carries dozens of CVEs (incl.
+# criticals) that `apk upgrade` cannot fix - they are baked into the static
+# binary, not provided by an Alpine package.
+FROM golang:1.26-alpine AS redisshake-builder
+ARG TARGETARCH
+ARG REDISSHAKE_VERSION=4.6.1
+# v4.6.1 is a lightweight tag -> pin its exact release commit for tamper-evidence.
+ARG REDISSHAKE_COMMIT=a2a7e4e46d15708b6ab203e1c10a108aa405a638
+RUN apk add --no-cache git
+WORKDIR /build
+RUN git clone --depth 1 --branch "v${REDISSHAKE_VERSION}" https://github.com/tair-opensource/RedisShake.git . && \
+    test "$(git rev-parse HEAD)" = "${REDISSHAKE_COMMIT}" && \
+    CGO_ENABLED=0 GOOS=linux GOARCH="${TARGETARCH}" \
+        go build -trimpath -ldflags "-s -w" -o /out/redis-shake ./cmd/redis-shake
+
+# ============================================
 # Production Stage
 # ============================================
 FROM node:25-alpine AS production
 
-# Install wget for healthcheck and tar (>=7.5.4) for security fix
-# Upgrade all packages to get latest security patches (including Go stdlib in binaries)
-RUN apk add --no-cache wget tar>=7.5.4 && \
-    apk upgrade --no-cache
+# Apply the latest Alpine security patches and drop the npm CLI bundled in the base
+# image. This image runs via `node` directly (deps are installed with pnpm at build
+# time), so npm is unused at runtime and only contributes CVEs from its own vendored
+# dependencies. No extra packages are installed: the healthcheck uses `node` (below)
+# instead of wget, and busybox already provides tar - avoiding the full wget and GNU
+# tar packages, both of which currently ship unfixable CVEs.
+RUN apk upgrade --no-cache && \
+    rm -rf /usr/local/lib/node_modules/npm /usr/local/bin/npm /usr/local/bin/npx
 
 WORKDIR /app
 
@@ -126,17 +150,11 @@ ENV NODE_ENV=production
 ENV PORT=3001
 ENV STORAGE_TYPE=memory
 
-# Install RedisShake binary for migration execution (with checksum verification)
-ARG TARGETARCH
-ARG REDISSHAKE_VERSION=4.6.0
-RUN REDISSHAKE_SHA256_AMD64="6ccab1ff2ba3c200950f8ada811f0c6fe6e2f5e6bd3b8e92b4d9444dc0aff4df" && \
-    REDISSHAKE_SHA256_ARM64="653298efa83ef3d495ae2ec21b40c773f36eb15e507f8b3f2931660509d09690" && \
-    if [ "${TARGETARCH}" = "amd64" ]; then EXPECTED_SHA256="${REDISSHAKE_SHA256_AMD64}"; else EXPECTED_SHA256="${REDISSHAKE_SHA256_ARM64}"; fi && \
-    wget -qO /tmp/redis-shake.tar.gz "https://github.com/tair-opensource/RedisShake/releases/download/v${REDISSHAKE_VERSION}/redis-shake-v${REDISSHAKE_VERSION}-linux-${TARGETARCH}.tar.gz" && \
-    echo "${EXPECTED_SHA256}  /tmp/redis-shake.tar.gz" | sha256sum -c - && \
-    tar -xzf /tmp/redis-shake.tar.gz --strip-components=0 -C /usr/local/bin ./redis-shake && \
-    chmod +x /usr/local/bin/redis-shake && \
-    rm /tmp/redis-shake.tar.gz
+# Install RedisShake binary for migration execution. Built from source in the
+# redisshake-builder stage with a current Go toolchain so the embedded Go stdlib
+# is patched (the upstream prebuilt release ships an EOL Go 1.21.13 runtime).
+COPY --from=redisshake-builder /out/redis-shake /usr/local/bin/redis-shake
+RUN chmod +x /usr/local/bin/redis-shake
 
 # Create non-root user for security (Docker Scout compliance)
 RUN addgroup --system --gid 1001 nodejs && \
@@ -151,9 +169,10 @@ USER betterdb
 # Note: EXPOSE is documentation only - actual port binding happens via -p flag
 EXPOSE 3001
 
-# Health check - uses PORT environment variable
+# Health check - uses PORT environment variable. Uses node (always present) rather
+# than wget so the image needs no extra wget package.
 HEALTHCHECK --interval=30s --timeout=3s --start-period=5s --retries=3 \
-  CMD wget --no-verbose --tries=1 --spider http://localhost:${PORT}/api/health || exit 1
+  CMD node -e "require('http').get('http://127.0.0.1:'+(process.env.PORT||3001)+'/api/health',r=>process.exit(r.statusCode===200?0:1)).on('error',()=>process.exit(1))"
 
 # Start the server
 CMD ["node", "apps/api/dist/apps/api/src/main.js"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -165,6 +165,13 @@ ENV STORAGE_TYPE=memory
 COPY --from=redisshake-builder /out/redis-shake /usr/local/bin/redis-shake
 RUN chmod +x /usr/local/bin/redis-shake
 
+# Make /app writable by the runtime user so features that create new paths under
+# it at runtime work - notably RedisShake's default relative `data` dir (it is
+# spawned with cwd=/app) and sqlite/license files. This chowns only the /app
+# directory node plus a pre-created data dir (non-recursive), so it does NOT
+# duplicate node_modules the way `chown -R /app` did.
+RUN mkdir -p /app/data && chown betterdb:nodejs /app /app/data
+
 # Drop to the non-root user (created above) for the runtime process.
 USER betterdb
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,7 +73,20 @@ RUN pnpm --filter "api..." --filter "web..." build
 # compiled with Go 1.21.13 (EOL), whose stdlib carries dozens of CVEs (incl.
 # criticals) that `apk upgrade` cannot fix - they are baked into the static
 # binary, not provided by an Alpine package.
-FROM golang:1.26-alpine AS redisshake-builder
+#
+# --platform=$BUILDPLATFORM pins this stage to the native build host so `go build`
+# cross-compiles (via GOARCH below) instead of running the toolchain under QEMU on
+# the arm64 leg - QEMU is 5-10x slower and a known source of spurious SIGSEGV.
+#
+# The `go get golang.org/x/text@v0.39.0` step raises x/text to clear the Go CVE wall
+# (v4.6.1's go.mod pins the vulnerable 0.14.0). Two caveats it introduces:
+#   - It mutates go.mod/go.sum AFTER the commit tamper-check, so the shipped binary
+#     is a v4.6.1 + x/text-0.39.0 combination upstream never released; an SBOM audit
+#     against RedisShake v4.6.1 will show that single dependency delta.
+#   - `go get pkg@exact-version` also DOWNGRADES. Once a future REDISSHAKE_VERSION's
+#     go.mod already requires x/text >= 0.39.0, remove that line - otherwise it
+#     silently pins x/text back down on the version bump.
+FROM --platform=$BUILDPLATFORM golang:1.26-alpine AS redisshake-builder
 ARG TARGETARCH
 ARG REDISSHAKE_VERSION=4.6.1
 # v4.6.1 is a lightweight tag -> pin its exact release commit for tamper-evidence.
@@ -84,7 +97,7 @@ RUN git clone --depth 1 --branch "v${REDISSHAKE_VERSION}" https://github.com/tai
     test "$(git rev-parse HEAD)" = "${REDISSHAKE_COMMIT}" && \
     go get golang.org/x/text@v0.39.0 && \
     CGO_ENABLED=0 GOOS=linux GOARCH="${TARGETARCH}" \
-        go build -trimpath -ldflags "-s -w" -o /out/redis-shake ./cmd/redis-shake
+        go build -trimpath -ldflags "-s -w -X main.Version=v${REDISSHAKE_VERSION} -X main.GitCommit=${REDISSHAKE_COMMIT}" -o /out/redis-shake ./cmd/redis-shake
 
 # ============================================
 # Production Stage
@@ -94,9 +107,9 @@ FROM node:26-alpine AS production
 # Apply the latest Alpine security patches and drop the npm CLI bundled in the base
 # image. This image runs via `node` directly (deps are installed with pnpm at build
 # time), so npm is unused at runtime and only contributes CVEs from its own vendored
-# dependencies. No extra packages are installed: the healthcheck uses `node` (below)
-# instead of wget, and busybox already provides tar - avoiding the full wget and GNU
-# tar packages, both of which currently ship unfixable CVEs.
+# dependencies. No extra packages are installed: the healthcheck uses the busybox
+# wget already in the base image, and busybox already provides tar - avoiding the
+# full wget and GNU tar packages, both of which currently ship unfixable CVEs.
 RUN apk upgrade --no-cache && \
     rm -rf /usr/local/lib/node_modules/npm /usr/local/bin/npm /usr/local/bin/npx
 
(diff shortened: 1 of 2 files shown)


jamby77 (commented):
The core mechanisms of the PR held up well under adversarial checks: the pinned commit matches the real v4.6.1 release commit, all runtime write paths resolve to the pre-created `/app/data`, and the wget-removal/npm-removal rationale is sound. Inline comments carry one-click suggestions where a concrete fix exists (posted on both Dockerfiles).

Dockerfile
```diff
@@ -65,53 +65,86 @@ ENV VITE_REGISTRATION_URL=$VITE_REGISTRATION_URL
...
 
+# ============================================
+# RedisShake Build Stage
+# ============================================
+# Build the migration binary from source with a current Go toolchain so the
+# embedded Go standard library is patched. The official prebuilt releases are
+# compiled with Go 1.21.13 (EOL), whose stdlib carries dozens of CVEs (incl.
+# criticals) that `apk upgrade` cannot fix - they are baked into the static
+# binary, not provided by an Alpine package.
+FROM golang:1.26-alpine AS redisshake-builder
```
jamby77: **[Confirmed] arm64 leg of this stage compiles under QEMU emulation.** `docker-publish.yml` sets up QEMU and builds `linux/amd64,linux/arm64` for both Dockerfiles; without `--platform=$BUILDPLATFORM` the arm64 pass pulls the arm64 golang image and runs the whole clone + `go build` under qemu-user, even though the stage is already a pure cross-compile (`CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH}`). On every cache miss that's typically 5-10x slower (minutes per release, twice - the two builds use disjoint cache scopes), and the Go toolchain under qemu-user is a known source of spurious SIGSEGV, i.e. intermittently red tag builds. The existing GOOS/GOARCH env already produces the correct target binary when the stage runs natively.

```suggestion
FROM --platform=$BUILDPLATFORM golang:1.26-alpine AS redisshake-builder
```

(Same applies to `Dockerfile.prod` line 151 - twin suggestion posted there.)

Dockerfile.prod
```diff
@@ -140,49 +140,82 @@ RUN rm -rf \
...
 
+# ============================================
+# RedisShake Build Stage
+# ============================================
+# Build the migration binary from source with a current Go toolchain so the
+# embedded Go standard library is patched. The official prebuilt releases are
+# compiled with Go 1.21.13 (EOL), whose stdlib carries dozens of CVEs (incl.
+# criticals) that `apk upgrade` cannot fix - they are baked into the static
+# binary, not provided by an Alpine package.
+FROM golang:1.26-alpine AS redisshake-builder
```
jamby77: **[Confirmed]** Twin of the `Dockerfile` line 76 comment: add `--platform=$BUILDPLATFORM` so the arm64 leg cross-compiles natively instead of running the Go toolchain under QEMU.

```suggestion
FROM --platform=$BUILDPLATFORM golang:1.26-alpine AS redisshake-builder
```

Dockerfile
```diff
@@ -126,34 +159,30 @@ ENV NODE_ENV=production
...
 # Expose port (can be overridden with -e PORT=<port> at runtime)
 # Note: EXPOSE is documentation only - actual port binding happens via -p flag
 EXPOSE 3001
 
-# Health check - uses PORT environment variable
+# Health check - uses PORT environment variable. Uses node (always present) rather
+# than wget so the image needs no extra wget package.
 HEALTHCHECK --interval=30s --timeout=3s --start-period=5s --retries=3 \
-  CMD wget --no-verbose --tries=1 --spider http://localhost:${PORT}/api/health || exit 1
+  CMD node -e "require('http').get('http://127.0.0.1:'+(process.env.PORT||3001)+'/api/health',r=>process.exit(r.statusCode===200?0:1)).on('error',()=>process.exit(1))"
```
jamby77: **[Confirmed] Node-based healthcheck can exceed the kept 3s timeout under CPU pressure - and busybox `wget` makes the node trick unnecessary.** Two things verified empirically:
1. busybox wget ships in `node:26-alpine` (`/usr/bin/wget -> /bin/busybox`) with **no package install** - the old `apk add wget` was always unnecessary, and busybox `--spider` has the right semantics (200 -> exit 0; 404 / connection refused -> exit 1).
2. The comment's rationale aside, `sh -> node -e` boots a full interpreter every 30s. Under a saturated CPU-capped cgroup (exactly what a RedisShake full sync causes - `migration-execution.service.ts:152` spawns it in this same container) the PR's command measured 2.3s / 5.2s / 6.0s at `--cpus=0.2`, vs 0.2-0.4s for busybox wget. Three misses (~90s of a full sync) flip the container unhealthy and restart-on-unhealthy orchestrators kill it mid-migration.

```suggestion
HEALTHCHECK --interval=30s --timeout=3s --start-period=5s --retries=3 \
  CMD wget -T 2 -q --spider "http://127.0.0.1:${PORT:-3001}/api/health" || exit 1
```

The `${PORT:-3001}` default also covers the fallback the node one-liner added. If you take this, the comment above ("Uses node ... no extra wget package") should be updated - no package is needed either way.

(Same applies to `Dockerfile.prod` lines 262-263 - twin suggestion posted there.)

Dockerfile.prod
```diff
@@ -204,36 +237,30 @@ ENV DB_USERNAME=default
...
 # Expose port (can be overridden with -e PORT=<port> at runtime)
 # Note: EXPOSE is documentation only - actual port binding happens via -p flag
 EXPOSE 3001
 
-# Health check - uses PORT environment variable
+# Health check - uses PORT environment variable. Uses node (always present) rather
+# than wget so the image needs no extra wget package.
 HEALTHCHECK --interval=30s --timeout=3s --start-period=5s --retries=3 \
-  CMD wget --no-verbose --tries=1 --spider http://localhost:${PORT}/api/health || exit 1
+  CMD node -e "require('http').get('http://127.0.0.1:'+(process.env.PORT||3001)+'/api/health',r=>process.exit(r.statusCode===200?0:1)).on('error',()=>process.exit(1))"
```
jamby77: **[Confirmed]** Twin of the `Dockerfile` lines 184-185 comment: busybox wget is already in the base image and avoids node's cold-start blowing the 3s timeout under CPU pressure (measured 2.3-6.0s at `--cpus=0.2` vs 0.2-0.4s for wget).

```suggestion
HEALTHCHECK --interval=30s --timeout=3s --start-period=5s --retries=3 \
  CMD wget -T 2 -q --spider "http://127.0.0.1:${PORT:-3001}/api/health" || exit 1
```

Dockerfile
```diff
@@ -65,53 +65,86 @@ ENV VITE_REGISTRATION_URL=$VITE_REGISTRATION_URL
...
+FROM golang:1.26-alpine AS redisshake-builder
+ARG TARGETARCH
+ARG REDISSHAKE_VERSION=4.6.1
+# v4.6.1 is a lightweight tag -> pin its exact release commit for tamper-evidence.
+ARG REDISSHAKE_COMMIT=a2a7e4e46d15708b6ab203e1c10a108aa405a638
+RUN apk add --no-cache git
+WORKDIR /build
+RUN git clone --depth 1 --branch "v${REDISSHAKE_VERSION}" https://github.com/tair-opensource/RedisShake.git . && \
+    test "$(git rev-parse HEAD)" = "${REDISSHAKE_COMMIT}" && \
+    go get golang.org/x/text@v0.39.0 && \
```
jamby77, line 98: **[Confirmed] This exact-version pin will silently downgrade x/text on a future RedisShake bump.** `go get pkg@version` explicitly *downgrades* when the module's requirement is already higher (`go help get`: "To upgrade or downgrade a package to a specific version"). When `REDISSHAKE_VERSION` is next bumped to a release whose go.mod already requires x/text >= v0.40.0, this line silently drops it back to v0.39.0 - reintroducing whatever CVEs the newer version fixed, with no build failure to alert anyone (or a confusing conflict error if another module strictly needs newer). The pin is also duplicated in `Dockerfile.prod` line 160 and nothing documents which advisory it serves, so nobody will know when it can be deleted. At minimum, document the intent and the downgrade hazard next to the line:

```suggestion
    # x/text >= 0.39.0 clears the Go CVE wall (v4.6.1's go.mod pins 0.14.0).
    # NB: exact-version `go get` also DOWNGRADES - delete this line once upstream requires >= v0.39.0.
    go get golang.org/x/text@v0.39.0 && \
```

(Comment lines inside RUN continuations are stripped by the Dockerfile parser, so this is build-safe. Twin suggestion posted on `Dockerfile.prod`.)

Dockerfile.prod
```diff
@@ -140,49 +140,82 @@ RUN rm -rf \
...
+FROM golang:1.26-alpine AS redisshake-builder
+ARG TARGETARCH
+ARG REDISSHAKE_VERSION=4.6.1
+# v4.6.1 is a lightweight tag -> pin its exact release commit for tamper-evidence.
+ARG REDISSHAKE_COMMIT=a2a7e4e46d15708b6ab203e1c10a108aa405a638
+RUN apk add --no-cache git
+WORKDIR /build
+RUN git clone --depth 1 --branch "v${REDISSHAKE_VERSION}" https://github.com/tair-opensource/RedisShake.git . && \
+    test "$(git rev-parse HEAD)" = "${REDISSHAKE_COMMIT}" && \
+    go get golang.org/x/text@v0.39.0 && \
```
jamby77, line 173: **[Confirmed]** Twin of the `Dockerfile` line 85 comment: exact-version `go get` downgrades too - document the pin so it gets deleted (not silently applied as a downgrade) when the RedisShake version moves.

```suggestion
    # x/text >= 0.39.0 clears the Go CVE wall (v4.6.1's go.mod pins 0.14.0).
    # NB: exact-version `go get` also DOWNGRADES - delete this line once upstream requires >= v0.39.0.
    go get golang.org/x/text@v0.39.0 && \
```

Dockerfile
```diff
@@ -65,53 +65,86 @@ ENV VITE_REGISTRATION_URL=$VITE_REGISTRATION_URL
...
+    CGO_ENABLED=0 GOOS=linux GOARCH="${TARGETARCH}" \
+        go build -trimpath -ldflags "-s -w" -o /out/redis-shake ./cmd/redis-shake
```
jamby77: **[Confirmed] The built binary self-reports `redis-shake version unknown ... (Git SHA: unknown)`.** RedisShake's `cmd/redis-shake/main.go` defaults `Version`/`GitCommit` to "unknown" and upstream's release build injects them via `-X main.Version -X main.GitCommit` ldflags (see upstream `build.sh`); this build passes only `-s -w`. Every migration run now logs an "unknown" version line, `redis-shake -v` can't confirm 4.6.1, and the runtime provenance the commit pin exists to provide is invisible when correlating customer migration logs with a known build. The ARGs are already in scope:

```suggestion
        go build -trimpath -ldflags "-s -w -X main.Version=v${REDISSHAKE_VERSION} -X main.GitCommit=${REDISSHAKE_COMMIT}" -o /out/redis-shake ./cmd/redis-shake
```

(Same applies to `Dockerfile.prod` line 162 - twin suggestion posted there.)

Dockerfile.prod
```diff
@@ -140,49 +140,82 @@ RUN rm -rf \
...
+    CGO_ENABLED=0 GOOS=linux GOARCH="${TARGETARCH}" \
+        go build -trimpath -ldflags "-s -w" -o /out/redis-shake ./cmd/redis-shake
```
jamby77: **[Confirmed]** Twin of the `Dockerfile` line 87 comment: inject the version ldflags so the shipped binary doesn't self-report "unknown".

```suggestion
        go build -trimpath -ldflags "-s -w -X main.Version=v${REDISSHAKE_VERSION} -X main.GitCommit=${REDISSHAKE_COMMIT}" -o /out/redis-shake ./cmd/redis-shake
```

Dockerfile
```diff
@@ -126,34 +159,30 @@ ENV NODE_ENV=production
...
-    rm /tmp/redis-shake.tar.gz
-
-# Create non-root user for security (Docker Scout compliance)
-RUN addgroup --system --gid 1001 nodejs && \
-    adduser --system --uid 1001 --ingroup nodejs betterdb
+# Install RedisShake binary for migration execution. Built from source in the
+# redisshake-builder stage with a current Go toolchain so the embedded Go stdlib
+# is patched (the upstream prebuilt release ships an EOL Go 1.21.13 runtime).
+COPY --from=redisshake-builder /out/redis-shake /usr/local/bin/redis-shake
+RUN chmod +x /usr/local/bin/redis-shake
```
jamby77: **[Confirmed] This `RUN chmod +x` is a no-op that ships the binary twice.** `go build` emits the binary 0755 and `COPY --from` preserves the mode - verified empirically, including the layer cost: overlay2 copy-up triggers on the setattr itself even when the mode value doesn't change (a no-op chmod on a 20MB test file produced a 21MB layer). So every published image carries the ~15-20MB redis-shake binary in two layers - the exact duplication cost the `COPY --chown` comments in this PR were written to avoid. One instruction does both:

```suggestion
COPY --chmod=755 --from=redisshake-builder /out/redis-shake /usr/local/bin/redis-shake
```

(Same applies to `Dockerfile.prod` lines 243-244 - twin suggestion posted there.)

Dockerfile.prod
```diff
@@ -204,36 +237,30 @@ ENV DB_USERNAME=default
...
-    apk del wget
-
-# Create non-root user for security (Docker Scout compliance)
-RUN addgroup --system --gid 1001 nodejs && \
-    adduser --system --uid 1001 --ingroup nodejs betterdb
+# Install RedisShake binary for migration execution. Built from source in the
+# redisshake-builder stage with a current Go toolchain so the embedded Go stdlib
+# is patched (the upstream prebuilt release ships an EOL Go 1.21.13 runtime).
+COPY --from=redisshake-builder /out/redis-shake /usr/local/bin/redis-shake
+RUN chmod +x /usr/local/bin/redis-shake
```
jamby77: **[Confirmed]** Twin of the `Dockerfile` lines 165-166 comment: the chmod is a no-op that duplicates the binary into a second layer; `COPY --chmod=755` does both in one instruction.

```suggestion
COPY --chmod=755 --from=redisshake-builder /out/redis-shake /usr/local/bin/redis-shake
```

Dockerfile
```diff
@@ -65,53 +65,86 @@ ENV VITE_REGISTRATION_URL=$VITE_REGISTRATION_URL
...
+# binary, not provided by an Alpine package.
+FROM golang:1.26-alpine AS redisshake-builder
+ARG TARGETARCH
+ARG REDISSHAKE_VERSION=4.6.1
+# v4.6.1 is a lightweight tag -> pin its exact release commit for tamper-evidence.
+ARG REDISSHAKE_COMMIT=a2a7e4e46d15708b6ab203e1c10a108aa405a638
+RUN apk add --no-cache git
+WORKDIR /build
+RUN git clone --depth 1 --branch "v${REDISSHAKE_VERSION}" https://github.com/tair-opensource/RedisShake.git . && \
+    test "$(git rev-parse HEAD)" = "${REDISSHAKE_COMMIT}" && \
```
jamby77, line 97: **[Observation - non-blocking] The integrity story is weaker than the sha256-of-artifact it replaces.** The commit check on this line verifies the cloned tree, but the very next step (`go get golang.org/x/text@v0.39.0`) mutates go.mod/go.sum *after* the tamper check, and the final binary hash is pinned nowhere - while the workflow's `provenance: true` / `sbom: true` attestations record an artifact whose dependency set no longer matches the pinned upstream commit. Practical residual risk is modest (the commit pin + Go's checksum DB make module tampering fail closed), but the shipped binary is a 4.6.1 + x/text-0.39.0 combination upstream never released or tested, and an SBOM audit against RedisShake v4.6.1 will show a mismatch. Worth a note in the stage comment, and ideally an upstream PR bumping x/text so the pin (and the mismatch) can go away.

Dockerfile
```diff
@@ -65,53 +65,86 @@ ENV VITE_REGISTRATION_URL=$VITE_REGISTRATION_URL
...
+# criticals) that `apk upgrade` cannot fix - they are baked into the static
+# binary, not provided by an Alpine package.
+FROM golang:1.26-alpine AS redisshake-builder
+ARG TARGETARCH
+ARG REDISSHAKE_VERSION=4.6.1
+# v4.6.1 is a lightweight tag -> pin its exact release commit for tamper-evidence.
+ARG REDISSHAKE_COMMIT=a2a7e4e46d15708b6ab203e1c10a108aa405a638
+RUN apk add --no-cache git
+WORKDIR /build
+RUN git clone --depth 1 --branch "v${REDISSHAKE_VERSION}" https://github.com/tair-opensource/RedisShake.git . && \
```
jamby77, line 96: **[Observation - non-blocking] This stage's failure modes are only ever exercised on release day.** `docker-publish.yml` triggers on `v*` tags/dispatch only, and no PR-time job builds either Dockerfile - so a moved/deleted tag (fails closed at the commit check, by design), a proxy.golang.org outage, or a `golang:1.26-alpine` patch regression is discovered for the first time mid-release. The blind spot predates this PR (the old wget lived in the same path), but the PR widens it from one artifact download to a full toolchain build. A cheap guard: a PR-time or scheduled job that builds just this stage (`docker build --target redisshake-builder .`).

Dockerfile.prod
jamby77: **[Cleanup] The hand-maintained duplication between the two Dockerfiles grows from 131 to 160 byte-identical lines in this PR** - this whole stage (with four coupled pins: `golang:1.26`, `REDISSHAKE_VERSION`, `REDISSHAKE_COMMIT`, the x/text version), the apk/npm-removal block, the user/chown/chmod blocks, and the 200-char healthcheck line are all character-identical with `Dockerfile`, with no CI build-arg as a single source of truth. Drift between exactly these two files already caused one real regression (commit 9708786b patched only `Dockerfile` to re-sync). The CI cache scopes are also disjoint (`with-ai`/`without-ai`), so this identical stage compiles 4x per release with zero sharing. `docker/build-push-action@v6` supports `target:`, so a single multi-target Dockerfile (shared builder + redisshake-builder stages, terminal `production` / `production-no-ai` targets) would need only a `target:` line per existing CI step - happy to leave that for a follow-up, but flagging it while the surface is growing.

Dockerfile
KIvanow: Fair point — this is fine for now. Depending on usage we can add a PR-time or scheduled `docker build --target redisshake-builder` guard later. Leaving as-is for this PR.

Dockerfile.prod
KIvanow: Great point — the duplication is real and this makes sense to refactor into a single multi-target Dockerfile (shared builder + redisshake-builder stages, terminal production / production-no-ai targets). Doing it in a separate PR to keep this one focused on the security/build fixes.

jamby77 (approved):

PR #378, opened 2026-08-12
feat(migration): fork-aware functions, clearer BUSYKEY error, scroll-to-bottom

KIvanow:
## Summary
Four migration correctness/UX fixes, isolated on their own branch (no Docker changes).

## Changes
- **Functions only migrate within the same engine.** For cross-fork migrations (Valkey ↔ Redis) the RedisShake config blocks the `FUNCTION` command (`[filter] block_command=["function"]`), so key data still migrates instead of aborting on `FUNCTION LOAD` (function libraries use engine-specific globals like Valkey's `server`).
- **Cross-fork warning.** The compatibility report now flags that functions won't be carried over between different engines.
- **Clearer BUSYKEY failure.** A migration that fails because the target already has data now reports an actionable message pointing to the **"Flush target before migration"** option, instead of `exited with code N`.
- **Scroll-to-bottom on validation.** Starting validation scrolls to the validation panel/controls at the bottom, not the top of the re-rendered report.

## Tests
Added unit coverage for the toml filter, the cross-fork warning, and the execution wiring. API + web typecheck clean; 73 migration unit tests pass.

## Checklist
- [ ] Docs added / updated

---

> [!NOTE]
> **Medium Risk**
> Changes migration execution (optional target FLUSHALL, RedisShake config filtering) and failure handling; behavior is heavily unit-tested but mistakes could affect live data during migrations.
> 
> **Overview**
> Improves **Valkey ↔ Redis migration** around server-side function libraries and RedisShake failure UX.
> 
> **Functions:** Valkey→Redis runs now add a RedisShake `[filter]` that blocks `FUNCTION-*` commands so key data still migrates. Analysis probes `FUNCTION LIST` (per cluster master when needed) and adds compatibility warnings—excluded on cross-fork, or “use Sync mode” when libraries exist on same-engine paths. Execution pushes **durable `notices`** (not rolling logs) when exclusion applies and the source may have functions.
> 
> **Failures & run lifecycle:** RedisShake logs are ANSI-stripped and line-buffered; completion waits on **`close`** so fatal lines (e.g. BUSYKEY) are classified into `failureCode` with actionable copy. The UI offers **flush target & retry** on BUSYKEY. Pre-spawn probe/target flush moved into `runRedisShake` with **cancel checks** so a stop during probe/flush does not still spawn or flush.
> 
> **Other:** TOML builders take a single options object; shared `parseNodeAddress` / fork-compat helpers; validation scroll uses `block: 'nearest'`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08ca1a43b59e0996a7b8f42ff6a9df3beeab7461. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Improved Redis and Valkey migration compatibility checks, including cluster-aware function-library detection and warnings.
  * Automatically excludes incompatible function commands during cross-engine migrations.
  * Added persistent migration notices and structured failure details, including BUSYKEY errors.
* **Bug Fixes**
  * Improved migration log handling to preserve complete messages across streamed output.
  * Enhanced failure messages with actionable final-error information and exit-code details.
  * Refined validation-panel scrolling to minimize unnecessary movement.


apps/api/src/migration/execution/toml-builder.ts
```diff
@@ -33,6 +33,21 @@ function validateHost(host: string): string {
...
+ * When source and target are different engines (e.g. Valkey -> Redis), server-side
+ * function libraries must not be replicated: they use engine-specific globals (such
+ * as Valkey's `server`) and fail to load on the other fork, aborting the migration.
+ * Blocking the FUNCTION command drops FUNCTION LOAD/RESTORE from the stream so the
+ * key data still migrates. Returns an empty string (no filter) when not excluding.
+ */
+function buildFilterSection(excludeFunctions: boolean): string {
+  if (!excludeFunctions) return '';
+  return `[filter]
+block_command = ["function"]
```
jamby77: **The filter is a no-op against the pinned RedisShake v4.6.0.** `internal/commands/keys.go` uppercases and container-expands the command name (`FUNCTION-LOAD` / `FUNCTION-RESTORE`), and `internal/filter/filter.go` does an exact `slices.Contains` against the config literal — so `"function"` never matches and function payloads are still forwarded to the target.

```suggestion
block_command = ["FUNCTION-LOAD", "FUNCTION-RESTORE"]
```

Worth adding coverage that pins the working literal; the current tests only assert the generated string, so they stay green either way.

apps/api/src/migration/migration-execution.service.ts
```diff
@@ -195,7 +208,7 @@ export class MigrationExecutionService {
         }
       } else if (statusAfterExit !== 'cancelled') {
         job.status = 'failed';
-        job.error = `RedisShake exited with code ${code}`;
+        job.error = this.explainRedisShakeFailure(code, job.logs);
```
jamby77: **Race: `job.logs` is read before stderr is drained.** The promise above resolves on `proc.on('exit')` (line 198), which Node emits before the stdio pipes flush — but RedisShake's fatal BUSYKEY line is typically its last stderr write, so `/BUSYKEY/i` often misses and the user gets the generic message this PR set out to replace.

Resolve on `'close'` instead (keep `'exit'` only to capture the code):

```ts
let exitCode = 1;
const code = await new Promise<number>((resolve, reject) => {
  proc.on('exit', (c) => { exitCode = c ?? 1; });
  proc.on('close', () => resolve(exitCode));
  proc.on('error', reject);
});
```

(Anchored here because line 198 is outside the diff.)

apps/api/src/migration/migration-execution.service.ts
```diff
@@ -51,6 +51,18 @@ export class MigrationExecutionService {
     const targetClusterSection = (targetInfo as Record<string, Record<string, string>>).cluster ?? {};
     const targetIsCluster = String(targetClusterSection['cluster_enabled'] ?? '0') === '1';
 
+    // Server-side functions use engine-specific globals (e.g. Valkey's `server`) and
+    // fail to load on a different fork. When source and target are different engines,
+    // exclude functions from the RedisShake stream so the key data still migrates.
+    const sourceDbType = sourceAdapter.getCapabilities().dbType;
+    const targetDbType = targetAdapter.getCapabilities().dbType;
+    const excludeFunctions = sourceDbType !== targetDbType;
```
jamby77: _Non-blocking._ **The condition is symmetric, so Redis→Valkey loses functions too.** Valkey is a Redis 7.2 fork and still exposes the `redis` global, so Redis-sourced libraries load fine there — once the filter literal above is fixed, this silently strips functions on the primary supported direction while the job reports success.

```suggestion
    const excludeFunctions = sourceDbType === 'valkey' && targetDbType === 'redis';
```

If you keep it symmetric, the warning text in `compatibility-checker.ts` needs correcting — "would fail to load on the target" is false for Redis→Valkey.

apps/api/src/migration/migration-execution.service.ts
```diff
@@ -221,6 +234,24 @@ export class MigrationExecutionService {
     }
   }
 
+  /**
+   * Turn a non-zero RedisShake exit into an actionable message. Scans the recent
+   * log tail for well-known failure signatures so the UI can tell the user what to
+   * do next, instead of a bare "exited with code N".
+   */
+  private explainRedisShakeFailure(code: number | null, logs: string[]): string {
+    const recent = logs.slice(-80).join('\n');
```
jamby77: _Non-blocking._ **The heuristic can both miss and misattribute.** An 80-line tail of the 500-line buffer is easily overrun by progress/teardown output or a Go panic dump, and conversely a stale non-fatal BUSYKEY within the window will tell a user whose job was OOM-killed to flush their target — destroying data for an unrelated failure.

```suggestion
    const recent = logs.join('\n');
```

Also worth appending `(exit code ${code})` to the BUSYKEY branch, which currently drops the code entirely, and giving `handleData` a partial-line carry-over buffer — `chunk.toString().split('\n')` can split the token across two entries.

jamby77 (commented):
Automated review of the fork-aware functions / BUSYKEY changes. 10 findings inline, most severe first; three are marked non-blocking.

The headline issue is that the `block_command` literal does not match how RedisShake v4.6.0 names function entries, so the filter never fires.

apps/api/src/migration/migration-execution.service.ts
```diff
@@ -51,6 +51,18 @@ export class MigrationExecutionService {
     const targetClusterSection = (targetInfo as Record<string, Record<string, string>>).cluster ?? {};
     const targetIsCluster = String(targetClusterSection['cluster_enabled'] ?? '0') === '1';
 
+    // Server-side functions use engine-specific globals (e.g. Valkey's `server`) and
+    // fail to load on a different fork. When source and target are different engines,
+    // exclude functions from the RedisShake stream so the key data still migrates.
+    const sourceDbType = sourceAdapter.getCapabilities().dbType;
+    const targetDbType = targetAdapter.getCapabilities().dbType;
+    const excludeFunctions = sourceDbType !== targetDbType;
+    if (excludeFunctions) {
```
jamby77: **The exclusion is invisible to the user.** This only reaches the API console; nothing is written to the job, so a user who starts execution without reading the analysis (which `startExecution` does not require) sees `completed` / 100% and learns functions are missing from production `FCALL` errors. Pushing a line into `job.logs` would surface it in `ExecutionLogViewer`.

The log also fires for `mode: 'command'`, which never receives `excludeFunctions` and does no filtering — computing the flag inside the redis_shake branch would avoid the false lead.

apps/api/src/migration/analysis/compatibility-checker.ts
```diff
@@ -125,6 +125,20 @@ export function checkCompatibility(
     });
   }
 
+  // 1b. Cross-fork functions are not migrated
+  if (source.dbType !== target.dbType) {
```
jamby77: **The warning fires without checking whether the source has any functions** — nothing in analysis issues `FUNCTION LIST`. Since `warningCount` gates the green banner in `VerdictSection.tsx`, a plain Redis→Valkey analysis with zero function libraries can now never report "No compatibility issues found", which trains users to ignore the panel.

apps/api/src/migration/migration-execution.service.ts
```diff
@@ -51,6 +51,18 @@ export class MigrationExecutionService {
     const targetClusterSection = (targetInfo as Record<string, Record<string, string>>).cluster ?? {};
     const targetIsCluster = String(targetClusterSection['cluster_enabled'] ?? '0') === '1';
 
+    // Server-side functions use engine-specific globals (e.g. Valkey's `server`) and
+    // fail to load on a different fork. When source and target are different engines,
+    // exclude functions from the RedisShake stream so the key data still migrates.
+    const sourceDbType = sourceAdapter.getCapabilities().dbType;
+    const targetDbType = targetAdapter.getCapabilities().dbType;
```
jamby77: **The cross-fork rule now exists in two independent copies** — here, and as `source.dbType !== target.dbType` in `compatibility-checker.ts:129` where it drives the user-facing promise. Refining one (e.g. making it direction-aware) leaves the report describing behavior the executor doesn't perform; no test ties them together. Worth one exported `shouldExcludeFunctions(sourceDbType, targetDbType)` used by both.

apps/api/src/migration/execution/toml-builder.ts
```diff
@@ -97,6 +113,7 @@ export function buildSyncReaderToml(
   options: SyncReaderOptions = {},
   targetIsCluster: boolean = false,
   rsOptions: RedisShakeOptions = {},
+  excludeFunctions: boolean = false,
```
jamby77: **`excludeFunctions` lands at a different position in each builder** (6th on `buildScanReaderToml`, 7th here), giving call sites three mutually type-compatible booleans — swapping `sourceIsCluster` / `targetIsCluster` / `excludeFunctions` compiles clean. The spec's `expect(call[call.length - 1]).toBe(true)` also silently asserts the wrong parameter as soon as an 8th is appended; a builder options object plus `objectContaining({ excludeFunctions: true })` fixes both.

apps/api/src/migration/migration-execution.service.ts
```diff
@@ -221,6 +234,24 @@ export class MigrationExecutionService {
...
+   * log tail for well-known failure signatures so the UI can tell the user what to
+   * do next, instead of a bare "exited with code N".
+   */
+  private explainRedisShakeFailure(code: number | null, logs: string[]): string {
+    const recent = logs.slice(-80).join('\n');
+    if (/BUSYKEY/i.test(recent)) {
+      return (
+        'Migration failed: the target already contains one or more of the keys being ' +
+        'migrated (BUSYKEY). RedisShake will not overwrite existing keys. Enable the ' +
+        '"Flush target before migration" option to clear the target first, or point the ' +
```
jamby77: _Non-blocking._ **This hard-codes the frontend checkbox label** (`MigrationPage.tsx:303`) into the service with no compiler or test link, so a pure UI rename leaves the API naming an option that no longer exists. The module already has a tested home for log interpretation in `execution/log-parser.ts` (already imported here) — better still, return a structured `{ code: 'BUSYKEY' }` and let the web layer own the remediation copy, as the analysis `Incompatibility` shape does.

apps/api/src/migration/execution/toml-builder.ts
```diff
@@ -33,6 +33,21 @@ function validateHost(host: string): string {
...
 
+/**
+ * When source and target are different engines (e.g. Valkey -> Redis), server-side
+ * function libraries must not be replicated: they use engine-specific globals (such
+ * as Valkey's `server`) and fail to load on the other fork, aborting the migration.
+ * Blocking the FUNCTION command drops FUNCTION LOAD/RESTORE from the stream so the
+ * key data still migrates. Returns an empty string (no filter) when not excluding.
+ */
+function buildFilterSection(excludeFunctions: boolean): string {
+  if (!excludeFunctions) return '';
```
jamby77: Repo `CLAUDE.md` ("Coding standards") states *"Don't use one line loops or conditionals."* and *"Use explicit boolean checks instead of negations."* — this line breaks both, and is the only unbraced one-liner in the file (the guards at lines 20 and 54 use blocks).

```suggestion
  if (excludeFunctions === false) {
    return '';
  }
```

jamby77:
Review notes — three things, the first one is a blocker.

### 1. `apps/api/src/migration/execution/toml-builder.ts:44` — `block_command = ["function"]` never matches

RedisShake 4.6.1 (the version pinned in `Dockerfile:91`) filters with an exact, case-sensitive `slices.Contains(config.Opt.Filter.BlockCommand, e.CmdName)`. `CmdName` comes from `strings.ToUpper(argv[0])`, and because `FUNCTION` is in the `containers` map it gets expanded to `<CMD>-<SUBCMD>`. The RDB path emits `Argv = ["function", "load", payload]`, so the entry's `CmdName` is literally `FUNCTION-LOAD`. The lowercase `"function"` matches neither `FUNCTION` nor `FUNCTION-LOAD`.

Net effect: on a Valkey→Redis `redis_shake_sync` migration with a function library, the `[filter]` block is written, RedisShake accepts it silently (no unknown-key error), the `FUNCTION LOAD` entry passes through unchanged, the target rejects it, and the migration aborts exactly as before. The unit tests only assert the literal string is present in the generated TOML, so they stay green while the behaviour is unchanged.

Suggested value:

```toml
block_command = ["FUNCTION-LOAD", "FUNCTION-RESTORE", "FUNCTION-DELETE", "FUNCTION-FLUSH"]
```

`block_command_group = ["SCRIPTING"]` also works but additionally drops `EVAL`/`SCRIPT`, so the explicit list is safer.

### 2. `apps/api/src/migration/analysis/compatibility-checker.ts:129` — warning text asserts something that doesn't happen, and fires unconditionally

The detail string says functions "are automatically excluded from the migration", which is not true given (1) — and won't be true for `scan_reader` or command mode even after (1) is fixed, since those never transfer functions at all.

Separately, the issue is pushed for any `source.dbType !== target.dbType` with no check that the source actually has function libraries (nothing under `apps/api/src/migration/` enumerates `FUNCTION LIST`/`FUNCTION DUMP`). So every Redis→Valkey analysis of an instance with zero functions now returns a warning and bumps `warningCount` (`migration.service.ts:443`), turning a previously clean report into a warning report.

### 3. `apps/api/src/migration/migration-execution.service.ts:211` — classify on `close`, not `exit`

`runRedisShake` resolves on `proc.on('exit')`, but `explainRedisShakeFailure(code, job.logs)` depends on log content delivered by the `stdout`/`stderr` `'data'` handlers. Node only guarantees stdio is drained at `'close'`. The BUSYKEY text is emitted by `log.Panicf` immediately before `os.Exit(1)`, i.e. it's the very last thing written — the output most likely to still be sitting in the pipe when `'exit'` fires. That makes the new message intermittent rather than deterministic. Resolving on `'close'` (or awaiting both `exit` and stream end) fixes it.

### Checked and fine

- `block_command` is the correct key and `[filter]` a valid section for 4.6.1; section ordering before `[advanced]` is fine.
- `sanitizeLogLine` does not strip the `BUSYKEY` token.
- The 80-line tail window is adequate — `log.Panicf` writes the BUSYKEY line first in a single multi-line record, then exits.
- Positional args for `buildScanReaderToml` / `buildSyncReaderToml` are correct at both call sites; `dbType` is a closed union so cross-fork detection can't misfire on `undefined`.
- `MigrationPage.tsx` `block: 'end'` — the two `validationRef` divs are mutually exclusive by phase and the ref is populated before the effect runs.

KIvanow:
@jamby77 thank you for the thorough review! All of the issues should have been fixed now

jamby77:
### Code review

Baseline verified before reviewing: all 12 migration suites (173 tests) pass on the PR head, and `tsc --noEmit` is clean on `apps/api`. All call sites of the two re-signatured TOML builders and of `checkCompatibility` were updated.

**One blocker, seven follow-ups.**

#### Blocker: the durable exclusion notice never reaches the user

`getExecution` returns `[...job.notices, ...job.logs]`, but the sole consumer renders `logs.slice(-500)`:

https://github.com/BetterDB-inc/monitor/blob/92d7c3275939194246d2bfb90d9a8f626c757d32/apps/web/src/components/migration/ExecutionLogViewer.tsx#L36-L40

Once a long run fills `MAX_LOG_LINES = 500`, the array is 501 entries and `slice(-500)` drops index 0 — exactly the prepended notice. The API-side test (`keeps the exclusion notice even after the log cap rolls over`) asserts on `getExecution` rather than the viewer, so the regression is invisible to CI. Even below the cap the notice is buried: the pane is `h-64` and auto-scrolls to the bottom on every poll, so a top-of-list line scrolls out of sight within ~15 lines.

A `notices: string[]` field on `MigrationExecutionResult`, rendered as a banner above the log pane, would deliver what the commit message promises.

https://github.com/BetterDB-inc/monitor/blob/92d7c3275939194246d2bfb90d9a8f626c757d32/apps/api/src/migration/migration-execution.service.ts#L366-L374

#### Follow-ups

1. Execution-time notice fires even when the source has no functions. `checkCompatibility` is deliberately gated on `sourceHasFunctions` so a clean instance keeps its "no issues found" report; the executor has no such gate, so any Valkey to Redis run emits the exclusion notice. A user who just saw a warning-free analysis then gets a scary notice about functions they never had.

https://github.com/BetterDB-inc/monitor/blob/92d7c3275939194246d2bfb90d9a8f626c757d32/apps/api/src/migration/migration-execution.service.ts#L133-L141

2. `/BUSYKEY/i` scans the whole retained buffer, not the fatal tail. RedisShake's writer logs per-key errors without always aborting, so a mid-stream BUSYKEY plus an unrelated fatal (target OOM, connection reset, disk full) makes `job.error` the flush-the-target advice while the genuine cause never surfaces — and the user flushes their target and retries into the same failure. Anchoring on the last non-empty line, or on a `panic:`/`FATAL`-prefixed line, avoids this.

https://github.com/BetterDB-inc/monitor/blob/92d7c3275939194246d2bfb90d9a8f626c757d32/apps/api/src/migration/execution/log-parser.ts#L91-L99

3. `RedisShakeFailureCode` never reaches a consumer. The JSDoc says the structured `code` exists "so the web layer can key its own remediation copy off the code instead of matching on prose", but the only call site takes `.message` and discards `.code`, and `MigrationExecutionResult` has no field to carry it.

https://github.com/BetterDB-inc/monitor/blob/92d7c3275939194246d2bfb90d9a8f626c757d32/apps/api/src/migration/execution/log-parser.ts#L71-L79

4. The new line buffer fixes newline splits but not multi-byte splits: `buffer += chunk.toString()` decodes each chunk independently, so a UTF-8 sequence straddling a chunk boundary becomes replacement characters before it reaches the buffer. The flaw predates this PR, but the buffer was added specifically for chunk-boundary correctness, so `proc.stdout.setEncoding('utf8')` (or a `StringDecoder` per stream) is the natural finish.

https://github.com/BetterDB-inc/monitor/blob/92d7c3275939194246d2bfb90d9a8f626c757d32/apps/api/src/migration/migration-execution.service.ts#L208-L216

5. `block: 'end'` contradicts its own comment. The effect fires on the transition into `phase === 'validating'`, when the panel has just mounted and is short, parking its bottom edge at the viewport bottom; since it never re-runs, everything rendered afterwards grows below the fold. For a panel taller than the viewport, `'end'` puts the header and controls above the fold — the opposite of "its controls/results are visible immediately". `'start'` (the previous behavior) or `'nearest'` keeps the top anchored.

https://github.com/BetterDB-inc/monitor/blob/92d7c3275939194246d2bfb90d9a8f626c757d32/apps/web/src/pages/MigrationPage.tsx#L106-L114

6. The warning is direction-gated, but two of the three execution modes drop functions in every direction. `redis_shake` (scan_reader) iterates keys via SCAN, and `command` mode has no `FUNCTION` handling at all — only `redis_shake_sync` carries libraries via the RDB/AOF stream. A Valkey to Valkey user on scan or command mode gets a clean report and silently loses their libraries. The mode is chosen after analysis so the report can't be fully mode-aware, but attributing the loss purely to fork direction is incomplete.

https://github.com/BetterDB-inc/monitor/blob/92d7c3275939194246d2bfb90d9a8f626c757d32/apps/api/src/migration/analysis/compatibility-checker.ts#L131-L139

7. The `FUNCTION LIST` probe fails silently in exactly the cases that matter. The bare `catch {}` treats "no permission" identically to "no functions": an ACL user without `function|list`, or a cluster source where the keyless command can't be routed, yields `sourceHasFunctions = false` and suppresses the warning — while the executor still writes the `block_command` filter and drops the libraries. Warning on a thrown error, and staying quiet only on a genuinely empty result, closes this.

https://github.com/BetterDB-inc/monitor/blob/92d7c3275939194246d2bfb90d9a8f626c757d32/apps/api/src/migration/migration.service.ts#L407-L415

Items 1, 2 and 7 share a shape worth naming: "couldn't determine" is being treated as "determined negative". Same pattern shows up in #380.

jamby77:
`license-signing-2026-01.pub` (+9, repo root) looks like an accidental carry-over — it is unrelated to the migration changes here.

Provenance seems to be the generation snippet in `proprietary/entitlement/.env.example:19-20`:

```
openssl genrsa -out license-signing-2026-01.pem 2048
openssl rsa -in license-signing-2026-01.pem -pubout   # embed in monitor
```

which produces exactly this filename. `*.pem` is gitignored so the private half was not exposed, but the `.pub` is not covered and got picked up.

Nothing consumes it: no code reads a root `.pub`, and `docs/offline-licenses.md` says the public key belongs in `proprietary/licenses/license-signing-keys.ts` keyed by kid (which is what `license-token.verifier.ts` imports). Merged as-is, master gains a dangling key artifact with no way to tell whether it is active or stale.

Could you drop it from the branch? Worth adding `*.pub` to `.gitignore` alongside `*.pem` so the generation step cannot leave one behind again.

jamby77:
### Code review

Re-reviewed at head `3a8410fc`. **The blocker and all seven follow-ups from the previous round are addressed**, and I verified the round-1 `block_command` fix against the pinned RedisShake source rather than the generated TOML. Three items left, none blocking.

#### Verified against RedisShake v4.6.1 (the version built in `Dockerfile:97`)

- `FUNCTION` is in the `containers` map ([`internal/commands/table.go:10`](https://github.com/tair-opensource/RedisShake/blob/v4.6.1/internal/commands/table.go#L10)), so `CalcKeys` expands `CmdName` to `FUNCTION-LOAD` — the new upper-case entries match.
- The RDB path emits exactly `["function", "load", payload]` ([`internal/rdb/rdb.go:148`](https://github.com/tair-opensource/RedisShake/blob/v4.6.1/internal/rdb/rdb.go#L148)), and `filter.Filter(e)` runs on RDB-derived entries too ([`cmd/redis-shake/main.go:223`](https://github.com/tair-opensource/RedisShake/blob/v4.6.1/cmd/redis-shake/main.go#L223)), so the filter does reach them.
- `rdb_restore_command_behavior` defaults to `panic` ([`internal/config/config.go:68`](https://github.com/tair-opensource/RedisShake/blob/v4.6.1/internal/config/config.go#L68)), so BUSYKEY genuinely aborts the run — the premise of the whole classification path holds.
- Cluster targets are covered: `redis_cluster_writer.go:45` builds standalone writers per node, so BUSYKEY handling applies there too.

---

#### 1. The fatal-line markers never match real RedisShake output

`findFatalLine` searches for `panic:`, `[PANIC]` or `FATAL`, and both the doc comment and the test at `apps/api/src/migration/__tests__/log-parser.spec.ts:126` assert that `log.Panicf` "raises a real Go panic, so Go appends a goroutine stack dump".

It doesn't — [`internal/log/func.go`](https://github.com/tair-opensource/RedisShake/blob/v4.6.1/internal/log/func.go):

```go
func Panicf(format string, args ...interface{}) {
	frames := stack.Trace()
	errMsg := fmt.Sprintf(format, args...)
	for _, frame := range frames {
		...
		errMsg += fmt.Sprintf("\n\t\t\t%v -> %n()", frameStr, frame)
	}
	logger.Error().Msg(errMsg)
	os.Exit(1)
}
```

No Go panic and no runtime stack dump — a single zerolog `Error()` whose ConsoleWriter level token is `ERR`, followed by `os.Exit(1)`. The logger is also built as `.With().Timestamp().Logger()` with no `.Caller()` ([`internal/log/init.go`](https://github.com/tair-opensource/RedisShake/blob/v4.6.1/internal/log/init.go)), so real lines carry no `file.go:NN` at all. Actual output for the case this PR targets:

```
2026-08-18 10:00:00 ERR [src-0] redisStandaloneWriter received BUSYKEY reply. cmd=[RESTORE ...]
			RedisShake/internal/writer/redis_standalone_writer.go:168 -> processReply()
```

So the marker branch is dead code against the pinned binary and classification survives only on the fallback: the appended frames all match the `\.go:\d+` guard and are dropped, `os.Exit(1)` follows immediately, so the `ERR …` line happens to be the last remaining candidate. That is accidental — anything the status goroutine emits in the window between `Msg()` and `Exit()` breaks it.

Suggest anchoring on `\bERR\b` alongside the existing markers, and replacing the fabricated-format test with one built from real v4.6.1 output. This is the same shape as the round-1 `block_command = ["function"]` bug: tests stay green over a log format the binary never produces.

Related, and only worth a line: follow-up 2's scenario cannot occur in 4.6.1. Mid-stream BUSYKEY goes to `log.Debugf` under `behavior == "skip"` (invisible at `log_level = "info"`) and aborts under `"panic"` — there is no "mid-stream BUSYKEY plus a different fatal" state to disambiguate. The fix is harmless, just modelled on behaviour that isn't there.

#### 2. `failureCode` still has no consumer

`git grep failureCode` hits only `apps/api/src/migration/execution/execution-job.ts:11`, `apps/api/src/migration/migration-execution.service.ts:275` and `:382`, and `packages/shared/src/types/migration.ts:176`. Nothing under `apps/web` reads it — `apps/web/src/components/migration/ExecutionPanel.tsx:177` still renders the `execution.error` prose.

The field's own JSDoc says it exists "so the web layer can key its own remediation copy off the code instead of matching on prose", which is exactly what follow-up 3 asked for and exactly what still isn't happening. Either wire a BUSYKEY-specific affordance into the failure banner (a "Flush target and retry" control is the obvious one, since the message already tells the user to go find that option), or hold the field until there's a consumer.

#### 3. Duplicated cluster-address parsing

`apps/api/src/migration/function-presence.ts:parseNodeAddress` is a line-for-line copy of the inline block at `apps/api/src/migration/migration.service.ts:167-175`, and the analysis path wasn't switched over to the new helper. Two copies of the IPv6-bracket-stripping logic will drift. Export the one in `function-presence.ts` and use it in both.

---

#### Out of scope — worth its own issue

`detectSyncStage` is untouched by this PR, but while I had the source open: its patterns largely miss v4.6.1. The real reader messages are `start receiving RDB. path=[…]`, `start sending RDB to target` and `send RDB finished`, so `'send rdb'` matches both the start and the finish line and `send RDB finished` classifies as `rdb_syncing` instead of advancing to `aof_replicating`; `'rdb send finished'`, `'full sync done'` and `'incr sync'` match nothing. That's what the file's existing TODO is for — happy to open a ticket.

Otherwise the branch is clean: `license-signing-2026-01.pub` is gone and `*.pub` is gitignored.

apps/api/src/migration/__tests__/migration-execution.service.spec.ts
```diff
@@ -32,17 +34,18 @@ jest.mock('../execution/command-migration-worker', () => ({
...
+  const targetDbType = overrides?.targetDbType ?? 'valkey';
 
   const mockSourceAdapter = {
     getCapabilities: jest.fn().mockReturnValue({ dbType: 'valkey', version: '8.1.0' }),
     getInfo: jest.fn().mockResolvedValue({ cluster: { cluster_enabled: sourceCluster ? '1' : '0' } }),
     getClient: jest.fn().mockReturnValue({ quit: jest.fn() }),
   };
   const mockTargetAdapter = {
-    getCapabilities: jest.fn().mockReturnValue({ dbType: 'valkey', version: '8.1.0' }),
+    getCapabilities: jest.fn().mockReturnValue({ dbType: targetDbType, version: '8.1.0' }),
```
jamby77, line 61: **Blocker: the notice test passes for the wrong reason, so the feature is effectively untested.**

`getClient()` on both mock adapters (lines 45 and 50) returns `{ quit: jest.fn() }` with no `call` method. `probeSourceFunctions` does `client.call('FUNCTION', 'LIST')`, which throws `TypeError`, gets caught, and returns `'unknown'` — and since `presence !== 'absent'` pushes the notice, the assertion passes through the error path.

That leaves both interesting paths uncovered:
- `'present'` — the notice firing because functions actually exist.
- `'absent'` — the *suppression* gate, which is the thing the last review asked for. Nothing proves a clean source stays notice-free.

Adding `call: jest.fn()` to the mock and driving it through all three presence values would close it.

jamby77 (commented):
Round-3 review. Everything from the previous round is properly addressed — I checked the code at head, not the commit messages. The container-expanded `FUNCTION-LOAD`/`FUNCTION-RESTORE` literals, the options-object builders, `'close'` instead of `'exit'` (with per-stream carry-over buffers and UTF-8 decoding — further than I asked), the shared `shouldExcludeFunctions`, the probe-gated warning, the structured `failureCode`, and `notices` living outside the log cap all look right. The `log-parser` suite classifying off real ANSI-coloured v4.6.1 output is a nice touch.

One blocker and three follow-ups, inline.

**Not inline** (the lines aren't in this diff): step 3.5's target `FLUSHALL` in `migration-execution.service.ts:56` runs *before* `findRedisShakeBinary()` (step 4), `evictOldJobs()` (step 6) and `writeFileSync(tomlPath)` — and `evictOldJobs()` throws `ServiceUnavailableException` when all ten slots hold non-terminal jobs. So a throw anywhere in there means the target is wiped with no migration started and nothing to repopulate it. The ordering is pre-existing, but this PR is what puts a one-click button on that path, so moving the flush to after job creation + binary/TOML resolution belongs here.

apps/api/src/migration/function-presence.ts
```diff
@@ -0,0 +1,93 @@
...
+        tls: config.tls ? {} : undefined,
+        lazyConnect: true,
+      });
+      try {
+        await client.connect();
+        return await probeSourceFunctions(client);
+      } catch {
+        return 'unknown';
+      } finally {
+        try { await client.quit(); } catch { /* ignore */ }
```
jamby77: **Connection leak when a master is unreachable.**

When `client.connect()` fails, the client lands in `reconnecting`, not `end`. `quit()` then queues into the offline queue and only rejects once `maxRetriesPerRequest` (20) is exhausted — and `manuallyClosing` is only set when the command actually reaches the stream (`iovalkey/built/Redis.js:401`), so it never gets set here. The client keeps reconnecting every 2s for the lifetime of the process, one per unreachable master, per execution start.

`client.disconnect()` sets `manuallyClosing` synchronously (`Redis.js:224`) and tears down properly. Alternatively construct these probe clients with `retryStrategy: () => null`.

Two smaller things in the same block: `new Valkey({...})` sits outside the `try`, so a constructor throw rejects the `Promise.all` and propagates out of `startExecution` *after* `jobs.set(id, job)` — leaving a permanently `pending` job that `evictOldJobs()` never reclaims. And `tls: config.tls ? {} : undefined` drops any CA/servername material the source config carries.

apps/api/src/migration/execution/log-parser.ts
```diff
@@ -72,6 +84,70 @@ function detectSyncStage(line: string): SyncStage | null {
...
+  // -> goexit()"). They aren't the cause, and dropping them keeps the fallback from
+  // returning a bare frame.
+  const candidates = logs
+    .map((line) => line.trim())
+    .filter((line) => line.length > 0 && !/\.(go|s):\d+/.test(line));
+  if (candidates.length === 0) {
+    return null;
+  }
+  for (let i = candidates.length - 1; i >= 0; i--) {
+    if (/\bERR\b|panic:|\[PANIC\]|\bFATAL\b/i.test(candidates[i])) {
```
jamby77: **`\bERR\b` with the `/i` flag also matches a bare `err` token.**

`err=nil` in a routine `INF` line satisfies `\bERR\b` case-insensitively (`=` is a non-word char). Since `findFatalLine` walks backwards and returns the first match, any teardown or progress line carrying `err=` after the real fatal line wins — BUSYKEY falls through to `UNKNOWN`, and the new "Flush target & retry" button never appears in exactly the case it was built for.

Suggest splitting the test: match the level token case-sensitively, e.g. `/(^|\s)ERR(\s|$)/`, and keep `panic:` / `[PANIC]` / `FATAL` on a separate case-insensitive check.

apps/api/src/migration/migration-execution.service.ts
```diff
@@ -119,16 +122,46 @@ export class MigrationExecutionService {
...
+        // and the user only discovers missing functions from later FCALL errors.
+        // Gate it on the source actually having (or maybe having) functions, so a
+        // clean instance that just saw a warning-free analysis doesn't get a scary
+        // "functions excluded" message about functions it never had. The filter is
+        // written regardless, so 'unknown' (probe failed) still warrants the notice.
+        // Cluster-aware: FUNCTION LIST is node-local, so a clustered source is probed
+        // per master, matching the analysis warning.
+        const presence = await probeSourceFunctionsClusterAware(sourceAdapter, sourceConfig, clusterEnabled);
+        if (presence !== 'absent') {
+          const notice = `Cross-engine migration (${sourceDbType} → ${targetDbType}): server-side functions are excluded and will not be transferred to the target.`;
```
jamby77: _Non-blocking._ **The execution notice covers only half of what the analysis warns about.**

This fires only when `excludeFunctions` is true (Valkey → Redis). But `compatibility-checker.ts` also warns "Functions require Sync mode to migrate" for same-engine scan/command mode — and that case gets no notice at execution time. A user who starts a Valkey → Valkey scan migration without running analysis first silently loses their function libraries, which is the same failure mode this PR closes for the cross-fork direction.

Also worth noting this probe is awaited inside `startExecution`, before the controller returns the job id — on a cluster source with a wedged master that's up to iovalkey's 10s `connectTimeout` added to the POST. Moving it into `runRedisShake` would keep the start call responsive; the notice would just land a beat later.

jamby77 (approved):
Approving on `08ca1a43`. All four round-3 findings are addressed, and I verified each against the code rather than the commit messages.

- **Probe leak** — `retryStrategy: () => null` + `maxRetriesPerRequest: 1` and `disconnect()` in place of `quit()`, with the constructor moved inside the `try`. Applied to the pre-flush clients too, which were the worse pre-existing case.
- **Flush ordering** — probe and flush both moved into `runRedisShake`, after the job exists and the TOML is written. A flush failure now finalizes through the existing catch/finally instead of 500-ing with a wiped target and no job to show for it.
- **`ERR` matching** — `/(^|\\s)ERR(\\s|$)/` case-sensitive, with `panic:`/`FATAL` split into their own case-insensitive test.
- **Notice test** — the mock now carries `call`, driven across `present`/`absent`/`throw`.

Nice catch on the cancel window your own restructure opened: with the probe and flush moved after the POST returns, the job sits `pending` with no process, so `stopExecution` had nothing to kill. The per-await re-checks plus the test asserting neither `spawn` nor `flushall` fires are exactly right.

**Correction to my previous review:** I claimed the `'absent'` suppression path had zero coverage. That was wrong — it already had a test with its own inline mock. Only the `'present'` path was passing for the wrong reason. Sorry for the noise.

Verified CI green, and ran `src/migration/__tests__` five times locally — 199/199 each time, so the new 30ms-timer tests aren't flaky.

Two non-blocking items left, fine as follow-ups:
1. Same-engine scan/command mode still gets no execution-time notice — `compatibility-checker` warns that only Sync carries functions, but a user who skips analysis and runs Valkey → Valkey in scan mode still loses them silently. Same failure mode this PR closes for the cross-fork direction.
2. The two new `detail` strings use inline backticks (\`server\`), but `VerdictSection.tsx:84` renders `detail` as plain text in a `<p>` — they'll show as literal backticks. No other incompatibility detail does that.

PR #379, opened 2026-08-12
refactor(docker): merge into a single multi-target Dockerfile

KIvanow:
## Summary
Merge `Dockerfile` and `Dockerfile.prod` into a single multi-target Dockerfile. Follow-up to the container-hardening PR (#377), which grew the two files to ~160 byte-identical lines with no shared source of truth — drift between them had already caused one regression.

## What changed
Consolidated into one file with shared stages and two terminal targets:

| Stage | Role |
|-------|------|
| `builder` | deps + web/api build (shared) |
| `redisshake-builder` | from-source migration binary, four coupled pins (shared) |
| `cleanup` | strips AI deps + devDeps (no-ai only) |
| `runtime-base` | apk/npm, user, redis-shake, env, healthcheck, CMD (shared) |
| `production` | `FROM runtime-base`, payload from `builder` → **AI image** |
| `production-no-ai` | `FROM runtime-base`, payload from `cleanup` → **no-ai image** |

CI (`docker-publish.yml`) and the `package.json` docker scripts now select the image via `--target` instead of `-f Dockerfile.prod`; `Dockerfile.prod` is deleted. **Tag scheme unchanged.** The shared stages also let the two CI builds reuse one cache instead of the current disjoint `with-ai`/`without-ai` scopes.

## Verification (both targets, amd64, local)
- `production`: **1.18 GB**, AI deps present, 10 proprietary modules load, 0 Critical/High (grype)
- `production-no-ai`: **739 MB**, AI deps stripped, `AI_ENABLED=false` + `NODE_PATH` set, 0 Critical/High
- Both: `redis-shake` reports v4.6.1, boot with no module-resolution errors, busybox wget healthcheck exits 0
- `docker build --check` clean on both targets

No behaviour or tag changes — purely a structural de-duplication.

## Checklist
- [ ] Unit / integration tests added _(n/a — Docker only)_
- [ ] Docs added / updated

---

> [!NOTE]
> **Medium Risk**
> Changes which image `latest` points to and how releases are built/published; runtime behavior is intended to be unchanged but mis-tagged pulls could affect deployments.
> 
> **Overview**
> **Consolidates** `Dockerfile` and `Dockerfile.prod` into one multi-stage file with shared `builder`, `redisshake-builder`, `cleanup`, and `runtime-base` stages, plus terminal targets `production` (AI) and `production-no-ai` (default last stage). **`Dockerfile.prod` is removed.**
> 
> CI and local scripts now build both variants from `./Dockerfile` via `--target production` / `--target production-no-ai` instead of switching files. **`package.json`** renames scripts so the default local publish is no-ai `latest`, with explicit `*:ai` for the full image.
> 
> **`latest` tagging is adjusted:** the AI metadata step sets `flavor: latest=false` and drops the raw `latest` tag; the no-ai step owns `latest`, only on stable version tags (semver without `-`). Versioned tags (`{{version}}` vs `{{version}}-no-ai`) stay as before.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 324954982712d72a21b9f750e5698ebe51ffbea9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added distinct production container images for AI-enabled and no-AI deployments.
  * Added dedicated commands to build, publish, push, and run each image variant.
  * The no-AI image excludes unnecessary AI and development components for a smaller production footprint.

* **Improvements**
  * The no-AI image receives the `latest` tag only for stable releases; AI images use a separate tag.
  * Both variants share consistent runtime configuration, health checks, and startup behavior.


jamby77:
Review notes. I compared the pre-merge `production` stages of both old Dockerfiles against the new terminal stages instruction by instruction — the merge is faithful. `WORKDIR`, `ENV`, `EXPOSE`, `HEALTHCHECK`, `CMD` and the `betterdb` user all inherit correctly from `runtime-base`; no env var, `redis-shake` copy, chown or `NODE_PATH` was dropped or leaked into the wrong variant. `production-no-ai` is genuinely last, so a bare `docker build` defaults to it as the comment claims. The `#` line inside the `tags:` block scalar is safe — metadata-action parses that input with `comment: '#'`. Nothing else in the repo still references the deleted `Dockerfile.prod` or the old `:no-ai` scripts.

Two things:

### 1. `.github/workflows/docker-publish.yml:57` — `latest` can end up stuck on the AI image

Removing the explicit `type=raw,value=latest` from `meta-with-ai` doesn't stop `latest` being applied there: metadata-action defaults to `flavor: latest=auto`, which auto-adds `latest` for any `type=semver,pattern=...` on a non-prerelease tag. The AI step runs first and publishes `latest` → AI image; the intended outcome only holds because the no-ai push overwrites it a few minutes later.

If the no-ai build fails or is cancelled, `latest` is left permanently pointing at the AI image, which lacks the `AI_ENABLED=false` / `DB_*` / `NODE_PATH` defaults. Adding `flavor: latest=false` to the `meta-with-ai` step makes the intent explicit.

### 2. `package.json:23` — `docker:build:ai` and `docker:run` disagree

`docker:build:ai` tags `betterdb/monitor:ai`, but `docker:run` still runs `betterdb/monitor` (`:latest`, now the no-ai build). So `pnpm docker:build:ai && pnpm docker:run` silently runs the other variant, or pulls from Docker Hub if nothing local matches. Worth adding a `docker:run:ai`.

KIvanow:
@jamby77 thank you for the thorough review! All of the issues should have been fixed now

KIvanow:
@coderabbitai full review

jamby77:
### Code review

The consolidation itself is faithful. I extracted the effective instruction set for each terminal target and compared it against the two originals (comments and ordering normalised):

- `runtime-base` + `production` vs. `master:Dockerfile` production stage — zero instructions added, zero dropped
- `runtime-base` + `production-no-ai` vs. `master:Dockerfile.prod` production stage — zero dropped
- `builder` and `redisshake-builder` — identical apart from comments
- `cleanup` — identical

`ENV`, `EXPOSE`, `HEALTHCHECK`, `CMD` and `USER` all inherit correctly across `FROM runtime-base`, so hoisting them loses nothing, and `production-no-ai` is genuinely the last stage, so the default-target claim holds.

On the `latest` question, the PR body ("tag scheme unchanged") and the Cursor summary ("`latest` tagging is adjusted") disagree, so I checked the live registry: `latest` resolves to `sha256:6fea82aa...`, identical to `0.37.0-no-ai`. `latest` already tracked the no-AI image on master, because the implicit `latest=auto` on the no-ai step overwrote the AI step's push about eight minutes later. **The tagging change is a fix, not a regression** — it removes a real flapping window. No finding raised there.

#### Findings

1. The raw `latest` gate uses `!contains(version, '-')` as a prerelease test. A non-semver `v*` tag such as `vnightly` satisfies both halves of the condition while `type=semver` silently fails to parse, so the run publishes *only* `latest`, from an unreleased commit. The `latest=auto` this replaced would have refused.

https://github.com/BetterDB-inc/monitor/blob/324954982712d72a21b9f750e5698ebe51ffbea9/.github/workflows/docker-publish.yml#L69-L77

2. `pnpm docker:build` flipped from the AI image to `--target production-no-ai`, but the docs still describe it unqualified — `docs/configuration.md:321` sits directly under the `AI_ENABLED`/Ollama config table, and `README.md:355` is the same. Following the docs and then setting `AI_ENABLED=true` yields a container that dies on `Cannot find module 'ollama'`. This is the one I would fix before merge.

https://github.com/BetterDB-inc/monitor/blob/324954982712d72a21b9f750e5698ebe51ffbea9/package.json#L20-L24

3. `meta-with-ai` got an explicit `flavor: latest=false`, but `meta-without-ai` was left on the default `latest=auto`, so `latest` is now produced by two differently-gated mechanisms. Duplicate on stable tags (benign), divergent on the `vnightly` case above.

https://github.com/BetterDB-inc/monitor/blob/324954982712d72a21b9f750e5698ebe51ffbea9/.github/workflows/docker-publish.yml#L64-L72

4. `docker:publish:ai` / `docker:push:ai` push `betterdb/monitor:ai` to the public repo — a fourth tag outside the documented scheme that CI never refreshes, so it will silently rot at whatever version was last pushed by hand. Removing `docker:publish:no-ai` also leaves no local script for the real `X.Y.Z-no-ai` tag.

https://github.com/BetterDB-inc/monitor/blob/324954982712d72a21b9f750e5698ebe51ffbea9/package.json#L25-L29

#### Notes, not findings

- `docs/ai-features.md` (lines 117, 244, 353) instructs users to run `betterdb/monitor:latest` with `AI_ENABLED=true`, which is wrong — confirmed against the registry digests above. Pre-existing, not introduced here, but this PR makes the invariant explicit and enforced, so it is worth fixing in the same pass.
- Stale `Dockerfile.prod` references remain in `docs/analisys/00-overview.md` and `docs/analisys/02-quality-security.md`.
- `docs/configuration.md:328` references `pnpm docker:build:multiarch`, which does not exist in `package.json`. Pre-existing.
- The PR body says the shared stages "let the two CI builds reuse one cache", but the diff leaves `scope=with-ai` and `scope=without-ai` disjoint, so `builder` is still built twice per release. Missed opportunity rather than a defect.
- Comment lines inside the `tags: |` block are a documented `docker/metadata-action` pattern, so those are fine.

Approvable once item 2 is addressed; the rest can follow.

jamby77 (approved):
Approving. The consolidation is verified equivalent target-by-target against both originals, and the `latest` tagging change turned out to be a fix rather than the regression the PR body implies — `latest` already resolved to the no-AI digest on master, so this removes a flapping window.

One thing to land before merge, from my earlier comment: `pnpm docker:build` now targets `production-no-ai`, but `docs/configuration.md:321` (directly under the `AI_ENABLED`/Ollama table) and `README.md:355` still describe it unqualified. Following the docs and setting `AI_ENABLED=true` gives a container that dies on `Cannot find module 'ollama'`. Two lines.

The `vnightly` gate on the raw `latest` tag and the `betterdb/monitor:ai` script are fine as follow-ups.

PR #402, opened 2026-08-19
fix(connect-defaults): keep host.docker.internal on DNS probe timeout…

### LINKED ISSUE #394: connect-defaults: DNS probe timeout misclassifies slow host.docker.internal as unresolvable
Follow-up from PR #380 review ([comment](https://github.com/BetterDB-inc/monitor/pull/380#issuecomment-5296612550)).

### Problem
`resolveDefaultDbHostChecked` treats a DNS lookup of `host.docker.internal` that takes longer than `HOST_RESOLVE_TIMEOUT_MS` (500 ms) as "unresolvable" and falls back to the default gateway. On Docker Desktop the name is genuinely resolvable but the embedded resolver can be slow (cold resolver, contention), so a *slow* lookup is misclassified as a *missing* one and we return the bridge gateway `172.17.0.1`, which on Docker Desktop is the Linux VM, not the macOS/Windows host where the database actually runs.

The root cause is that the probe conflates "timed out" with "does not exist". On timeout the optimistic answer (`host.docker.internal`) is the safer default.

### Options to explore
- Distinguish timeout from NXDOMAIN. `dns.lookup` does not cleanly surface the cause, so this likely means using `dns.resolve`/`dns.promises.Resolver` with its own cancellation, or inspecting the error code.
- On timeout specifically, keep `host.docker.internal` rather than falling through to the gateway.
- Alternatively, raise or make the 500 ms budget configurable, given the answer is cached after the first successful lookup.

### Notes
- Only bites on the *first* probe; subsequent lookups are cached.
- Repro: Docker Desktop with a cold/contended resolver, monitor container on a default bridge.

Relevant code: `apps/api/src/system/runtime.util.ts` (`defaultCanResolveHost`, `resolveDefaultDbHostChecked`).

Shikha-code36:
Fixes #394. `resolveDefaultDbHostChecked` was treating a **timed-out** DNS probe for
`host.docker.internal` the same as a **definitively unresolvable** one, so a slow/cold
resolver on Docker Desktop would cause the default DB host prefill to fall through to the
bridge gateway IP (the Docker Desktop Linux VM) instead of the real host.

## Changes
- `defaultCanResolveHost` now returns a tri-state `HostResolution`
  (`'resolved' | 'not-found' | 'timed-out'`) instead of a boolean, so a timeout can be
  told apart from a definitive NXDOMAIN-style failure.
- `resolveDefaultDbHostChecked` treats `timed-out` the same as `resolved` — it keeps
  `host.docker.internal` on timeout instead of falling through to the
  gateway/loopback disambiguation, which is now reserved for a confirmed `not-found`.
- Added a unit test covering the timed-out case.

## Checklist
- [ ] Docs added / updated
- [ ] Roborev review passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database host detection when running in Docker.
  * Preserves `host.docker.internal` when DNS checks time out or encounter temporary resolver failures.
  * Applies fallback hosts only when the preferred host is definitively unavailable.

* **Tests**
  * Added coverage for successful, unavailable, timed-out, and indeterminate host detection scenarios.

### DIFF
--- a/apps/api/src/system/runtime.util.spec.ts
+++ b/apps/api/src/system/runtime.util.spec.ts
@@ -81,7 +81,7 @@ describe('resolveDefaultDbPort', () => {
 describe('resolveDefaultDbHostChecked', () => {
   const resolves = () => Promise.resolve('resolved' as const);
   const doesNotResolve = () => Promise.resolve('not-found' as const);
-  const timesOut = () => Promise.resolve('timed-out' as const);
+  const indeterminate = () => Promise.resolve('indeterminate' as const);
   // Defaults for the unresolvable branch: bare Docker bridge with a gateway.
   const bridge = {
     isHostNetwork: () => false,
@@ -98,15 +98,16 @@ describe('resolveDefaultDbHostChecked', () => {
     ).resolves.toEqual({ host: 'host.docker.internal', source: 'docker' });
   });
 
-  it('offers host.docker.internal on a timed-out probe instead of falling through to the gateway', async () => {
+  it('offers host.docker.internal on an indeterminate probe (timeout or transient resolver failure) instead of falling through to the gateway', async () => {
     // A cold/contended resolver on Docker Desktop can be slower than the probe
-    // budget even though the name genuinely resolves; misreading that as
-    // "missing" would wrongly return the bridge gateway (the Docker Desktop
-    // Linux VM, not the real host). See issue #394.
+    // budget, or fail transiently (e.g. EAI_AGAIN), even though the name
+    // genuinely resolves; misreading that as "missing" would wrongly return the
+    // bridge gateway (the Docker Desktop Linux VM, not the real host). See
+    // issue #394.
     await expect(
       resolveDefaultDbHostChecked(
         { dbHost: undefined, containerized: true },
-        { canResolveHost: timesOut, ...bridge },
+        { canResolveHost: indeterminate, ...bridge },
       ),
     ).resolves.toEqual({ host: 'host.docker.internal', source: 'docker' });
   });
--- a/apps/api/src/system/runtime.util.spec.ts
+++ b/apps/api/src/system/runtime.util.spec.ts
@@ -79,8 +79,9 @@ describe('resolveDefaultDbPort', () => {
 });
 
 describe('resolveDefaultDbHostChecked', () => {
-  const resolves = () => Promise.resolve(true);
-  const doesNotResolve = () => Promise.resolve(false);
+  const resolves = () => Promise.resolve('resolved' as const);
+  const doesNotResolve = () => Promise.resolve('not-found' as const);
+  const timesOut = () => Promise.resolve('timed-out' as const);
   // Defaults for the unresolvable branch: bare Docker bridge with a gateway.
   const bridge = {
     isHostNetwork: () => false,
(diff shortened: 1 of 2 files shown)


Shikha-code36:

KIvanow (approved):
Thank you for your contribution @Shikha-code36 ! And thank you for dealing with CodeRabbit's feedback. It's experimentational at this point for us, so I'd live to hear your feedback.

On the PR itself - nice fix. The tri-state cleanly separates a slow resolver from a missing name, and keeping host.docker.internal on timeout is the right default, since a timeout means Docker Desktop's resolver is cold rather than the name being gone. One non-blocking nice-to-have: the new test covers the decision in resolveDefaultDbHostChecked, but the timedOut/Promise.race mechanics in defaultCanResolveHost aren't tested directly, so a hanging lookup asserting 'timed-out' and a rejecting one asserting 'not-found' would lock down the trickiest part.

Shikha-code36:
> Thank you for your contribution @Shikha-code36 ! And thank you for dealing with CodeRabbit's feedback. It's experimentational at this point for us, so I'd live to hear your feedback.
> 
> On the PR itself - nice fix. The tri-state cleanly separates a slow resolver from a missing name, and keeping host.docker.internal on timeout is the right default, since a timeout means Docker Desktop's resolver is cold rather than the name being gone. One non-blocking nice-to-have: the new test covers the decision in resolveDefaultDbHostChecked, but the timedOut/Promise.race mechanics in defaultCanResolveHost aren't tested directly, so a hanging lookup asserting 'timed-out' and a rejecting one asserting 'not-found' would lock down the trickiest part.

Thanks @KIvanow! Glad the tri-state approach reads clearly.

On the nice-to-have: agreed it'd lock down the trickiest part, but I'll leave it as
a follow-up rather than block this PR on it, the existing resolveDefaultDbHostChecked
tests already cover the decision that matters (keep host.docker.internal on any
non-definitive outcome), and this is scoped to the specific misclassification bug
from #394.

Re: CodeRabbit  worked well overall. Both flagged comments were legitimate
(the comment-wording nit and the real bug where non-timeout resolver errors like
EAI_AGAIN were wrongly treated as a confirmed "not found"), so it caught something
real rather than noise.

KIvanow:
> > Thank you for your contribution @Shikha-code36 ! And thank you for dealing with CodeRabbit's feedback. It's experimentational at this point for us, so I'd live to hear your feedback.
> > On the PR itself - nice fix. The tri-state cleanly separates a slow resolver from a missing name, and keeping host.docker.internal on timeout is the right default, since a timeout means Docker Desktop's resolver is cold rather than the name being gone. One non-blocking nice-to-have: the new test covers the decision in resolveDefaultDbHostChecked, but the timedOut/Promise.race mechanics in defaultCanResolveHost aren't tested directly, so a hanging lookup asserting 'timed-out' and a rejecting one asserting 'not-found' would lock down the trickiest part.
> 
> Thanks @KIvanow! Glad the tri-state approach reads clearly.
> 
> On the nice-to-have: agreed it'd lock down the trickiest part, but I'll leave it as a follow-up rather than block this PR on it, the existing resolveDefaultDbHostChecked tests already cover the decision that matters (keep host.docker.internal on any non-definitive outcome), and this is scoped to the specific misclassification bug from #394.
> 
> Re: CodeRabbit worked well overall. Both flagged comments were legitimate (the comment-wording nit and the real bug where non-timeout resolver errors like EAI_AGAIN were wrongly treated as a confirmed "not found"), so it caught something real rather than noise.

Thank you @Shikha-code36 ! Makes sense on the follow up and I appreciate the feedback on coderabbit

PR #403, opened 2026-08-20
feat(cluster): cluster-bus CRC-mismatch observability

KIvanow:
## Summary
Surface cluster-bus corruption as a first-class signal, off the back of the valkey-io/valkey#4201 CRC work. The monitor now exposes the CRC-mismatch counter, alerts on any increase, and warns when the check is turned off.

## Changes
- Prometheus gauge `cluster_stats_messages_crc_mismatch`, read from the `CLUSTER INFO` the poller already fetches. No-op on builds without the field.
- Fires `cluster.bus.corruption` on any increase, over OTLP and the Pro webhook, mirroring the `cluster.failover` edge. Baseline-seeded so a pre-existing counter does not fire on startup. New `CLUSTER_BUS_CORRUPTION` event (Pro tier).
- Config-hazard advisory when cluster mode is on but `cluster-crc-enabled` is off.

## Checklist
- [ ] Docs added / updated

---

> [!NOTE]
> **Medium Risk**
> Touches cluster metric polling, event dispatch, and hazard-cache semantics. Incorrect caching or baseline handling could miss or duplicate corruption alerts.
> 
> **Overview**
> Adds first-class observability for Valkey cluster-bus CRC integrity (valkey#4201): a Prometheus gauge, alerts on any counter increase, and a config-hazard when the check is off or cannot be verified.
> 
> **Runtime signal.** `CLUSTER INFO` now feeds `cluster_stats_messages_crc_mismatch`. Any increase (after a first-observation baseline) fires `cluster.bus.corruption` over OTLP and the Pro webhook. Concurrent poller vs `/metrics` updates are coalesced per connection so a stale CRC baseline cannot re-emit the same delta.
> 
> **Config hazard.** `evaluateClusterCrcHazard` flags cluster mode with `cluster-crc-enabled no` as advisory, and a filtered/null `cluster-enabled` read as unverified. Incomplete CRC probes still return AOF findings but are **not TTL-cached**, so the next poll re-probes instead of treating a partial result as clean.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6c65bdf1d44bd6ff2bd82307e361c5e7161bc98b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added detection and warnings when cluster CRC protection is disabled or cannot be verified.
  - Added monitoring for cluster-bus CRC mismatches and corruption events.
  - Added optional webhook notifications with mismatch counts, node details, and connection context.
  - Added a “Cluster Bus Corruption” event label in webhook settings.

- **Bug Fixes**
  - Improved configuration checks to retry incomplete probes instead of caching partial results.
  - Prevented overlapping metric updates and stale CRC baselines from producing inconsistent results.


apps/api/src/monitor/config-hazard.service.ts
```diff
@@ -73,42 +81,68 @@ export class ConfigHazardService {
...
+    let clusterCrcEnabled: string | null;
     try {
-      aclGetUserResult = await client.call('ACL', ['GETUSER', 'default']);
+      clusterEnabled = await client.getConfigValue('cluster-enabled');
+      clusterCrcEnabled = await client.getConfigValue('cluster-crc-enabled');
     } catch (err) {
+      // A failed cluster read must not discard AOF findings already collected,
+      // so return what we have. But the probe is incomplete, so mark it
+      // uncacheable: caching now would mask the missing CRC check as a clean
+      // result for a full TTL and skip re-probing on the next poll.
```
jamby77, line 133: **`getConfigValue` returns null without throwing, so the unverified case is cached as clean.**

The `catch` here only covers a rejected promise. `UnifiedAdapter.getConfigValue` ends with `config[parameter] || null` (`unified.adapter.ts:570`), so a `CONFIG GET` that comes back without the key resolves to `null` — no throw, no `cacheable: false`.

`clusterEnabled === null` then falls through `clusterEnabled !== 'yes'` in `evaluateClusterCrcHazard`, no finding is pushed, and line 145 caches `[]` for the full TTL. We do not know whether cluster mode is on, and we record "clean" for 5 minutes.

That contradicts three things this PR states:

- `evaluateClusterCrcHazard`'s own doc — "Only a positively verified state (cluster off, feature absent, or the check already on) returns null — never a silent false negative". Cluster state *unknown* is a fourth case, and it returns null.
- `ProbeResult.cacheable` — "Only a fully-completed probe may be cached".
- `evaluateAclAofHazard`, which carries the identical contract and honours it with an explicit `unverified` status rather than null.

`clusterCrcEnabled` has the same conflation: the interface doc reads it as "the build predates valkey#4201", but null is equally what a filtered or empty `CONFIG GET` produces, so a probe we could not perform reads as a feature that does not exist.

Reachable on anything that restricts `CONFIG GET` — managed offerings and proxies commonly filter it rather than erroring, which is exactly the throw path already handled.

Suggest treating null as unverified rather than absent: return an `unverified` finding when `clusterEnabled` is null, and mark the probe uncacheable, matching the `'denied'` path in the AOF evaluator.

The rest holds up — advancing `state.previousCrcMismatch` before the `await` closes the re-dispatch window, and seeding the baseline on first observation is the right call for a counter that survives across polls but resets on restart.

jamby77: The `cluster-enabled` half is fully addressed — `unverified` finding, `cacheable: !clusterStateUnknown`, the interface doc now records that `getConfigValue` resolves a filtered reply to null without throwing, and there are tests at both the evaluator and service level covering the re-probe. Thanks.

The `clusterCrcEnabled` half is still open, and it's narrower than what I originally described, so I don't think it should block this PR.

`clusterStateUnknown` only tests `clusterEnabled`, so with `clusterEnabled === 'yes'` and `clusterCrcEnabled === null` we still return null and cache clean. A full `CONFIG GET` filter now trips the new branch first, so what's left needs a *partial* filter — a proxy that allowlists the older, benign `cluster-enabled` but not the newer `cluster-crc-enabled`. Real, but a much smaller surface. Separating "build predates #4201" from "read was filtered" needs a server-version gate off INFO since both return an empty reply, which is its own change rather than a repush here. I'll file it as a follow-up.

One thing worth fixing before merge, since it's a one-liner: the docstring on `evaluateClusterCrcHazard` still says

> Only a positively verified state (cluster off, feature absent, or the check already on) returns null — never a silent false negative.

That's now true for `clusterEnabled` and not true for `clusterCrcEnabled`. Could you soften it or note the known gap, so the next reader doesn't trust a guarantee the code doesn't make? Same for the `clusterCrcEnabled` field doc, which still reads as "null when the build predates valkey#4201" with no mention of the filtered-read case.

Approving on that basis.

jamby77 (approved):
CRC observability looks good. The unverified-vs-clean fix on `cluster-enabled` is correct and well covered by tests, and the Prometheus gauge-child removal plus the compare-and-delete on `updateMetricsInFlight` both hold up.

Approving with one non-blocking ask on the `evaluateClusterCrcHazard` docstring (see thread) — the "never a silent false negative" guarantee no longer holds for `clusterCrcEnabled`. Follow-up filed for the version gate.

jamby77:
Post-merge note, not a defect in this PR — flagging so it is recorded rather
than rediscovered.

While picking up the #410 follow-up I checked the upstream state:

- `valkey-io/valkey#4201` is **open**, `merged=false`, base `unstable`, last
  updated 2026-08-20
- a code search for `cluster-crc-enabled` across `valkey-io/valkey` returns
  **0 hits** — not in `unstable`, only on the PR branch
- latest releases: 9.1.1, 9.0.5, 8.1.9, 8.0.10, 7.2.14 — none carries it

So `CONFIG GET cluster-crc-enabled` returns null on every server that exists
today, which means `evaluateClusterCrcHazard`'s advisory branch — gated on
`clusterCrcEnabled === 'no'` — is currently unreachable. Same for the
`cluster_stats_messages_crc_mismatch` handling in `prometheus.service.ts`, which
reads a CLUSTER INFO field introduced by the same unmerged PR.

Shipping the detector ahead of upstream so it fires the day the feature lands is
a legitimate strategy, and the unverified/uncacheable path added here for
`clusterEnabled === null` is live and useful regardless. The point is only that
the CRC advisory itself cannot be validated against a real server yet, and that
#410 cannot be implemented until the merge decides which release carries the
parameter.

Recorded on #410 too, with a proposal to park it until then.

PR #404, opened 2026-08-20
feat(ai): add @betterdb/ai meta package

jamby77:
## Summary

Adds `@betterdb/ai`, a facade package so the AI stack is one install and one import root instead of five.

```ts
import { AgentCache, SemanticCache, MemoryStore, Retriever } from '@betterdb/ai';
import { BetterDBLlmCache } from '@betterdb/ai/langchain';
import { createOpenAIEmbed } from '@betterdb/ai/embed/openai';
```

Purely additive: no existing package source is modified. The five packages keep their names, versions, release workflows, and direct consumers. The only pre-existing files touched are the five child release workflows, which each gain one trailing job.

**Export surface.** Measured with the TypeScript API against the built `.d.ts` files: the children export 158 names, of which 25 are declared in more than one place. Those 25 are reachable only through the namespaces `agentCache`, `semanticCache`, `retrieval`, `memory`, `searchKit`; the other 133 are flat. Nothing is renamed and nothing is unreachable.

Flattening the 25 would be a correctness hazard rather than a cosmetic one. `ValkeyCommandError` is two different classes — `extends AgentCacheError` in agent-cache, `extends Error` in semantic-cache — so picking one would silently break `instanceof` for the other.

**Versioning.** Children keep independent versions; the facade pins them exactly, resolved from `npm view` at publish time rather than from workspace manifests. The existing workflows write versions into `package.json` during CI without committing them back, which is why `packages/cli/package.json` reads 0.25.0 against npm's 0.38.0 — trusting workspace files would eventually publish a facade pinning a version that does not exist.

## Changes

- `packages/ai` — new package: flat root barrel (133 names + 5 namespaces), 7 framework adapter subpaths, 5 `embed/*` subpaths
- Adapters grouped by framework rather than mirrored per package; `/vercel` renames the children's `/ai` subpath
- `src/__tests__/completeness.spec.ts` — guards that every unambiguous child export reaches the facade, that no flat export shadows an ambiguous name, and that every child `exports` subpath is mirrored or explicitly excluded
- `.github/workflows/ai-facade-release.yml` — `workflow_call`-able release; pins children from npm, guards against a surviving `workspace:` protocol, packs the artifact and type-checks it against the pinned children before publishing
- Trailing `publish-facade` job in the five child release workflows, gated on their existing `verify` job
- `api-tests.yml` runs the facade suite on PRs
- Docs: `packages/ai/README.md`, `docs/packages/ai.md`, `LICENSE`

## Known limitations

- **Automatic bumps are always patch.** A child's breaking release publishes a facade patch pinning it, so consumers needing strict semver should pin `@betterdb/ai` exactly. Documented in the README; worth revisiting when the facade leaves 0.x.
- **Subpath types need modern resolution.** `typesVersions` gets a classic-`node` consumer to the facade's `dist/*.d.ts`, but that file's own re-export of a child subpath then fails, because the children declare no `typesVersions`. Fixing it properly means editing the child packages, which this PR deliberately does not do. Subpath imports need `node16`/`nodenext`/`bundler`; the root import works under classic `node`.
- **A queued facade run can be cancelled.** GitHub keeps one pending run per concurrency group, so three simultaneous child releases can cancel one facade publish. Harmless — any later child release re-pins all five — but it may make a green child release look amber. `continue-on-error` is not a legal key on a reusable-workflow-call job, so working around it would need a wrapper job.
- **PyPI has no equivalent yet.** The Python mirrors have drifted (semantic-cache 0.10.0, retrieval 0.5.0, agent-memory 0.7.0) and need reconciling first.

## Follow-ups worth separate issues

- `packages/semantic-cache/src/embed/google.ts` exports `createGoogleEmbed` but `./embed/google` is missing from that package's `exports` map, so it is unreachable. Not worked around here.
- Third-party GitHub Actions are pinned to mutable tags across 16 of 17 workflow files. A repo-wide pinning pass plus Dependabot for `github-actions` would close it; pinning only the new file would not reduce real exposure, since the callers already run the same unpinned actions with `NPM_TOKEN` in scope.

## First release

The facade seeds its version from the committed manifest on a cold start, so the first automatic publish lands on `0.1.0` without needing a manual dispatch first.

## Checklist
- [x] Docs added / updated

---

> [!NOTE]
> **Medium Risk**
> New npm publish path with NPM_TOKEN and automatic version pinning after every child release. Logic is additive and gated, but a bad pin or cancelled concurrent facade run can ship a mismatched meta-package.
> 
> **Overview**
> Adds `@betterdb/ai` as a one-install facade over agent-cache, semantic-cache, retrieval, agent-memory, and valkey-search-kit. Existing packages are unchanged; consumers can still install them directly.
> 
> The root barrel flattens unambiguous exports and keeps colliding names (e.g. two different `ValkeyCommandError` classes) behind `agentCache` / `semanticCache` / `retrieval` / `memory` / `searchKit` namespaces. Framework adapters are grouped by framework (`/langchain`, `/vercel`, etc.; children’s `/ai` becomes `/vercel`) plus `embed/*` factories.
> 
> Release: each child workflow now calls a reusable facade publish that pins children from `npm view` (not workspace versions), type-checks the packed tarball, then patch-bumps `@betterdb/ai`. Completeness tests and a PR CI job guard the export surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f4c87f59dda2395fd98eebabaa5576d5b1ddb40a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added the `@betterdb/ai` all-in-one package for AI caching, retrieval, memory, and search.
  - Added integrations for LangChain, LangGraph, Vercel AI SDK, OpenAI, Anthropic, and LlamaIndex.
  - Added embedding provider exports for OpenAI, Bedrock, Voyage, Cohere, and Ollama.
  - Added installation, usage, exports, and versioning guidance.

- **Release Automation**
  - Added automated publishing, verification, release creation, and package availability checks.
  - Added automated build and test coverage for the package.


jamby77:
Two bots landed on the same lines with opposite conclusions. Addressed in f4c87f5 — with one disputed.

## Disputed: `process.argv[1]` is not the `-p` flag

Bugbot reports that `node -p "…" "$CURRENT"` reads the flag rather than the version, making every automatic bump `NaN`. That is not how node parses `-e`/`-p`: neither the flag nor the code string enters `argv`, so `argv[1]` is the first positional argument.

```
$ node -p "JSON.stringify(process.argv)" "0.1.0"
["/Users/…/bin/node","0.1.0"]

$ node -p "const [maj,min,patch]=process.argv[1].split('.').map(Number); [maj,min,patch+1].join('.')" "0.1.0"
0.1.1
```

The automatic path was never broken. Not fixing this one.

## Fixed: prerelease versions did produce `NaN`

CodeRabbit flagged the same lines for a different and correct reason. `0.2.0-rc.1` splits to `['0','1','0-rc','1']`, so `Number('0-rc')` is `NaN` and the step emits `0.2.NaN`:

```
$ node -p "…" "0.2.0-rc.1"
0.2.NaN
```

This is reachable rather than theoretical, because the workflow already anticipates prereleases — it sets `prerelease: ${{ contains(…, '-') }}` on the GitHub Release. Publishing `0.2.0-rc.1` by dispatch makes it npm's `latest`, and the next automatic run reads it back.

Rather than pick a bump policy, the step now refuses. Whether a prerelease should be promoted to a stable patch or incremented as a prerelease is a release decision, so it fails loudly and asks for an explicit `workflow_dispatch` version:

```
0.1.0       -> 0.1.1
1.9.9       -> 1.9.10
0.2.0-rc.1  -> refused, exit 1
```

## Fixed: `secrets: inherit` → `NPM_TOKEN` only

`NPM_TOKEN` is now declared under `workflow_call.secrets` and passed explicitly by all five callers. Verified none still inherit, and that every reusable-workflow-call job keeps only schema-legal keys.

## Fixed: `persist-credentials: false`

Added to the facade release checkout, which runs `pnpm install` with `contents: write` in scope.

The 15 sibling workflows have the same pattern and are deliberately out of scope here — pinning or hardening only the new file gives no real protection when the callers run the same unpinned actions with the same token moments earlier. Tracked locally alongside the action-pinning pass.

## Fixed: doc examples were not self-contained

Both namespace examples now declare what they use. While fixing them I found the example I had written called `cache.getLlm(…)`, which does not exist — the real call is `cache.llm.check(…)`. Both examples are now type-checked against the built package with `skipLibCheck: false`, not eyeballed.

## Not actioned: docstring coverage

0% against an 80% threshold. Every file in `packages/ai/src` is a re-export statement — there is nothing to document that the underlying packages do not already document, and adding docstrings to `export { X } from 'y'` would be noise.

KIvanow (approved):
Approving - this is a clean, well-engineered facade. Nice work @jamby77.

What I checked and liked:

- **The release path has real safety, not just a publish step.** Pinning children from `npm view` at publish time (instead of trusting workspace manifests) is the right call, and the packed-artifact validation runs `tsc` with `skipLibCheck: false` so every re-export in the facade's `.d.ts` is type-checked against the actually-pinned children - not just the symbols in `check.ts`. Combined with the `workspace:` guard, a facade that pins a version whose API it doesn't match can't get published.
- **The completeness spec encodes the invariant that matters.** Unambiguous child exports must flatten, ambiguous ones must stay namespace-only - so the two different `ValkeyCommandError` classes stay reachable as `agentCache.ValkeyCommandError` / `semanticCache.ValkeyCommandError` and can never be silently collapsed into a broken `instanceof`. Explicit named re-exports mean drift fails the build, and `ai-facade-tests` runs it on every PR.
- **The `NaN` version bump is fixed** (`node -e '…' "$CURRENT"`), and the auto-bump refuses anything that isn't plain `MAJOR.MINOR.PATCH` so prerelease promotion stays an explicit decision.
- **The `skip_npm` guard is consistent** across all five children, so a dry-run release won't trigger a facade publish.

The PR is also honest about its own limitations in the description, which I appreciate.

A few non-blocking polish items for whenever: derive the adapter-subpath check from the children the way the root barrel is guarded (today it's a hardcoded presence list), drop the whitespace reindent in the five child workflows to keep the diff to just the new job, and optionally set `persist-credentials: false` on those child checkouts to close CodeRabbit's last thread. None of these block merge.

LGTM.

PR #405, opened 2026-08-20
fix(mcp): propagate real HTTP status instead of generic 500 (#155)

### LINKED ISSUE #155: MCP tools return generic 500 for non-existent instanceId override
Calling any MCP tool with a non-existent instanceId override returns a generic 500 error instead of a meaningful response. This happens because the validation layer only checks the format of the ID, not whether it exists. When the lookup fails, the error is swallowed and converted into a generic 500 response

How to reproduce:
Call get_health({ instanceId: 'nonexistent' })

Expected behavior:
Return a clear 404 error indicating the instance does not exist

Actual behavior:
Returns 500 Failed to get health

Scope:
Affects all MCP tools that accept an optional instanceId override

Shikha-code36:
Fixes #155 — every MCP tool that accepts an `instanceId` override returned a generic `500 Internal Server Error` for a non-existent instance instead of a `404`.

`ConnectionRegistry.get(id)` already throws a `NotFoundException` with a clear message when the connection isn't registered, but every handler in `mcp.controller.ts` caught **all** errors indiscriminately and rethrew a hardcoded `HttpException(..., HttpStatus.INTERNAL_SERVER_ERROR)`, discarding the original status. Sibling controllers (`mcp-analytics.controller.ts`, `mcp-ai/mcp-ai.controller.ts`, `vector-search.controller.ts`) already use the `mapMcpError` helper from `mcp-helpers.ts`, which preserves an `HttpException` as-is and only maps unknown errors to 500 — `mcp.controller.ts` just never adopted it.

## Changes

- `mcp.controller.ts`: replace every catch block's manual `logger.error(...) + throw new HttpException(..., INTERNAL_SERVER_ERROR)` with `throw mapMcpError(this.logger, error, '<fallback>', '<log message with connection id>')`, matching the pattern already used elsewhere in the codebase.
- `mcp-helpers.ts`: `mapMcpError` gains an optional 4th `logMessage` param (defaults to `fallback`, so the 8 pre-existing call sites in the sibling controllers are unaffected) so the connection id can stay in the log line without leaking into the client-facing message.
- **Repo-wide response change:** `mapMcpError` no longer appends `: ${error.message}` to the 500 response body for non-HTTP errors — it now returns just the generic `fallback` text. This was flagged as a security/privacy concern (an arbitrary internal error could otherwise reach an authenticated MCP caller verbatim). Since `mapMcpError` is shared, this changes the 500 body shape for `mcp-analytics`, `mcp-ai`, and `vector-search` too, not just this controller. No existing test asserted on the old suffixed text, and all their suites still pass unchanged. The full error detail is still captured via `logger.error(logMessage, error.stack)`.
- `mcp.controller.spec.ts` (new): ports #204's tests — `registry.get()` throwing `NotFoundException`, a service-level `NotFoundException`, and the non-HTTP 500-fallback case — plus a new test asserting the connection id survives in the log line even though the client message stays generic.

## Verification

- Confirmed a regression test fails on the pre-fix code (`500` instead of `404`) and passes after the fix.
- `tsc --noEmit` across the whole `apps/api` project: no new errors.
- `apps/api/src/mcp/__tests__/mcp.controller.spec.ts`, `apps/api/src/mcp/ai/__tests__/mcp-ai.controller.spec.ts`, `apps/api/src/mcp/__tests__/mcp-analytics.controller.spec.ts`: 30/30 tests pass.
- Full `pnpm test` couldn't be run end-to-end in this environment due to a pre-existing, unrelated issue: the `betterdb-test-cluster-init` Docker Compose service fails during global test setup (`Wrong number of arguments for specified --cluster sub command`), which also blocks running the *unmodified* suite. Verified instead via a scoped `ts-jest` run (isolated-modules) against the affected spec files.

## Test plan

- [ ] `pnpm --filter api test src/mcp/__tests__/mcp.controller.spec.ts` once the Docker cluster-init issue is resolved in CI
- [ ] Manually call `get_health({ instanceId: 'nonexistent' })` via the MCP tool and confirm a `404` with a clear message, not a `500`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling across MCP endpoints with consistent HTTP responses and clearer client-facing messages.
  * Prevented internal error details from being exposed in fallback responses.
  * Preserved “not found” responses and capability-related errors.
  * Enhanced diagnostic logging with connection and instance details for easier troubleshooting.

### DIFF
--- a/apps/api/src/mcp/mcp.controller.ts
+++ b/apps/api/src/mcp/mcp.controller.ts
@@ -82,7 +82,7 @@ export class McpController {
       const info = await client.getInfoParsed(sections);
       return info;
     } catch (error) {
-      throw mapMcpError(this.logger, error, 'Failed to get info');
+      throw mapMcpError(this.logger, error, 'Failed to get info', `Failed to get info for ${id}`);
     }
   }
 
@@ -96,7 +96,7 @@ export class McpController {
       const parsedCount = safeLimit(count, 25);
       return await client.getSlowLog(parsedCount);
     } catch (error) {
-      throw mapMcpError(this.logger, error, 'Failed to get slowlog');
+      throw mapMcpError(this.logger, error, 'Failed to get slowlog', `Failed to get slowlog for ${id}`);
     }
   }
 
@@ -106,7 +106,7 @@ export class McpController {
       const client = this.registry.get(id);
       return await client.getLatestLatencyEvents();
     } catch (error) {
-      throw mapMcpError(this.logger, error, 'Failed to get latency');
+      throw mapMcpError(this.logger, error, 'Failed to get latency', `Failed to get latency for ${id}`);
     }
   }
 
@@ -120,7 +120,12 @@ export class McpController {
       ]);
       return { doctor, stats };
     } catch (error) {
-      throw mapMcpError(this.logger, error, 'Failed to get memory diagnostics');
+      throw mapMcpError(
+        this.logger,
+        error,
+        'Failed to get memory diagnostics',
+        `Failed to get memory for ${id}`,
+      );
     }
   }
 
@@ -142,7 +147,7 @@ export class McpController {
       if (msg.includes('unknown command') || msg.includes('COMMANDLOG')) {
         return { entries: [], note: 'COMMANDLOG not available on this instance' };
       }
-      throw mapMcpError(this.logger, error, 'Failed to get commandlog');
+      throw mapMcpError(this.logger, error, 'Failed to get commandlog', `Failed to get commandlog for ${id}`);
     }
   }
 
@@ -152,7 +157,7 @@ export class McpController {
       const client = this.registry.get(id);
       return await client.getClients();
     } catch (error) {
-      throw mapMcpError(this.logger, error, 'Failed to get clients');
+      throw mapMcpError(this.logger, error, 'Failed to get clients', `Failed to get clients for ${id}`);
     }
   }
 
@@ -165,7 +170,12 @@ export class McpController {
       const parsedLimit = limit !== undefined ? safeLimit(limit, MAX_LIMIT) : undefined;
       return await this.metricsService.getSlowLogPatternAnalysis(parsedLimit, id);
     } catch (error) {
-      throw mapMcpError(this.logger, error, 'Failed to get slowlog patterns');
+      throw mapMcpError(
+        this.logger,
+        error,
+        'Failed to get slowlog patterns',
+        `Failed to get slowlog patterns for ${id}`,
+      );
     }
   }
 
@@ -193,7 +203,12 @@ export class McpController {
       if (msg.includes('unknown command') || msg.includes('COMMANDLOG')) {
         return { entries: [], note: 'COMMANDLOG not available on this instance' };
       }
-      throw mapMcpError(this.logger, error, 'Failed to get commandlog history');
+      throw mapMcpError(
+        this.logger,
+        error,
+        'Failed to get commandlog history',
+        `Failed to get commandlog history for ${id}`,
+      );
     }
   }
 
@@ -216,7 +231,12 @@ export class McpController {
       if (msg.includes('unknown command') || msg.includes('COMMANDLOG')) {
         return { entries: [], note: 'COMMANDLOG not available on this instance' };
       }
-      throw mapMcpError(this.logger, error, 'Failed to get commandlog patterns');
+      throw mapMcpError(
+        this.logger,
+        error,
+        'Failed to get commandlog patterns',
+        `Failed to get commandlog patterns for ${id}`,
+      );
     }
   }
 
@@ -237,7 +257,12 @@ export class McpController {
         id,
       );
     } catch (error) {
-      throw mapMcpError(this.logger, error, 'Failed to get client activity');
+      throw mapMcpError(
+        this.logger,
+        error,
+        'Failed to get client activity',
+        `Failed to get client activity for ${id}`,
+      );
     }
   }
 
@@ -250,7 +275,12 @@ export class McpController {
       if (msg.includes('CLUSTERDOWN') || msg.includes('cluster mode')) {
         return { error: 'not_cluster', message: 'This instance is not running in cluster mode.' };
       }
-      throw mapMcpError(this.logger, error, 'Failed to get cluster nodes');
+      throw mapMcpError(
+        this.logger,
+        error,
+        'Failed to get cluster nodes',
+        `Failed to get cluster nodes for ${id}`,
+      );
     }
   }
 
@@ -263,7 +293,12 @@ export class McpController {
       if (msg.includes('CLUSTERDOWN') || msg.includes('cluster mode')) {
         return { error: 'not_cluster', message: 'This instance is not running in cluster mode.' };
       }
-      throw mapMcpError(this.logger, error, 'Failed to get cluster node stats');
+      throw mapMcpError(
+        this.logger,
+        error,
+        'Failed to get cluster node stats',
+        `Failed to get cluster node stats for ${id}`,
+      );
     }
   }
 
@@ -280,7 +315,12 @@ export class McpController {
       if (msg.includes('CLUSTERDOWN') || msg.includes('cluster mode')) {
         return { error: 'not_cluster', message: 'This instance is not running in cluster mode.' };
       }
-      throw mapMcpError(this.logger, error, 'Failed to get cluster slowlog');
+      throw mapMcpError(
+        this.logger,
+        error,
+        'Failed to get cluster slowlog',
+        `Failed to get cluster slowlog for ${id}`,
+      );
     }
   }
 
@@ -305,7 +345,12 @@ export class McpController {
       if (msg.includes('CLUSTERDOWN') || msg.includes('cluster mode')) {
         return { error: 'not_cluster', message: 'This instance is not running in cluster mode.' };
       }
-      throw mapMcpError(this.logger, error, 'Failed to get cluster slot stats');
+      throw mapMcpError(
+        this.logger,
+        error,
+        'Failed to get cluster slot stats',
+        `Failed to get cluster slot stats for ${id}`,
+      );
     }
   }
 
@@ -320,7 +365,12 @@ export class McpController {
     try {
       return await this.metricsService.getLatencyHistory(eventName, id);
     } catch (error) {
-      throw mapMcpError(this.logger, error, 'Failed to get latency history');
+      throw mapMcpError(
+        this.logger,
+        error,
+        'Failed to get latency history',
+        `Failed to get latency history for ${id}/${eventName}`,
+      );
     }
   }
 
@@ -343,7 +393,12 @@ export class McpController {
         connectionId: id,
       });
     } catch (error) {
-      throw mapMcpError(this.logger, error, 'Failed to get audit entries');
+      throw mapMcpError(
+        this.logger,
+        error,
+        'Failed to get audit entries',
+        `Failed to get audit entries for ${id}`,
+      );
     }
   }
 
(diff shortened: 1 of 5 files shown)


apps/api/src/mcp/mcp.controller.ts
```diff
@@ -83,11 +82,7 @@ export class McpController {
       const info = await client.getInfoParsed(sections);
       return info;
     } catch (error) {
-      this.logger.error(
-        `Failed to get info for ${id}`,
-        error instanceof Error ? error.stack : error,
-      );
-      throw new HttpException('Failed to get info', HttpStatus.INTERNAL_SERVER_ERROR);
+      throw mapMcpError(this.logger, error, 'Failed to get info');
```
jamby77: **Every log line lost the connection id.**

The old call was `` this.logger.error(`Failed to get info for ${id}`, …) ``; the new fallback is the bare `'Failed to get info'`. Same across all 19 sites — `grep 'for \${id}'` on the branch returns nothing.

On a multi-connection deployment that's the one field that made these logs triageable: 19 endpoints now emit identical text regardless of which instance failed, and the stack won't name it either since the throw comes from `registry.get()`.

`mapMcpError` already takes the message as an argument, so this is just `` mapMcpError(this.logger, error, `Failed to get info for ${id}`) ``. The id is also fine to keep in the client-facing string — the caller passed it in the path.

The status mapping itself is right: `registry.get()` throws `NotFoundException` (`connection-registry.service.ts:329`), `mapMcpError` rethrows `HttpException` untouched, so #155 gets its 404.

jamby77:
#204 fixes the same issue — same two files, also "Fixes #155". This one supersedes it: #204 predates `mapMcpError` (added 2026-07-27 in #330), so it hand-rolls a private helper that would now be a duplicate, and it has gone conflicting since May. Reusing the shared helper is the right call, and it picks up `CapabilityUnavailableError` → 501 for free.

Three details #204 got right that are worth porting before this lands.

**Log context**, as above — `throwMcpError(error, logMessage, fallbackMessage)` takes the log line and the client message separately, so the id survives.

**Client messages stay generic.** `mapMcpError` appends `: ${error.message}` to the response, so raw internal error text now reaches callers on all 19 endpoints where it previously did not. Authenticated-only, behind `AgentTokenGuard`, so hardening rather than a hole — but it is a behaviour change nobody asked for.

**Test drives the real path.** #204 asserts on `registry.get()` throwing, which is what actually produces #155, plus a non-HTTP error still mapping to 500. The test here mocks `metricsService.getHealthSummary`, which proves `mapMcpError` rethrows an `HttpException` but never exercises the lookup that fails in practice — and nothing covers the 500 fallback.

These two are the same root cause: `mapMcpError(logger, error, fallback)` uses one string for both the log and the response, so you cannot have "id in the log, generic to the client". Either give it a separate log message — the cleaner fix, and it matches what the sibling controllers would want too — or pass the id in the single message and accept it in the response, which is harmless since the caller supplied it in the path. The `error.message` suffix is a separate decision either way.

Worth pulling #204's three test cases across regardless; they cover more than the one here and epowell40 wrote them first.

Shikha-code36:
> #204 fixes the same issue — same two files, also "Fixes #155". This one supersedes it: #204 predates `mapMcpError` (added 2026-07-27 in #330), so it hand-rolls a private helper that would now be a duplicate, and it has gone conflicting since May. Reusing the shared helper is the right call, and it picks up `CapabilityUnavailableError` → 501 for free.
> 
> Three details #204 got right that are worth porting before this lands.
> 
> **Log context**, as above — `throwMcpError(error, logMessage, fallbackMessage)` takes the log line and the client message separately, so the id survives.
> 
> **Client messages stay generic.** `mapMcpError` appends `: ${error.message}` to the response, so raw internal error text now reaches callers on all 19 endpoints where it previously did not. Authenticated-only, behind `AgentTokenGuard`, so hardening rather than a hole — but it is a behaviour change nobody asked for.
> 
> **Test drives the real path.** #204 asserts on `registry.get()` throwing, which is what actually produces #155, plus a non-HTTP error still mapping to 500. The test here mocks `metricsService.getHealthSummary`, which proves `mapMcpError` rethrows an `HttpException` but never exercises the lookup that fails in practice — and nothing covers the 500 fallback.
> 
> These two are the same root cause: `mapMcpError(logger, error, fallback)` uses one string for both the log and the response, so you cannot have "id in the log, generic to the client". Either give it a separate log message — the cleaner fix, and it matches what the sibling controllers would want too — or pass the id in the single message and accept it in the response, which is harmless since the caller supplied it in the path. The `error.message` suffix is a separate decision either way.
> 
> Worth pulling #204's three test cases across regardless; they cover more than the one here and epowell40 wrote them first.

Thanks for the thorough review — addressed all three points.

**1. Log context** — Fixed. `mapMcpError` now takes an optional 4th `logMessage` param (defaults to `fallback`, so it's backward-compatible for the 8 existing call sites in `mcp-analytics`/`mcp-ai`/`vector-search`). All 19 sites in `mcp.controller.ts` now pass the id back in (and `getLatencyHistory` also keeps `eventName`, matching the original log line). Confirmed in test output — the log line reads `Failed to get info for conn-1` again.

**2. Client-facing `error.message` suffix** — Left as-is deliberately. It's `mapMcpError`'s existing, already-shipped behavior used by the other 3 controllers; changing it here would silently alter their responses too, which felt out of scope for this fix. You flagged it as a separate/orthogonal decision either way — happy to split it into its own PR if you'd rather it go away everywhere.

**3. Test coverage** — Ported #204's three tests (which exercise the real `registry.get()` → `NotFoundException` path, plus the non-HTTP 500-fallback case) into a new `mcp.controller.spec.ts`, and added one more test asserting the id survives in the log line. Dropped my original test that only mocked `metricsService.getHealthSummary`, since it's now superseded.

**Regression check** — `tsc --noEmit` is clean across the whole `apps/api` project, and all 23 existing tests in the sibling controllers (`mcp-ai`, `mcp-analytics`) still pass unchanged.

Since #204 predates `mapMcpError` and duplicates its logic, this PR should supersede it — let me know if you'd like me to note that there or if you'll close it on your end.

KIvanow:
Hi @Shikha-code36 - thanks for the PR, and for turning around @jamby77's feedback so quickly; the id-in-logs, generic client message, and the reworked test all look good. A couple of small things from me:

1. **PR description is stale** - it still references `mcp.controller.health.spec.ts`, which the fix commits deleted (it's `mcp.controller.spec.ts` now). The Test-plan command points at the old path too.
2. **Flag the shared-helper change** - dropping `: ${error.message}` lives in `mapMcpError`, so it changes the 500 body for the sibling controllers (`mcp-analytics`, `mcp-ai`, `vector-search`) too, not just this one. No tests break, but worth a line in the description / a changelog note since it's a repo-wide response change.

Otherwise LGTM. 👍

Shikha-code36:
> Hi @Shikha-code36 - thanks for the PR, and for turning around @jamby77's feedback so quickly; the id-in-logs, generic client message, and the reworked test all look good. A couple of small things from me:
> 
> 1. **PR description is stale** - it still references `mcp.controller.health.spec.ts`, which the fix commits deleted (it's `mcp.controller.spec.ts` now). The Test-plan command points at the old path too.
> 2. **Flag the shared-helper change** - dropping `: ${error.message}` lives in `mapMcpError`, so it changes the 500 body for the sibling controllers (`mcp-analytics`, `mcp-ai`, `vector-search`) too, not just this one. No tests break, but worth a line in the description / a changelog note since it's a repo-wide response change.
> 
> Otherwise LGTM. 👍

Thanks! Description's updated now, fixed the stale `mcp.controller.health.spec.ts` reference (it's `mcp.controller.spec.ts` now) and the test plan command, and added a note calling out that the `mapMcpError` change affects the 500 body shape for `mcp-analytics`, `mcp-ai`, and `vector-search` too, not just this controller.

apps/api/src/mcp/mcp.controller.ts
```diff
@@ -277,11 +257,12 @@ export class McpController {
         id,
       );
     } catch (error) {
-      this.logger.error(
+      throw mapMcpError(
```
jamby77, line 260: **This is the one path that loses a log line.**

`ClientAnalyticsAnalysisService.getActivityTimeline` throws `new HttpException('Failed to fetch analytics data', 500)` at `client-analytics-analysis.service.ts:293`, so `mapMcpError` returns it untouched and nothing is logged here. Before this PR the catch logged `Failed to get client activity for ${id}` first.

The error isn't lost — the service logs it at `:292` — but that line reads `'Failed to fetch snapshots for activity timeline'` with no connection id, even though `connectionId` is in scope at that point. So on a multi-connection deployment this endpoint's 500s become unattributable, which is the same gap as the one from my last review, one level down.

The client-facing message also changes here: callers now see the downstream `'Failed to fetch analytics data'` rather than `'Failed to get client activity'`.

Cheapest fix is in the service, adding `connectionId` to its own log line. Fine as a follow-up — I'm not holding the PR for it.

jamby77 (approved):
Approving. Both of the earlier points are properly addressed, and the `logMessage` parameter is a better answer than what I suggested — I'd proposed putting the id back into the client-facing string, which would have collided with the "keep the 500 generic" finding. This satisfies both at once.

I verified the log messages rather than eyeballing them: the 18 `logMessage` arguments are byte-identical to the strings on `master`, including the `${id}/${eventName}` one on latency history. No bare `this.logger.error` is left in the controller, and `mapMcpError` now returns the bare `fallback` for non-HTTP errors. #155's goal still holds — `error instanceof HttpException` rethrows untouched, so `registry.get()`'s `NotFoundException` surfaces as a 404.

Two notes, neither blocking.

**Passing `HttpException` through means this layer stops logging it.** For 17 of the 18 endpoints that costs nothing — I checked `MetricsService`, `CommandLogAnalyticsService`, `ClusterDiscoveryService`, `ClusterMetricsService` and the storage layer, and none of them throw `HttpException`. `getClientActivity` is the exception; details inline.

**The blast radius is wider than #155.** Dropping the `: ${error.message}` detail changes the 500 body for every existing `mapMcpError` caller, not just this controller: five sites in `mcp-ai.controller.ts`, three in `mcp-analytics.controller.ts`, one in `vector-search.controller.ts`, and one in the proprietary anomaly controller. Their logs are unaffected, since `logMessage` defaults to `fallback`. It is the right change — just worth stating that it lands outside the issue's scope, in case anything downstream was parsing those messages.

Also noting a pre-existing quirk this PR doesn't introduce and doesn't need to fix: the `msg.includes('COMMANDLOG')` and `msg.includes('CLUSTERDOWN')` guards run before `mapMcpError`, so a missing connection whose id is literally `COMMANDLOG` (valid under `ValidateInstanceIdPipe`) still returns 200 with an empty-entries note instead of the 404 this PR is otherwise delivering.

apps/api/src/mcp/mcp.controller.ts
Shikha-code36: Good catch, thanks, verified: `getActivityTimeline` receives `connectionId` but the catch around `getClientSnapshots` (line 292-293) logged without it, and since that path already throws its own `HttpException`, `mapMcpError` rethrows it untouched before the controller's id-bearing log message ever fires. Rather than a follow-up, fixed it directly in this PR: `client-analytics-analysis.service.ts` now logs `Failed to fetch snapshots for activity timeline for ${connectionId}`.

Also confirmed the client-facing message shift you noted (`'Failed to get client activity'` → `'Failed to fetch analytics data'`) is real but harmless, both are generic, no internal detail leaks either way.

And thanks for flagging the pre-existing `COMMANDLOG`/`CLUSTERDOWN` string-match quirk, agreed it's outside this PR's scope, left as-is.

KIvanow:
Thank you both -  for contributing and for reviewing it :)

PR #409, opened 2026-08-24
feat(connections): SSH tunnel support

KIvanow:
## Summary
Add SSH tunnel support so a connection can reach its database through an SSH bastion/jump host (single hop). Auth is password or private key; keys are supplied inline (encrypted at rest) or by a server-side path gated by the new BETTERDB_SSH_KEY_DIR env var.

## Changes
- New SshTunnelService (ssh2) opens the tunnel and forwards a local port; adapter dials 127.0.0.1 while TLS still validates the real hostname
- Tunnel config threaded through shared types, DTO, registry (secrets encrypted), sqlite/postgres storage (new ssh_tunnel column), and the add-connection UI
- BETTERDB_SSH_KEY_DIR with a path-traversal guard for server-side key files
- Docs and .env.example updated

## Checklist
- [x] Docs added / updated

---

> [!NOTE]
> **High Risk**
> Adds SSH session management, private-key file reads, and encryption of tunnel secrets. Incorrect path checks or host-key handling would be a security issue; cluster nodes still connect outside the tunnel.
> 
> **Overview**
> Connections can now reach Valkey/Redis through a **single-hop SSH bastion** (password or private key), including from the add-connection UI.
> 
> The API opens an `ssh2` local forward and the Valkey client dials `127.0.0.1` while TLS still validates the real hostname. Keys may be pasted inline or read from a path that must resolve inside `BETTERDB_SSH_KEY_DIR` (disabled if unset; traversal and symlink escape are rejected). SSH passwords, passphrases, and inline keys are envelope-encrypted at rest when `ENCRYPTION_KEY` is set; callers cannot set `secretsEncrypted` to skip encryption.
> 
> Host keys can be pinned (`SHA256:…`); if omitted, the first observed fingerprint is stored (TOFU) and verified later. Tunnel drop/reconnect tears down the local listener and clients so a dead loopback port is not reused. Cluster/Sentinel peer nodes are **not** tunnelled (documented limitation).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c70338be65d13aa8dc8fdb99fbaa7e992b67a99f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added SSH tunnel support for database connections through bastion hosts.
  * Configure password or private-key authentication, inline keys, server-side key files, and optional host-key fingerprint verification.
  * Connection testing and saved connections now support SSH tunnel settings.
  * Tunnel credentials are encrypted, and tunnel status indicates host-key pinning.
  * Trust-on-first-use records observed host keys for future verification.

* **Bug Fixes**
  * Improved tunnel cleanup, drop handling, and reconnection to prevent stale connections.

* **Documentation**
  * Added guidance for SSH tunneling, key paths, encryption, host-key verification, and cluster or Sentinel limitations.

### DIFF
--- a/apps/api/src/database/adapters/unified.adapter.ts
+++ b/apps/api/src/database/adapters/unified.adapter.ts
@@ -75,7 +75,20 @@ function isIpAddress(host: string): boolean {
 
 export class UnifiedDatabaseAdapter implements DatabasePort {
   private readonly logger = new Logger(UnifiedDatabaseAdapter.name);
-  private client!: Valkey;
+  // Backing field is genuinely nullable: a tunnelled adapter has no client
+  // until connect() runs, and teardown discards it. Every internal read goes
+  // through the `client` getter below, which throws a clear error instead of
+  // letting an undefined slip through to a TypeError deep in a query method.
+  private _client: Valkey | null = null;
+  private get client(): Valkey {
+    if (!this._client) {
+      throw new Error(
+        'Database connection is not established (no active client). ' +
+          'The server may be unreachable or its SSH tunnel may be down.',
+      );
+    }
+    return this._client;
+  }
   private connected: boolean = false;
   private capabilities: DatabaseCapabilities | null = null;
   private readonly config: UnifiedDatabaseAdapterConfig;
@@ -86,6 +99,10 @@ export class UnifiedDatabaseAdapter implements DatabasePort {
   private readonly tunnelKey: string;
   private readonly usesTunnel: boolean;
   private tunnelActive: boolean = false;
+  // SHA256 fingerprint the SSH server presented, captured on first connect when
+  // no fingerprint was pinned (trust-on-first-use). The registry reads this back
+  // after a successful connect to persist it for subsequent verification.
+  private observedHostKeyFingerprint?: string;
   // Host/port the Valkey clients actually dial. Rewritten to 127.0.0.1:<localPort>
   // once an SSH tunnel is established.
   private connectHost: string;
@@ -115,18 +132,19 @@ export class UnifiedDatabaseAdapter implements DatabasePort {
   }
 
   private initClient(): void {
-    this.client = this.createValkeyClient(this.config.connectionName ?? 'BetterDB-Monitor');
+    const client = this.createValkeyClient(this.config.connectionName ?? 'BetterDB-Monitor');
+    this._client = client;
 
-    this.client.on('connect', () => {
+    client.on('connect', () => {
       this.connected = true;
     });
 
-    this.client.on('error', (err) => {
+    client.on('error', (err) => {
       this.logger.error(`Connection error: ${err.message}`);
       this.connected = false;
     });
 
-    this.client.on('close', () => {
+    client.on('close', () => {
       this.connected = false;
     });
   }
@@ -165,12 +183,40 @@ export class UnifiedDatabaseAdapter implements DatabasePort {
       privateKeyPath: tunnel.privateKeyPath,
       passphrase: tunnel.passphrase,
       hostKeyFingerprint: tunnel.hostKeyFingerprint,
+      // Trust-on-first-use: record the key the server presents so the registry
+      // can persist it and pin it on the next connect.
+      onHostKey: (fingerprint) => {
+        this.observedHostKeyFingerprint = fingerprint;
+      },
+      // If the tunnel drops on its own, stop dialing the dead local port.
+      onUnexpectedClose: () => this.handleTunnelDropped(),
       remoteHost: this.config.host,
       remotePort: this.config.port,
     });
     this.connectHost = '127.0.0.1';
     this.connectPort = localPort;
     this.tunnelActive = true;
+  }
+
+  /**
+   * Called when the SSH tunnel drops unexpectedly. Discards the client (so
+   * iovalkey stops retrying the now-dead loopback port, whose number the OS may
+   * reassign to another process) and resets state so the next connect()
+   * re-establishes the tunnel.
+   */
+  private handleTunnelDropped(): void {
+    if (!this.tunnelActive) {
+      return;
+    }
+    this.logger.warn('SSH tunnel dropped; marking connection down until re-established');
+    this.tunnelActive = false;
+    this.connected = false;
+    if (this._client) {
+      this._client.disconnect();
+      this._client = null;
+    }
+    this.connectHost = this.config.host;
+    this.connectPort = this.config.port;
   }
 
   private async teardownTunnel(): Promise<void> {
@@ -180,13 +226,18 @@ export class UnifiedDatabaseAdapter implements DatabasePort {
       // The current client's options still point at the now-closed local port,
       // so it must be rebuilt (against a freshly established tunnel) before the
       // next connect(). Discard it and restore the pre-tunnel target.
-      if (this.client) {
-        this.client.disconnect();
-        this.client = undefined as unknown as Valkey;
+      if (this._client) {
+        this._client.disconnect();
+        this._client = null;
       }
       this.connectHost = this.config.host;
       this.connectPort = this.config.port;
     }
+  }
+
+  /** The host-key fingerprint observed on connect, if learned via TOFU. */
+  getObservedHostKeyFingerprint(): string | undefined {
+    return this.observedHostKeyFingerprint;
   }
 
   async connect(): Promise<void> {
@@ -194,7 +245,7 @@ export class UnifiedDatabaseAdapter implements DatabasePort {
       if (this.usesTunnel && !this.tunnelActive) {
         await this.establishTunnel();
       }
-      if (!this.client) {
+      if (!this._client) {
         this.initClient();
       }
       this.logger.log(`Connecting to ${this.client.options.host}:${this.client.options.port}...`);
@@ -216,12 +267,12 @@ export class UnifiedDatabaseAdapter implements DatabasePort {
       await this.cliClient.quit().catch(() => {});
       this.cliClient = null;
     }
-    if (this.client) {
+    if (this._client) {
       // iovalkey rejects quit() when the client is already closed / never
       // connected. Swallow it (falling back to a hard disconnect) so a failed
       // quit can never skip the tunnel teardown below and leak the SSH session.
-      await this.client.quit().catch(() => {
-        this.client?.disconnect();
+      await this._client.quit().catch(() => {
+        this._client?.disconnect();
       });
     }
     await this.teardownTunnel();
@@ -229,7 +280,7 @@ export class UnifiedDatabaseAdapter implements DatabasePort {
   }
 
   isConnected(): boolean {
-    return this.connected && this.client?.status === 'ready';
+    return this.connected && this._client?.status === 'ready';
   }
 
   async ping(): Promise<boolean> {
@@ -1046,6 +1097,8 @@ export class UnifiedDatabaseAdapter implements DatabasePort {
   }
 
   getClient(): Valkey {
+    // Throws a clear "connection not established" error (via the getter) rather
+    // than handing back an undefined that blows up as a TypeError in the caller.
     return this.client;
   }
 }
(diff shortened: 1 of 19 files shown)


apps/api/src/database/ssh/ssh-tunnel.service.ts
```diff
@@ -0,0 +1,327 @@
...
+        sockets.add(socket);
+        socket.on('close', () => sockets.delete(socket));
+        sshClient.forwardOut(
+          '127.0.0.1',
+          0,
+          params.remoteHost,
+          params.remotePort,
+          (err, stream) => {
+            if (err) {
+              socket.destroy(
```
jamby77: **An unhandled socket `error` here takes down the whole API process.**

The only `socket.on('error', …)` listener is registered at line 255, inside the *success* branch of the `forwardOut` callback. On this failure path `socket.destroy(err)` emits `'error'` on a socket with no listener, so it becomes an uncaught exception — and `main.ts:121` installs a global `uncaughtException` handler that calls `process.exit(1)`. The monitor dies.

Reachable whenever a channel open fails on a tunnel that is already up: the DB behind the bastion restarts, or a security group changes. The same gap covers a client socket that errors (ECONNRESET) between `net.createServer` firing and the async `forwardOut` callback — for that whole window the socket has no error listener at all.

Fix is to attach the listener as soon as the socket exists, next to `sockets.add(socket)`:

```ts
sockets.add(socket);
socket.on('close', () => sockets.delete(socket));
socket.on('error', () => socket.destroy());
```

and then upgrade it to `() => stream.destroy()` once the stream is available.

apps/api/src/database/ssh/ssh-tunnel.service.ts
```diff
@@ -0,0 +1,327 @@
...
+
+      // Tear the local listener down if the SSH connection drops. Compare
+      // identity so a stale handler from a replaced tunnel never closes the
+      // tunnel that superseded it under the same id.
+      sshClient.on('error', () => {
+        if (this.tunnels.get(connectionId) === info) {
+          this.closeTunnel(connectionId).catch(() => {});
+        }
+      });
+      sshClient.on('close', () => {
```
jamby77: **A dropped SSH connection is never signalled to the adapter, so the connection never recovers on its own.**

This deletes the tunnel and closes the local listener, but nothing tells `UnifiedDatabaseAdapter`: `tunnelActive` stays `true` and `connectHost`/`connectPort` still point at the now-dead loopback port. iovalkey's default retry strategy then reconnects to `127.0.0.1:<freed port>` indefinitely and `establishTunnel()` is never called again — only a manual `POST /connections/:id/reconnect`, which builds a fresh adapter, restores service.

That is a regression against the non-tunnelled path, where a client recovers automatically once the DB comes back.

There is a security edge too: `server.close()` releases the ephemeral port, so the OS is free to reassign it to an unrelated local process — which would then receive the DB `AUTH` credentials from the still-retrying client. Worth a callback/event so the adapter can mark the tunnel dead and re-establish.

apps/api/src/database/adapters/unified.adapter.ts
```diff
@@ -111,8 +131,72 @@ export class UnifiedDatabaseAdapter implements DatabasePort {
...
+  private async teardownTunnel(): Promise<void> {
+    if (this.tunnelActive && this.config.sshTunnelService) {
+      await this.config.sshTunnelService.closeTunnel(this.tunnelKey).catch(() => {});
+      this.tunnelActive = false;
+      // The current client's options still point at the now-closed local port,
+      // so it must be rebuilt (against a freshly established tunnel) before the
+      // next connect(). Discard it and restore the pre-tunnel target.
+      if (this.client) {
+        this.client.disconnect();
+        this.client = undefined as unknown as Valkey;
```
jamby77: **`this.client` can legitimately be `undefined` while roughly 70 call sites dereference it unguarded, and `getClient(): Valkey` can hand back `undefined`.**

Between this line and the constructor skipping `initClient()` for tunnelled adapters, there are two real paths to a registered adapter with no client. Concrete case: a saved SSH-tunnelled connection whose bastion is down at startup — `connection-registry.service.ts:148` still stores the adapter, with no client. A later `getInfo()` / `call()` / `getClient()` throws `TypeError: Cannot read properties of undefined (reading 'info')` rather than a connection error. `cluster-discovery.service.ts:183` does `dbClient.getClient().options.username` and `cluster-metrics.service.ts:299` does `registry.get().getClient().call(...)`, both unguarded.

The `undefined as unknown as Valkey` cast is what hides this from the compiler, and it is the kind of thing the repo's no-`any`/strict rules exist to prevent. Suggest making the field genuinely nullable (`private client: Valkey | null`) and having `getClient()` throw a clear "connection is not established" error, so callers get a real message instead of a TypeError.

apps/api/src/database/adapters/unified.adapter.ts
```diff
@@ -111,8 +131,72 @@ export class UnifiedDatabaseAdapter implements DatabasePort {
...
+      password: tunnel.password,
+      keySource: tunnel.keySource,
+      privateKey: tunnel.privateKey,
+      privateKeyPath: tunnel.privateKeyPath,
+      passphrase: tunnel.passphrase,
+      hostKeyFingerprint: tunnel.hostKeyFingerprint,
+      remoteHost: this.config.host,
+      remotePort: this.config.port,
+    });
+    this.connectHost = '127.0.0.1';
```
jamby77, line 203: **Cluster and Sentinel node fan-out bypasses the tunnel.**

Only the primary client is rewritten to `127.0.0.1`. `cluster-discovery.service.ts` builds per-node `new Valkey({host, port})` straight from the addresses advertised by `CLUSTER NODES`, and the Sentinel topology path does the same. For the headline use case in this PR's own docs — ElastiCache/MemoryDB behind a bastion — those are private-subnet addresses the monitor cannot reach, so every cluster page hangs until each per-node connect times out rather than failing fast.

Not necessarily in scope to fix here, but it should either be documented as a known limitation or the cluster features should be gated when `sshTunnel.enabled` so users get a clear message instead of a slow blank page.

apps/api/src/database/ssh/ssh-tunnel.service.ts
```diff
@@ -0,0 +1,327 @@
...
+          `[${connectionId}] No SSH host-key fingerprint pinned for ${params.sshHost}:${params.sshPort}; ` +
+            'the server identity is not verified (set hostKeyFingerprint to prevent MITM).',
+        );
+      }
+
+      sshClient.connect({
+        host: params.sshHost,
+        port: params.sshPort,
+        username: params.sshUsername,
+        password: params.authMethod === 'password' ? params.password : undefined,
```
jamby77: **Password auth omits `tryKeyboard`, so it will fail on most real bastions.**

ssh2 only attempts the `password` method unless `tryKeyboard: true` is set. Any bastion with `KbdInteractiveAuthentication yes` (the common PAM setup) — and anything with MFA — advertises `keyboard-interactive` instead, so a correct password is rejected and the user gets "SSH authentication failed … check your credentials".

Adding `tryKeyboard: true` plus a `keyboard-interactive` handler that replies with the same password covers the ordinary PAM case.

apps/api/src/database/ssh/ssh-tunnel.service.ts
```diff
@@ -0,0 +1,327 @@
...
+        privateKey: params.authMethod === 'privateKey' ? privateKey : undefined,
+        passphrase: params.authMethod === 'privateKey' ? params.passphrase : undefined,
+        // When a fingerprint is pinned, reject any server whose host key does
+        // not match. Without a pin we accept the key (and warned above).
+        hostVerifier: (key: Buffer) => {
```
jamby77, line 262: **The host key is accepted unconditionally when no fingerprint is pinned, and that is the UI default.**

`if (!pinnedFingerprint) return true;` means the default configuration trusts any key the server presents. Anyone able to MITM the path to the bastion can terminate the SSH session and read the DB username/password that the tunnel carries — the tunnel gives the appearance of protection while providing none.

I can see this is a deliberate, documented tradeoff, and requiring a fingerprint up front would hurt onboarding. But TOFU would get most of the safety for none of the friction: persist the key on first connect and refuse on change. As it stands there is no known_hosts fallback and no way for a user to notice the key changed.

apps/api/src/database/ssh/ssh-tunnel.service.ts
```diff
@@ -0,0 +1,327 @@
...
+    if (!keyDir || keyDir.trim() === '') {
+      throw new Error(
+        `Server-side SSH key files are disabled. Set the ${SSH_KEY_DIR_ENV} environment ` +
+          'variable to a directory containing allowed private keys to enable this option.',
+      );
+    }
+
+    const baseDir = path.resolve(keyDir);
+    const resolved = path.resolve(baseDir, params.privateKeyPath);
+    const rel = path.relative(baseDir, resolved);
```
jamby77: **The `BETTERDB_SSH_KEY_DIR` containment check is purely lexical.**

`path.resolve` + `path.relative` correctly blocks `../` traversal (and the tests cover that), but it does not resolve symlinks — a symlink inside the key dir pointing at `/etc/shadow` or `~/.aws/credentials` passes the check. It also degenerates completely if an operator sets the variable to `/`.

Since this guard is described as what stops the API being coerced into reading arbitrary files, `fs.realpathSync` on `resolved` before comparing would make the implementation match the documentation.

jamby77 (commented):
Reviewed the full diff — the SSH service, adapter changes, registry encryption paths, both storage adapters, the DTO and the UI, plus the surrounding callers.

The storage plumbing is solid: the `ssh_tunnel` column migrations are correct in both SQLite and Postgres, the encrypt/decrypt-at-rest paths hold up, `secretsEncrypted` is properly stripped from input with `forbidNonWhitelisted` on the global pipe, the per-adapter tunnel key correctly avoids collisions across a reconnect, and TLS `servername` is handled right through the tunnel.

Seven comments below. **The first one is a process-killer and should block merge** — an unhandled socket `error` on the forwarding failure path reaches the global `uncaughtException` handler, which calls `process.exit(1)`. The next two are recovery/robustness issues around a dropped tunnel and a possibly-undefined client. The rest are lower severity: cluster fan-out bypassing the tunnel, `tryKeyboard` for PAM bastions, host-key TOFU, and a symlink gap in the key-dir guard.

jamby77 (approved):
Verified all seven findings against `c70338be` rather than taking the summary at face value — all addressed, two of them better than what I suggested.

- **Socket error crash:** fixed twice over. The listener is now attached immediately after `sockets.add`, *and* the forwarding-failure path uses a bare `socket.destroy()` that does not emit `'error'`. Either alone would have closed it.
- **Dropped tunnel:** `handleDrop` on both `'error'` and `'close'`, identity check intact, and the new `onUnexpectedClose` → `handleTunnelDropped()` resets `tunnelActive`, discards the primary and CLI clients and restores the pre-tunnel target. The freed-port credential exposure I flagged is closed with it.
- **Undefined client:** `_client: Valkey | null` behind a private `get client()` that throws a clear "connection is not established" error. Better than the nullable-plus-guard I proposed — every existing call site keeps working and surfaces a real error instead of a TypeError.
- **Cluster fan-out:** documented as a known limitation in README.md and docs/configuration.md. Fine as a documented boundary.
- **tryKeyboard / host key / key dir:** `tryKeyboard` with a keyboard-interactive handler; TOFU via `onHostKey` so the fingerprint gets persisted and pinned; `realpathSync` on both the base dir and the resolved path, so the symlink escape is closed.

The test coverage added alongside is substantial — 446 lines for the tunnel service and 263 for the adapter's SSH paths.

Approving. One thing worth a manual smoke test before release: the tunnel-drop recovery path involves real socket timing that unit tests can only approximate, so it is worth killing a live bastion session once and watching the connection come back.

PR #421, opened 2026-08-26
fix(entitlement): make tenant deprovisioning idempotent/retryable

KIvanow:
## Problem

`deprovisionTenant` is fired async (fire-and-forget from the controller). If the entitlement pod restarts or is evicted mid-run — likely under a big batch of concurrent deprovisions — the pipeline can be interrupted **after** the K8s namespace and Postgres schema are torn down but **before** the final DB-cleanup transaction commits.

The tenant is then stranded in `status='deleting'`, and re-running deprovision threw `P2025` ("record not found") on `prisma.tenant.delete`, so recovery required manual SQL surgery. This happened at scale during recent inactive-tenant cleanups: **11 stuck tenants in one batch, 5 in another**, each needing a hand-written DELETE transaction.

## Fix

Every step of the pipeline is *already* NotFound-safe (schema drop uses `DROP SCHEMA IF EXISTS`, namespace/Route53/scale-down all swallow NotFound) — the only non-idempotent operations were the final row delete and the status writer:

| Before | After |
|--------|-------|
| `prisma.tenant.delete({ where: { id } })` | `prisma.tenant.deleteMany({ where: { id } })` |
| `prisma.tenant.update({ where: { id } })` (in `updateTenantStatus`) | `prisma.tenant.updateMany({ where: { id } })` |

`deleteMany`/`updateMany` no-op instead of throwing when the row is already gone. Result: **re-firing `POST /tenants/:id/deprovision` on a stuck tenant now always drives it to completion** — no direct DB access needed. It also makes concurrent/duplicate deprovision calls for the same tenant safe (the loser no-ops instead of erroring).

## Scope / risk

- Two lines changed in `provisioning.service.ts`, plus comments. No schema/API/behavior change on the happy path.
- `updateTenantStatus` is also used by the provision flow, where the tenant always exists, so `updateMany` behaves identically there.

## Testing

- `tsc --noEmit` on the entitlement package: clean (exit 0).
- Verified manually against production during the batch cleanups: re-firing the endpoint on tenants left in `deleting` (namespace+schema already gone) completes them cleanly with this change's semantics.

## Follow-up (not in this PR)

A periodic reconciler that auto-re-drives tenants stuck in `deleting` past a threshold would remove the need to re-fire manually altogether — intentionally left out to keep this small (would add a scheduler dependency).

---

> [!NOTE]
> **Low Risk**
> Two Prisma call-shape changes in deprovision/status updates only; no schema or API changes, and provision still updates existing tenant rows the same way.
> 
> **Overview**
> Makes **tenant deprovisioning** safe to retry after partial runs (e.g. pod restart after K8s/schema teardown but before DB cleanup), so tenants are no longer stuck in `deleting` with no API recovery path.
> 
> The final cleanup transaction now uses **`tenant.deleteMany`** instead of **`tenant.delete`**, so a second run completes without Prisma **P2025** when the row is already gone. **`updateTenantStatus`** switches to **`updateMany`** for the same reason—especially in the deprovision **catch** block, so a status write after a concurrent delete does not mask the real failure.
> 
> Happy-path behavior is unchanged; earlier pipeline steps were already NotFound-safe.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac10a8555bb1256e244ea0bb2dee0620ceb3363e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tenant deprovisioning reliability when operations are retried.
  * Prevented errors when a tenant has already been removed.
  * Improved handling of concurrent or repeated deprovisioning attempts.

### DIFF
--- a/proprietary/entitlement/src/provisioning/provisioning.service.ts
+++ b/proprietary/entitlement/src/provisioning/provisioning.service.ts
@@ -309,12 +309,18 @@ export class ProvisioningService {
       // with RESTRICT (no cascade), so they must be removed first or the delete
       // throws a foreign-key violation. The tenant's k8s resources (including any
       // Valkey instances) are already gone with the namespace in Step 5.
+      //
+      // Use deleteMany (not delete) for the tenant row so this step is idempotent:
+      // if a previous run was interrupted after tearing down the schema/namespace
+      // but before committing this transaction, re-running deprovisionTenant drives
+      // it to completion instead of throwing P2025 ("record not found") and leaving
+      // the tenant stuck in 'deleting'. Every earlier step is already NotFound-safe.
       this.logger.log(`[${tenant.subdomain}] Deleting tenant record`);
       await this.prisma.$transaction([
         this.prisma.user.deleteMany({ where: { tenantId } }),
         this.prisma.invitation.deleteMany({ where: { tenantId } }),
         this.prisma.valkeyInstance.deleteMany({ where: { tenantId } }),
-        this.prisma.tenant.delete({ where: { id: tenantId } }),
+        this.prisma.tenant.deleteMany({ where: { id: tenantId } }),
       ]);
 
       this.logger.log(`[${tenant.subdomain}] Deprovisioning complete`);
@@ -810,7 +816,11 @@ export class ProvisioningService {
   }
 
   private async updateTenantStatus(id: string, status: TenantStatus, statusMessage?: string): Promise<void> {
-    await this.prisma.tenant.update({
+    // updateMany (not update) so a status write to an already-deleted tenant is a
+    // no-op rather than throwing P2025. This keeps the deprovision error handler
+    // safe when a concurrent/retried run has already removed the row, so it never
+    // masks the original failure with a secondary "record not found".
+    await this.prisma.tenant.updateMany({
       where: { id },
       data: {
         status,


proprietary/entitlement/src/provisioning/provisioning.service.ts
```diff
@@ -309,12 +309,18 @@ export class ProvisioningService {
       // with RESTRICT (no cascade), so they must be removed first or the delete
       // throws a foreign-key violation. The tenant's k8s resources (including any
       // Valkey instances) are already gone with the namespace in Step 5.
+      //
+      // Use deleteMany (not delete) for the tenant row so this step is idempotent:
```
jamby77, line 313: This mechanism cannot occur. `$transaction([...])` at 319-324 is a Prisma batch transaction (single BEGIN..COMMIT, all-or-nothing) and is the last step; a pod restart before commit rolls back and leaves the row in `deleting`. On retry, `findUnique` (261) finds it and the old `tenant.delete` succeeds. P2025 at 323 requires the row to vanish between 261 and 323 within the same run, i.e. a concurrent `deprovisionTenant` or manual SQL. And if the row is already gone at retry time, 262-264 throws NotFound before 323 is reached, so "re-running drives it to completion" is impossible for the missing-row case.

So the `delete` → `deleteMany` swap only changes the concurrent-duplicate case, and whatever actually stranded the 11+5 tenants is still unaddressed. Suggest replacing this block with one accurate line ("a concurrent deprovision may already have removed the row") so the next person investigating a stuck tenant does not look here.

proprietary/entitlement/src/provisioning/provisioning.service.ts
```diff
@@ -309,12 +309,18 @@ export class ProvisioningService {
...
+      // if a previous run was interrupted after tearing down the schema/namespace
+      // but before committing this transaction, re-running deprovisionTenant drives
+      // it to completion instead of throwing P2025 ("record not found") and leaving
+      // the tenant stuck in 'deleting'. Every earlier step is already NotFound-safe.
```
jamby77, line 317: "Every earlier step is already NotFound-safe" is false for the retry-after-restart window this PR describes.

**403 NamespaceTerminating** (Step 3, line 288): if run A restarts inside `deleteNamespace` / `waitForNamespaceDeletion` (up to 60 s, longer with PVC finalizers), the namespace is Terminating when the operator re-fires. Step 2 patch 404 → skipped; Step 3 `createNamespacedJob` → apiserver NamespaceLifecycle admission returns **403** "unable to create new content in namespace … because it is being terminated". `isNotFoundError` (810-816) only matches 404 / `NotFound` / `HTTP-Code: 404`; `runPostgresJob` wraps it as "Failed to create schema-drop job", the Step 3 catch rethrows, and 331 writes `error`. The retry does not complete. Fix: `readNamespace` at the top of `deprovisionTenant` and skip Steps 2-3 when phase is Terminating or 404 (the schema was already dropped before namespace deletion began).

**Concurrent calls are not safe** (Step 1, line 273): the status write is an unconditional `updateMany` with no predicate and the controller has no dedupe, so the loser runs the full destructive pipeline. Its `runPostgresJob` (881-892) deletes the winner's in-flight `schema-drop` Job and recreates it; `deleteRoute53Record` (1671-1698) is list-then-DELETE with no catch, so the loser fails with `InvalidChangeBatch` → 331 writes `error` while the winner is still in `waitForNamespaceDeletion`. `error` passes `provisionTenant`'s status check (162), so an admin can start re-provisioning a tenant whose namespace is mid-delete. Only the final `deleteMany` is idempotent. The claim idiom this needs is already in this file: `const { count } = await prisma.valkeyInstance.updateMany({ where: { id, status: {...} } }); if (count === 0) return;` (381-390) — noting it must still permit re-driving a row already in `deleting`, e.g. via a stale `updatedAt` cutoff or a per-process in-flight `Map<tenantId, Promise>`.

proprietary/entitlement/src/provisioning/provisioning.service.ts
```diff
@@ -810,7 +816,11 @@ export class ProvisioningService {
   }
 
   private async updateTenantStatus(id: string, status: TenantStatus, statusMessage?: string): Promise<void> {
-    await this.prisma.tenant.update({
+    // updateMany (not update) so a status write to an already-deleted tenant is a
+    // no-op rather than throwing P2025. This keeps the deprovision error handler
+    // safe when a concurrent/retried run has already removed the row, so it never
+    // masks the original failure with a secondary "record not found".
+    await this.prisma.tenant.updateMany({
```
jamby77, line 823: This regresses the provision flow. `updateTenantStatus` is shared, and `provisionTenant` relied on the P2025 throw as its only existence check: `deprovisionTenant` has no status precondition, and `provisionTenant` never inspects `count` or status.

Admin fires `POST /deprovision` while provision is between `findUnique` (157) and Step 1 (180), or during the multi-minute waits in Steps 10-11; deprovision's transaction deletes the row. Before: `tenant.update` threw P2025 at 180 (nothing created) or at 242 (before the email). After: Step 1 / Step 12 `updateMany` match 0 rows and return normally, so provisioning creates namespace, schema, deployment, ingress and Route53 CNAME — or, at 242, logs "Provisioning complete!" and `sendWelcomeEmail` (246-250) emails the customer a URL for a deleted tenant. The orphan can never be torn down via the API because `deprovisionTenant` 261-264 throws NotFound for the missing row.

The PR premise "in the provision flow the tenant always exists" is exactly the case that changed. Narrower fix: keep the helper as `update` and tolerate P2025 only in the deprovision catch at 331, or return `count` and have callers bail on 0 as `provisionValkeyInstance` does at 381-390 / 422-430.

jamby77 (commented):
Reviewed at `ac10a855`. The two-line swap is safe on the happy path and does stop the secondary P2025 in the deprovision catch, but the stated root cause is not reachable from the code, and making the shared `updateTenantStatus` helper silent regresses the provision flow. Both CodeRabbit Major comments line up with findings here. Inline comments on the diff lines; two more that fall outside the hunks:

**`runPostgresJob` poll treats 404 as transient** (line 966) — a Job garbage-collected when the winner deletes the namespace returns 404 on every poll; the catch only rethrows on `job failed`, so the run burns the full 300 s timeout and then throws a plain `Error` the Step 3 `isNotFoundError` skip cannot recognise → status `error`. Rethrow when `isNotFoundError(error)` in that catch so the existing skip handles it.

**`deprovisionValkeyInstance` has the identical hole** (line 515 + `updateValkeyInstanceStatus` at 608) — `valkeyInstance.delete` / `.update` throw P2025 in the same shape this PR fixes for tenants, and the catch's status write masks the original error exactly as the new comment at 819 describes. Same two-line fix, or better, one shared status-claim helper for both paths.

Also worth a follow-up: `createInstance` / `provisionValkeyInstance` do not reject a tenant in `deleting`, so an in-flight Valkey provision can recreate the namespace after Step 5 and the tenant transaction then deletes the rows, leaving an orphan namespace no API path can reach. Pre-existing, but the idempotency story here reinforces the false assumption.

PR #430, opened 2026-09-02
feature: configurable local data retention for self-hosted deployments

KIvanow:
## Summary
Self-hosted deployments kept most stored monitoring history forever — but a few high-volume stores self-pruned at a hardcoded 7 days. Now self-hosted deletes **nothing** by default; an opt-in retention window (`LOCAL_RETENTION_DAYS` env var or Settings → Data Retention) enables a daily sweep plus hourly trimming of the high-volume stores. Cloud keeps the tier-based policy (7/90/365).

## Changes
- Hoist tier retention windows (7/90/365) behind one canonical `TIER_RETENTION_DAYS` constant in `@betterdb/shared` (cloud-only)
- `RetentionPolicyService`: cloud → tier window; self-hosted → the configured window or null (no pruning); the four features that hardcoded a 7-day prune now use it
- `LocalRetentionService`: daily self-hosted sweep (no-op in cloud mode / when unset); shares one `runRetentionSweep()` prune list with the cloud sweep
- New `localRetentionDays` setting (shared type, sqlite column + lazy ALTER, memory adapter, UI panel)

## Checklist
- [x] Docs added / updated

## Summary by CodeRabbit

- **New Features**
  - Added configurable data retention for self-hosted deployments through **Settings → Data Retention** or `LOCAL_RETENTION_DAYS`.
  - Users can choose a retention period or leave it blank to keep data indefinitely.
  - Cloud retention follows tier limits: Community (7 days), Pro (90 days), and Enterprise (365 days).
  - High-volume cloud samples are retained for up to seven days.
  - Retention cleanup runs automatically on a recurring schedule.
- **Bug Fixes**
  - Invalid or out-of-range retention values are rejected.
- **Documentation**
  - Clarified that the environment variable seeds settings only on fresh installations.

---

> [!NOTE]
> **High Risk**
> Changes when and how stored monitoring data is deleted across many tables and pollers; misconfiguration or cache races could prune history unexpectedly, though the PR adds explicit guards (loaded settings only, deferred first prune, validation).
> 
> **Overview**
> Self-hosted installs **no longer hard-delete** monitoring history at a fixed 7 days. History stays **until the operator opts in**: `LOCAL_RETENTION_DAYS` seeds `localRetentionDays` on first settings creation, then **Settings → Data Retention** owns the window; **null means keep forever**. A new **`RetentionModule`** (`RetentionPolicyService` + `LocalRetentionService`) runs a **startup-delayed daily sweep** over all monitoring stores when a window is set; cloud still uses **tier windows** via `TIER_RETENTION_DAYS` (7/90/365).
> 
> **Pollers** (AI observability, command/latency stats, vector index) now read **`getSampleRetentionMs()`** instead of a baked-in 7 days, skip pruning when unset, and **defer the first hourly trim** so enabling retention over a large backlog does not mass-delete on boot. AI observability also prunes **OTel spans** on the same local cycle.
> 
> **Safety and ops**: shared **`isCloudMode()`** aligns boot validation, OTLP ingest, and proprietary loading with any truthy `CLOUD_MODE` (not only `"true"`). Settings cache uses **`getLoadedSettings()`** so env fallbacks cannot trigger deletion; **chunked SQLite/Postgres deletes** and **trace-atomic OTel pruning** reduce lock/orphan risk. Storage gains **`local_retention_days`**, retention indexes, and webhook prune predicates that can use indexes.
> 
> **Web**: migration analysis uses runtime **`isCloudMode`** from `cloudUser`, not `VITE_CLOUD_MODE`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3d680700909dd36639eee50eb089bf0119c0e12. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jamby77 (commented):
Review of the retention change. Verified each finding against `232337ed`.

The self-hosted half of this is well built — `RetentionPolicyService` is the right seam, `runRetentionSweep` is a good way to stop the two sweeps drifting, and the `parseLocalRetentionDays` strictness is a nice touch. Seven findings below, of which two independently defeat the PR's headline promise (Postgres never stores the setting; `key-analytics` still deletes self-hosted data on its own timer) and one regresses cloud in the opposite direction (the three sample stores lose their 7-day cap).

apps/api/src/storage/adapters/sqlite.adapter.ts
```diff
@@ -1385,6 +1386,7 @@ export class SqliteAdapter implements StoragePort {
         throughput_forecasting_default_rolling_window_ms INTEGER NOT NULL DEFAULT 21600000,
         throughput_forecasting_default_alert_threshold_ms INTEGER NOT NULL DEFAULT 7200000,
         inference_sla_config TEXT NOT NULL DEFAULT '{}',
+        local_retention_days INTEGER,
```
jamby77, line 1408: **The new setting does not exist on PostgreSQL.**

SQLite gained the column here, the lazy `ALTER` above it, and the read/write mapping; Memory was updated too. `postgres.adapter.ts` has zero occurrences of `local_retention_days`: no column in the `app_settings` DDL, no `ALTER TABLE app_settings ADD COLUMN IF NOT EXISTS local_retention_days INTEGER;` alongside the existing migrations, and `saveSettings` still lists 11 columns and 11 params.

`updateSettings` merges and then calls `saveSettings`, so the value is dropped on write, and `getSettings` maps `row.local_retention_days ?? null` to `null` on read. A self-hosted operator on `STORAGE_TYPE=postgres` sets 30 days in Settings -> Data Retention, gets a 200 whose body already reads back `null`, watches the field blank itself, and is never pruned. `LOCAL_RETENTION_DAYS` is discarded at first init the same way.

Postgres is the documented production backend, so this is the path most self-hosted installs are on.

packages/shared/src/license/types.ts
```diff
@@ -74,6 +74,19 @@ export const TIER_FEATURES: Record<Tier, Feature[]> = {
...
 };
 
+/**
+ * Canonical per-tier retention window for stored monitoring history, in days.
+ * Applies to cloud deployments only (the nightly sweep and feature-level
+ * pruning derive their cutoffs from it). Self-hosted retention is governed
+ * solely by the `localRetentionDays` app setting (or `LOCAL_RETENTION_DAYS`
+ * env var) — unset means keep forever.
+ */
+export const TIER_RETENTION_DAYS: Record<Tier, number> = {
```
jamby77, line 84: **A second `TIER_RETENTION_DAYS` survives in `proprietary/key-analytics`, and it still deletes self-hosted data.**

`proprietary/key-analytics/key-analytics.service.ts:49` declares its own module-local `TIER_RETENTION_DAYS` (community 7 / pro 30 / enterprise `null`) and prunes on a private 24h timer in `pruneOldData`. This PR touches only `data-retention.service.ts` and `licenses/types.ts` under `proprietary/`, so that path is untouched and never consults `RetentionPolicyService`.

Two consequences. A self-hosted Pro operator who upgrades expecting "nothing is deleted while the window is unset" still loses key-pattern snapshots and hot-key history at 30 days; if they set `localRetentionDays = 365`, key analytics is still cut at 30. And the "one canonical constant" goal isn't reached — two constants now share the name with different values, and since `proprietary/licenses/types.ts` re-exports this one, adding it to that file's imports would be a redeclaration error rather than a drop-in.

Note `key_patterns` *is* in `runRetentionSweep`, so those rows are pruned from two places on two different windows.

apps/api/src/settings/settings.service.ts
```diff
@@ -74,11 +81,23 @@ export class SettingsService implements OnModuleInit, OnModuleDestroy {
         10,
       ),
       inferenceSlaConfig: {},
+      localRetentionDays: this.parseLocalRetentionDays(
```
jamby77: **`LOCAL_RETENTION_DAYS` only takes effect on a brand-new database.**

`buildSettingsFromEnv()` is persisted only by `initializeFromEnv()`, and `onModuleInit` calls that solely when `getSettings()` returns nothing. Every existing deployment already has that row, and `refreshCache()` prefers it (`dbSettings || buildSettingsFromEnv()`).

So an operator who upgrades, sets `LOCAL_RETENTION_DAYS=30` and restarts keeps `local_retention_days = NULL` in the DB and is never pruned. Both `docs/configuration.md` ("either via `LOCAL_RETENTION_DAYS` or ... the settings page value wins once saved") and the UI hint ("Can also be seeded with the `LOCAL_RETENTION_DAYS` environment variable") promise the env var works on an existing install.

Separately, `LOCAL_RETENTION_DAYS` is missing from `.env.example` and from `env.schema.ts`, so it is neither discoverable nor validated.

apps/api/src/metrics/commandstats-poller.service.ts
```diff
@@ -164,10 +165,11 @@ export class CommandstatsPollerService extends MultiConnectionPoller implements
 
     await this.storage.saveCommandStatsSamples(batch, ctx.connectionId);
 
+    const retentionMs = this.retentionPolicy.getRetentionMs();
```
jamby77: **Cloud loses the only bound on its three highest-volume tables.**

This poller (and `latencystats-poller.service.ts:131`, `vector-search.service.ts:116`) previously capped its store at a hard 7 days *in every deployment, cloud included*. `getRetentionMs()` now returns the tier window when `CLOUD_MODE=true`, and these three stores are deliberately absent from `runRetentionSweep`'s prune list — so this hourly trim is the only thing that ever deletes from them.

A cloud Enterprise tenant's `command_stats_samples` therefore goes from a 7-day cap to 365: one row per command per connection per 60s poll, on the highest-cardinality table in the schema, is roughly a 52x growth. `latency_stats_samples` follows at 52x and `vector_index_snapshots` at 13x for Pro.

The self-hosted half of the change is right; it's the cloud branch that regresses. Either add these three stores to `runRetentionSweep` and let the tier window govern them there, or keep a separate tighter cap for sample stores and use `getLocalRetentionDays()` here instead of `getRetentionDays()`.

apps/api/src/retention/retention-policy.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    @Optional() private readonly licenseService?: LicenseService,
+  ) {}
+
+  /**
+   * The operator-configured self-hosted retention window in days, or null
+   * when unset (or when running in cloud mode).
+   */
+  getLocalRetentionDays(): number | null {
+    if (process.env.CLOUD_MODE === 'true') return null;
+    const days = this.settingsService.getCachedSettings().localRetentionDays;
```
jamby77: **The env fallback can resurrect a window the operator explicitly cleared.**

`getCachedSettings()` returns `buildSettingsFromEnv()` whenever `cachedSettings` is still `null` — which is precisely the window between process start and the first `refreshCache()` await. `MultiConnectionPoller.start()` fires a fire-and-forget initial poll from `onModuleInit`, and `MetricsModule` / `VectorSearchModule` / `AiObservabilityModule` all precede `SettingsModule` in `baseImports`, so that first tick can read the fallback while `SettingsService.onModuleInit` is still awaiting storage.

Scenario: `LOCAL_RETENTION_DAYS=7` is left in the env from an earlier experiment, the operator later clears the window in the UI (DB row -> `null`, keep forever). On the next restart the first poll sees 7 days and prunes command/latency-stats rows the operator opted out of keeping — a deletion that is not recoverable.

Returning `null` here when the settings cache has not loaded yet keeps the default safe: the tick simply skips pruning and the next one, 60s later, has real settings.

apps/api/src/retention/retention-sweep.ts
```diff
@@ -0,0 +1,52 @@
...
+ * self-hosted local sweep so the two never drift on which tables are covered.
+ *
+ * A store that fails to prune is reported as -1 and does not stop the sweep.
+ */
+export async function runRetentionSweep(
+  storage: StoragePort,
+  cutoff: number,
+  logger: Logger,
+): Promise<Record<string, number>> {
+  const pruneOps: Array<{ name: string; fn: () => Promise<number> }> = [
```
jamby77, line 22: **The daily sweep can't reclaim orphaned rows from the high-volume stores.**

`command_stats_samples`, `latency_stats_samples` and `vector_index_snapshots` are absent from this list, so they are trimmed only inside each poller's `pollConnection` — which requires the connection to be registered *and* reachable.

Remove a connection, or leave one down, and its rows in those three tables are kept forever even with a retention window configured. The Settings panel tells the operator the opposite: "High-volume stat samples (command/latency stats, vector index snapshots, AI samples) are additionally trimmed to this window on an hourly cycle."

Adding them here would fix this and the cloud-unbounded-growth problem in one move, since the sweep prunes by cutoff without needing a live connection.

apps/web/src/pages/Settings.tsx
```diff
@@ -562,6 +563,45 @@ export function Settings({ isCloudMode = false }: { isCloudMode?: boolean }) {
...
+                    min={1}
+                    value={formData.localRetentionDays ?? ''}
+                    placeholder="Keep forever"
+                    onChange={(e) => {
+                      // valueAsNumber understands everything a number input
+                      // accepts (e.g. "1e2" is 100, not parseInt's 1).
+                      const parsed = e.currentTarget.valueAsNumber;
+                      handleInputChange(
+                        'localRetentionDays',
+                        Number.isFinite(parsed) && parsed >= 1 ? Math.floor(parsed) : null,
```
jamby77: **A sub-1 entry silently means "keep forever" instead of being rejected.**

`0`, negatives and `0.5` all map to `null`, and because the input is controlled by `value={formData.localRetentionDays ?? ''}` the field blanks itself the moment it happens. A user who types `0` meaning "no limit, tell me if that's wrong", or who steps the spinner down from 1, sees the field empty and on save gets retention disabled rather than an error.

The server-side `BadRequestException` for `0` in `updateSettings` is unreachable from this input for the same reason — the client converts the invalid value to a valid one before it is ever sent.

KIvanow:
@jamby77 all seven findings addressed in 45b8126c — thanks, the Postgres one especially was a real hole:

1. **Postgres**: `local_retention_days` added to DDL, the `ADD COLUMN IF NOT EXISTS` block, and `saveSettings`. Verified live against a real Postgres, both fresh-install and dropped-column upgrade paths.
2. **key-analytics**: private tier map removed; it now consults `RetentionPolicyService` (tier window in cloud, operator window or nothing self-hosted). Cloud moves to 7/90/365 — pro 30→90, enterprise hot-keys unlimited→365 (key patterns were already swept at 365).
3. **Env var**: kept the seed-on-first-init model (re-applying on boot would conflict with your finding 5's "cleared stays cleared"), and made docs/UI say so; added to `.env.example` + `env.schema.ts`. Open to revisiting the precedence if you'd rather the env var win.
4. **Cloud sample stores**: new `getSampleRetentionMs()` keeps the longstanding hard 7-day cloud cap regardless of tier; self-hosted follows the operator window.
5. **Env-fallback race**: retention reads `getLoadedSettings()` (null until the cache loads) — never the env-derived fallback.
6. **Orphaned rows**: the three sample stores are now in `runRetentionSweep`, so both sweeps reclaim rows without needing a live connection.
7. **Input UX**: invalid entries stay visible with an inline error and Save is disabled, instead of silently becoming keep-forever.

jamby77:
## Review findings (against `bed2b048`)

### Blocking

**1. Upgrading self-hosted installs silently lose the 7-day prune** — `apps/api/src/retention/retention-policy.service.ts:70`
The lazy ALTER adds `local_retention_days` as NULL to the existing `app_settings` row, and `SettingsService.onModuleInit` only calls `initializeFromEnv()` when `getSettings()` returns null. So `command_stats_samples` / `latency_stats_samples` / `vector_index_snapshots` / `ai_cache_samples` (previously trimmed hourly to 7 days by the removed `RETENTION_MS` constants) now grow unbounded after upgrade. Setting `LOCAL_RETENTION_DAYS=7` and restarting has no effect and only emits a debug log. Needs a one-time backfill of an existing NULL row from env, or at minimum a startup warning plus an upgrade note.

**2. Number input commits `null` on a stray keystroke** — `apps/web/src/pages/Settings.tsx:606`
For `type=number` inputs the browser sanitises invalid text to `''`. With saved value 30, typing `e` (or `-`, `+`, `.`) makes `e.target.value === ''`, the `raw.trim() === ''` branch fires, `handleInputChange('localRetentionDays', null)` runs and Save is enabled, while the DOM still shows `30e`. Clicking Save silently disables retention. Check `e.currentTarget.validity.badInput` and keep the error state instead of committing `null`.

**3. `CLOUD_MODE` is parsed two different ways** — `apps/api/src/retention/retention-policy.service.ts:34`
`app.module.ts` tests truthiness (`if (process.env.CLOUD_MODE)`, lines 127/147); `RetentionPolicyService` and `LocalRetentionService` test the literal `'true'`. With `CLOUD_MODE=1` the cloud cron loads but the pollers take the self-hosted branch, `getRetentionDays()` reads a NULL `localRetentionDays` for every tenant, `hot_keys` / `key_pattern_snapshots` are never pruned, and `LocalRetentionService` schedules its own sweep alongside the tier cron. Normalise on one `isCloudMode()` helper.

### Data-loss behaviour changes not called out

**4. Reset to Defaults re-arms deletion** — `apps/api/src/settings/settings.service.ts:181`
`resetToDefaults()` re-seeds `localRetentionDays` via `initializeFromEnv()`, contradicting the docs' promise that clearing the window in the UI sticks even if the env var stays set. An operator who cleared the window and then hits Reset to fix an unrelated poll interval gets everything older than N days deleted within the hour. Either exclude the field from reset or make the confirm dialog say the window will be re-armed from env.

**5. Enterprise `hot_keys` retention silently changes from unlimited to 365 days** — `proprietary/key-analytics/key-analytics.service.ts:95`
The per-service map `{community: 7, pro: 30, enterprise: null}` was replaced with the shared `TIER_RETENTION_DAYS`. Enterprise goes from never-pruned to 365 days, Pro from 30 to 90. First 24h tick after deploy deletes >1-year enterprise history irreversibly. `docs/configuration.md:331` still says "Enterprise keeps data indefinitely" (contradicting line 239 in the same file). No spec covers cloud mode here (all key-analytics specs stub `getRetentionMs -> null`). If intentional, needs a release note and doc fix; if not, preserve the enterprise-unlimited case.

**6. Sweep deletes the ACL audit trail without disclosure** — `apps/api/src/retention/retention-sweep.ts:23`
The self-hosted sweep runs `pruneOldEntries` (`DELETE FROM acl_audit`), plus key pattern snapshots, correlated groups and OTel spans. Neither the Settings copy (`Settings.tsx:587-591` lists slow log, command log, snapshots, anomaly events, webhook deliveries, captures, AI samples) nor `docs/configuration.md` mention audit entries. `docs/vector-ai/README.md:94-95` also still claims the sample tables are "retained 7 days". Name every swept store in the UI copy and docs.

### Correctness gaps

**7. `hot_keys` is missing from the shared sweep** — `apps/api/src/retention/retention-sweep.ts:37`
`runRetentionSweep` is documented as covering every store, but `hot_keys` is only trimmed by `KeyAnalyticsService`'s own 24h `setInterval`, which never fires if the process restarts within 24h (no initial run) and does not start when the feature is unlicensed. Add `{ name: 'hot_keys', fn: () => storage.pruneOldHotKeys(cutoff) }` and drop the store-specific pruner.

**8. `LOCAL_RETENTION_DAYS` validated inconsistently** — `apps/api/src/settings/settings.service.ts:116`
`env.schema.ts:43` uses `z.coerce.number().int().min(1)`; `parseLocalRetentionDays` uses a strict `/^\d+$/` regex; and `ConfigService` serves raw `process.env` (no `validate` in `config.module.ts`). `LOCAL_RETENTION_DAYS=1e2` or `30.0` passes boot validation but seeds `null`, so the operator believes a window is active and nothing is pruned. Use one validator.

**9. `refreshCache()` falls back to env-derived settings** — `apps/api/src/settings/settings.service.ts:58`
When `getSettings()` returns null at runtime (restored SQLite file, truncated table), `refreshCache` stores `buildSettingsFromEnv()` in `cachedSettings`. `getLoadedSettings()` then hands `RetentionPolicyService` an env-derived `localRetentionDays` that was never persisted, breaking the "only persisted settings count" guarantee the docblock relies on. Keep the previous cache instead: `if (dbSettings) { this.cachedSettings = dbSettings; }`.

### Performance

**10. Sweep is a blocking full-table scan on SQLite** — `apps/api/src/retention/retention-sweep.ts:43`
Every `pruneOld*` is one synchronous better-sqlite3 DELETE (`sqlite.adapter.ts:3376, 3470, 3597, 3733` etc.), and `await op.fn()` only yields a microtask, so the first sweep on a year-old keep-forever DB freezes HTTP, websockets and pollers for the whole pass. The new global prunes on `command_stats_samples` / `latency_stats_samples` / `ai_cache_samples` also have no usable index — every index leads with `connection_id` — so they seq-scan daily. Prune in bounded chunks (`WHERE rowid IN (SELECT rowid ... LIMIT 10000)`) with a `setImmediate` yield between chunks, and add standalone timestamp indexes or iterate `SELECT DISTINCT connection_id` and use the per-connection variant. Also `pruneOldOtelSpans`' `GROUP BY trace_id HAVING MIN(start_time_ms) < ?` forces a full aggregate scan; `WHERE trace_id IN (SELECT DISTINCT trace_id FROM otel_spans WHERE start_time_ms < ?)` is equivalent and index-driven.

### Minor

- `DataRetentionService` still computes the cloud cutoff itself instead of calling the new `RetentionPolicyService`.
- `MAX_RETENTION_DAYS` and the integer-range validation are duplicated across `env.schema.ts`, `settings.service.ts` (twice), `retention-policy.service.ts` and `Settings.tsx`.

KIvanow:
@jamby77 round two addressed in ed4d97d7:

- **1 (upgrade loses 7-day trim)**: intended behavior per the new keep-forever-by-default model — will be called out in the release notes rather than auto-backfilled.
- **2**: `validity.badInput` now keeps the error state; a stray `e` can no longer commit keep-forever.
- **3**: single `isCloudMode()` helper across app.module, both retention services, ai-observability and the cloud cron.
- **4**: `resetToDefaults()` preserves the retention window.
- **5**: kept as intended (unified policy); the contradicting docs line is fixed and it joins the release notes.
- **6**: UI + docs now name every swept store, ACL audit entries in bold; the stale 7-day claims in vector-ai docs corrected.
- **7**: `hot_keys` added to the shared sweep; the private key-analytics pruner is gone.
- **8**: one strict `parseRetentionDaysToken()` + `MAX_RETENTION_DAYS` in `@betterdb/shared`, used by env schema, API and UI — `1e2` now fails at boot.
- **9**: `refreshCache()` keeps the previous cache when the DB read comes back empty.
- **10**: all sqlite prune paths (adapter, slowlog/webhook repos, captures) now delete in 10k-row chunks with event-loop yields (chunk-boundary tested at 25k rows); global timestamp indexes added on the sample stores + hot_keys in both sqlite and postgres; OTel prune rewritten to the index-driven `DISTINCT` form in both adapters.
- **Minors**: `DataRetentionService` reads its window from `RetentionPolicyService`; the max-days constant is defined once.

Not done: normalising the remaining `CLOUD_MODE === 'true'` sites outside retention (auth guards, telemetry) — happy to do that as a follow-up if you want it in this PR.

KIvanow:
@jamby77 **Re: finding 1 — upgrading self-hosted installs silently lose the 7-day prune**

This one is working as intended. The point of the change is that self-hosted deletes nothing unless the operator opts in — the old 7-day trim on the sample stores was exactly the kind of silent, hardcoded deletion we're removing, so backfilling a window from env on upgrade would reintroduce it through the back door (and conflict with your finding 5's "cleared stays cleared" semantics: we'd have no way to distinguish a deliberate NULL from a migrated one).

The consequence you're describing — sample stores growing unbounded after upgrade until a window is set — is real, and we'll call it out prominently in the v0.42 release notes as an action item for upgrading operators: "set Settings → Data Retention (or `LOCAL_RETENTION_DAYS` on fresh installs) if you want the previous trimming behavior back." The notes will also cover the cloud key-analytics window change from your finding 5.

jamby77:
## Review findings (against `2046e1ed`)

Closed since the last round: blank `LOCAL_RETENTION_DAYS=` boot failure, capture-chunk prune, reset-to-defaults carve-out, and the extra validation rule in `RetentionPolicyService`. Thanks.

### Still open

**1. `isCloudMode()` is lax while the rest of the codebase is strict** — `apps/api/src/common/utils/cloud-mode.ts:9`
The helper accepts any value except `'false'`/`'0'`, but `LicenseService`, cloud-auth guard/middleware, agent-token guard, otel-ingest and `main.ts` still test `=== 'true'`. With `CLOUD_MODE=1`, `DataRetentionModule` loads, `LocalRetentionService` disables itself (the operator's Settings window is ignored), and the 03:00 cron asks `getLicenseTier()`, which resolved entitlement through the self-hosted branch and returns community → every store older than 7 days is deleted, while cloud-auth still passes requests through unauthenticated. Migrate the remaining raw checks to the helper, or reject non-`'true'` values in `env.schema`.

**2. Upgraded installs silently ignore `LOCAL_RETENTION_DAYS`** — `apps/api/src/settings/settings.service.ts:36`
The env var only reaches the DB via `initializeFromEnv()` when no `app_settings` row exists. On upgrade the lazy ALTER adds `local_retention_days = NULL` to the existing row; the operator sets `LOCAL_RETENTION_DAYS=30` and restarts; `env.schema` accepts it, `onModuleInit` skips seeding, and the sample stores that master pruned unconditionally at 7 days now grow unbounded with only a debug-level log. Add a `warn` at boot when the env var differs from the stored value, or seed the column once when it is first added and NULL.

**3. `MemoryAdapter.pruneOldKeyPatternSnapshots` still throws** — `apps/api/src/storage/adapters/memory.adapter.ts:773`
Only `pruneOldHotKeys` was changed to return 0. Every sweep on `STORAGE_TYPE=memory` logs `Failed to prune key_patterns: Key analytics not supported in memory adapter` and reports `key_patterns=-1`.

**4. Enterprise cloud `hot_key_stats` goes from keep-forever to 365 days** — `apps/api/src/retention/retention-sweep.ts:29`
Master's cloud sweep did not include `pruneOldHotKeys`, and the deleted `KeyAnalyticsService` pruner returned early for enterprise. The first cron after deploy permanently removes rows older than a year; Pro moves from 30 to 90 days. The PR body calls this a hoist. Either call out the one-time deletion in the description/release note or exempt enterprise.

**5. Retention error silently disables Save on every tab** — `apps/web/src/pages/Settings.tsx:784`
`retentionError` gates the shared Save button but only renders inside the Data Retention tab and is not cleared on tab switch. User types `0`, switches to Audit Trail, edits the poll interval → Save is greyed out with no message; Discard clears the error but also throws away the audit edit. Render the error near the button or clear/validate it on tab change.

### New in this round

**6. Cloud sample cap does not cover AI samples or OTel spans** — `apps/api/src/retention/retention-sweep.ts:39`
The new `CLOUD_SAMPLE_RETENTION_DAYS` comment, the Settings UI copy and docs list AI samples and OTel spans among the 7-day-capped stores, but both ops still receive the tier `cutoff` rather than `sampleCutoff`. In cloud the AiObservability poller skips local pruning, so nothing applies the claimed cap. Pass `sampleCutoff` for those two ops or fix the text.

**7. Postgres prunes remain single unbounded DELETEs** — `apps/api/src/storage/adapters/postgres.adapter.ts:3633`
SQLite got `chunkedSqliteDelete`, but all 21 Postgres prune methods are still one `DELETE ... WHERE ts < $1`. The first self-hosted Postgres sweep over years of keep-forever history holds row locks and a pool connection for minutes per table, writes GBs of WAL in one transaction, and blocks the hourly poller prunes. Batch with `DELETE ... WHERE ctid = ANY(ARRAY(SELECT ctid FROM t WHERE ts < $1 LIMIT n))` in a loop. The webhook predicate `EXTRACT(EPOCH FROM created_at) * 1000 < $1` is also non-sargable.

**8. Sweep timer has no in-flight guard** — `apps/api/src/retention/local-retention.service.ts:41`
`runSweep` is driven by `setInterval` with no overlap protection. A first-time sweep that outlives the 24h interval (see 7) gets a second identical sweep started on top of it. `MultiConnectionPoller` and `BasePollingService` both re-arm a `setTimeout` after completion; do the same here.

KIvanow:
@jamby77 round three in 2bc5d9b1 — fixed 1, 3, 5, 6, 7, 8:

- **1**: every remaining raw `CLOUD_MODE` check (main.ts, agent-token guard/providers, otel-ingest, telemetry, cloud-auth guard/middleware, license service, memory-proposals gate) now uses `isCloudMode()`. Only exception: the entitlement service — separate deployment, its check concerns its own env.
- **3**: `pruneOldKeyPatternSnapshots` returns 0 on the memory adapter, same as hot keys.
- **5**: leaving the Data Retention tab clears a pending invalid entry, so Save is never gated by an invisible error.
- **6**: resolved as a text fix — cloud AI samples/OTel spans were tier-window on master too (the sweep always owned them there), so capping them at 7d now would delete tenant history. The `CLOUD_SAMPLE_RETENTION_DAYS` comment now states the split explicitly; the UI/docs copy was self-hosted-scoped and stays.
- **7**: `chunkedPostgresDelete()` (ctid batches of 10k, one transaction each) across all postgres prune paths incl. the slowlog/webhook repos; per-trace batching for OTel; the webhook predicate is now sargable (`created_at < to_timestamp($1/1000.0)`). Verified live across batch boundaries (25k rows, 3 batches).
- **8**: the sweep re-arms a `setTimeout` after completion (MultiConnectionPoller pattern) — no overlap possible.

**2** and **4** stay as intended behavior per the earlier thread — both are going in the v0.42 release notes (upgrade note for the sample stores; cloud key-analytics window change).

KIvanow:
@jamby77 All three rounds are addressed as of `7f83a8d5` — everything is fixed except findings 2 and 4 from your last batch, which stay as intended behavior and are earmarked for the v0.42 release notes. One heads-up: the branch was force-pushed to re-sign two commits, so SHAs moved (`2bc5d9b1` → `7f83a8d5`) but the diff is byte-identical. Ready for your final pass whenever you are.

apps/web/src/pages/Settings.tsx
```diff
@@ -698,7 +796,7 @@ export function Settings({ isCloudMode = false }: { isCloudMode?: boolean }) {
               <div className="flex items-center gap-3 mt-6 pt-6 border-t">
                 <button
                   onClick={handleSave}
-                  disabled={!hasChanges || saving}
+                  disabled={!hasChanges || saving || retentionError !== null}
```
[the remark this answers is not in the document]
KIvanow: false positive

jamby77 (approved):
Final pass on `e3d68070`. Approving.

Withdrawing my two repeat findings (upgrade loses the 7-day sample trim; enterprise hot keys 365) — you'd already marked both as intended with release-note coverage, I should have read that before re-raising them. Same for the AI-samples/OTel cap; the text-fix resolution is right.

Non-blocking follow-ups, none of which need to hold this PR:

- **Stale cache when the settings row disappears** (`apps/api/src/settings/settings.service.ts:65`) — the flip side of the `refreshCache()` fix. Keeping the previous cache was the right call; a `warn` when `getSettings()` returns null after a successful load would make the trade-off visible.
- **`CLOUD_MODE=no` / `off` count as cloud** (`apps/api/src/common/utils/cloud-mode.ts:12`) — only `false`/`0` are negatives. Now that every site shares the helper, worth pinning the accepted values (or rejecting anything but `true`/`false` in `env.schema`).
- **SQLite chunk size** (`apps/api/src/storage/adapters/sqlite-chunked-delete.ts:9`) — 10k rows is copied from the Postgres helper, but better-sqlite3 batches are synchronous and maintain every index; measured p50 245 ms / max 552 ms per batch on 1M rows. 1–2k rows keeps the yield meaningful.
- **Tab-switch discard leaves `hasChanges` stale** (`apps/web/src/pages/Settings.tsx:343`) — typing a valid then an invalid value and switching tabs reverts the form but keeps Save enabled; clicking it sends an empty update.
- **`getRetentionDays()!` / `getSampleRetentionMs()!`** (`proprietary/data-retention/data-retention.service.ts:36`) — unreachable today, but a future null becomes cutoff = now and deletes everything. An explicit guard is cheap insurance.

PR #436, opened 2026-09-04
chore: retention follow-ups from the PR #430 approval notes

KIvanow:
## Summary
The five non-blocking follow-ups Petar listed while approving #430. All are small hardening or hygiene items; none change the retention model.

## Changes
- Warn when the settings row disappears on a cache refresh (previously the old cache was kept silently)
- `CLOUD_MODE` treats `no` and `off` as negatives, matching the `BETTERDB_TELEMETRY` parser
- SQLite chunk size lowered to 2k rows: synchronous better-sqlite3 batches at 10k rows blocked the event loop ~245ms p50 (Petar's measurement on 1M rows)
- The tab-switch draft discard recomputes `hasChanges`, so Save is not left enabled by a draft that no longer exists
- The cloud sweep guards against a null policy window instead of using non-null assertions (a future null would have meant a zero cutoff, deleting everything)

## Checklist
- [ ] Docs added / updated (not needed: no behavior contract changed)

---

> [!NOTE]
> **Medium Risk**
> Touches cloud retention sweep cutoffs and CLOUD_MODE semantics, but changes are defensive (skip sweep, align env parsing) rather than broad policy shifts.
> 
> **Overview**
> Small hardening and hygiene follow-ups around retention and deployment flags; the retention model itself is unchanged.
> 
> **`CLOUD_MODE`** now treats `no` and `off` (with existing `false`/`0`) as explicit non-cloud values via a shared negative set aligned with **`BETTERDB_TELEMETRY`** parsing; env schema tests cover those spellings.
> 
> **Settings** logs a warning when a periodic cache refresh finds no DB settings row but still keeps the last loaded cache (instead of going silent).
> 
> **SQLite chunked deletes** use a **2k** row batch size (down from 10k) so synchronous better-sqlite3 work yields the event loop more often during large retention prunes.
> 
> **Settings UI**: leaving another tab while Data Retention has an invalid draft still discards the retention draft, and now **recomputes `hasChanges`** so Save is not left enabled.
> 
> **Cloud retention sweep** skips the run if tier policy ever returns a null window, avoiding a zero cutoff that would prune everything; non-null assertions are removed in favor of that guard.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14ad31b23b5805619a3d7a2993fb2aebc71c863e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Cloud mode now correctly recognizes additional “off” values, including `no`, `off`, and `False`.
  * Invalid retention entries are fully reverted when leaving the Data Retention settings, preventing an incorrectly enabled Save button.
  * Retention cleanup now safely skips execution when required cloud retention settings are unavailable, preventing unintended data deletion.
  * Settings remain available during refresh if persisted settings temporarily cannot be found.

### DIFF
--- a/apps/api/src/common/utils/cloud-mode.ts
+++ b/apps/api/src/common/utils/cloud-mode.ts
@@ -1,6 +1,10 @@
+const NEGATIVE_VALUES = new Set(['false', '0', 'no', 'off']);
+
 /**
  * Single source of truth for the deployment mode: any non-empty CLOUD_MODE
- * value except 'false'/'0' is cloud. Parts of the codebase used to disagree
+ * value outside the negative set is cloud. The negative set matches the
+ * BETTERDB_TELEMETRY parser so the two flags read the same spellings the
+ * same way. Parts of the codebase used to disagree
  * (`if (process.env.CLOUD_MODE)` vs `=== 'true'`), splitting a deployment
  * into half-cloud behavior for values like CLOUD_MODE=1.
  *
@@ -9,7 +13,7 @@
  */
 export function isCloudModeValue(value: string | undefined): boolean {
   const v = value?.trim().toLowerCase();
-  return !!v && v !== 'false' && v !== '0';
+  return !!v && !NEGATIVE_VALUES.has(v);
 }
 
 export function isCloudMode(): boolean {
--- a/apps/api/src/settings/settings.service.ts
+++ b/apps/api/src/settings/settings.service.ts
@@ -64,6 +64,10 @@ export class SettingsService implements OnModuleInit, OnModuleDestroy {
     // must not be able to trigger deletion.
     if (dbSettings) {
       this.cachedSettings = dbSettings;
+    } else if (this.cachedSettings) {
+      this.logger.warn(
+        'Settings row missing on cache refresh; keeping the previously loaded settings',
+      );
     }
   }
 
--- a/apps/web/src/pages/Settings.tsx
+++ b/apps/web/src/pages/Settings.tsx
@@ -333,13 +333,27 @@ export function Settings({ isCloudMode = false }: { isCloudMode?: boolean }) {
                 if (category.id !== 'dataRetention' && retentionError) {
                   // Discard the WHOLE draft, including any valid prefix that
                   // was committed to formData while typing (e.g. "3" en route
-                  // to "3650") — reverting only the visible input would let a
+                  // to "3650"): reverting only the visible input would let a
                   // hidden partial value ride along with a save made from
                   // another tab and silently shrink the retention window.
-                  setFormData((prev) => ({
-                    ...prev,
+                  // Recompute hasChanges from the reverted form so Save is
+                  // not left enabled by a draft that no longer exists.
+                  const reverted = {
+                    ...formData,
                     localRetentionDays: settings?.localRetentionDays ?? null,
-                  }));
+                  };
+                  setFormData(reverted);
+                  setHasChanges(
+                    settings
+                      ? (Object.keys(reverted) as Array<keyof AppSettings>).some(
+                          (key) =>
+                            key !== 'id' &&
+                            key !== 'createdAt' &&
+                            key !== 'updatedAt' &&
+                            reverted[key] !== settings[key],
+                        )
+                      : false,
+                  );
                   syncRetentionFrom(settings?.localRetentionDays);
                 }
               }}
--- a/proprietary/data-retention/data-retention.service.ts
+++ b/proprietary/data-retention/data-retention.service.ts
@@ -30,14 +30,21 @@ export class DataRetentionService {
       return;
     }
 
-    // In cloud mode the policy always resolves to the tier window, never
-    // null; the day count identifies the tier, so we don't resolve the tier
-    // a second time just for the log (two lookups could disagree mid-sweep).
-    const retentionDays = this.retentionPolicy.getRetentionDays()!;
+    // In cloud mode the policy resolves to the tier window today, but guard
+    // anyway: if a future change ever returned null here, the arithmetic
+    // below would compute cutoff = now and the sweep would delete everything.
+    const retentionDays = this.retentionPolicy.getRetentionDays();
+    const sampleRetentionMs = this.retentionPolicy.getSampleRetentionMs();
+    if (retentionDays == null || sampleRetentionMs == null) {
+      this.logger.error(
+        'Retention policy returned no window in cloud mode; skipping the sweep rather than pruning with a zero cutoff',
+      );
+      return;
+    }
     const cutoff = Date.now() - retentionDays * MS_PER_DAY;
-    // The sample stores keep their tighter cloud cap even in the sweep — it
+    // The sample stores keep their tighter cloud cap even in the sweep: it
     // is the only pruner that reaches removed/unreachable connections' rows.
-    const sampleCutoff = Date.now() - this.retentionPolicy.getSampleRetentionMs()!;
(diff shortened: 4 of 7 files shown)


apps/api/src/settings/settings.service.ts
```diff
@@ -64,6 +64,10 @@ export class SettingsService implements OnModuleInit, OnModuleDestroy {
     // must not be able to trigger deletion.
     if (dbSettings) {
       this.cachedSettings = dbSettings;
+    } else if (this.cachedSettings) {
+      this.logger.warn(
```
jamby77, line 68: No once-per-episode dedup. `refreshCache()` runs on a 30 s interval and nothing re-seeds a missing row, so a wiped `app_settings` (or a bad migration dropping the row) logs this identical line ~2,880 times a day until restart, burying the one actionable signal. `RetentionPolicyService.warnedInvalidStoredValue` already has the warn-once-then-reset pattern to copy: set a flag here, clear it in the `if (dbSettings)` branch.

proprietary/data-retention/data-retention.service.ts
```diff
@@ -30,14 +30,21 @@ export class DataRetentionService {
...
-    // In cloud mode the policy always resolves to the tier window, never
-    // null; the day count identifies the tier, so we don't resolve the tier
-    // a second time just for the log (two lookups could disagree mid-sweep).
-    const retentionDays = this.retentionPolicy.getRetentionDays()!;
+    // In cloud mode the policy resolves to the tier window today, but guard
+    // anyway: if a future change ever returned null here, the arithmetic
+    // below would compute cutoff = now and the sweep would delete everything.
+    const retentionDays = this.retentionPolicy.getRetentionDays();
+    const sampleRetentionMs = this.retentionPolicy.getSampleRetentionMs();
+    if (retentionDays == null || sampleRetentionMs == null) {
```
jamby77, line 38: This guard is now the only thing between a null policy window and `cutoff = Date.now()`, but nothing exercises it: `data-retention.service.spec.ts` stubs `getRetentionDays` / `getSampleRetentionMs` to always return numbers (lines 58-59), and the PR adds no spec under `proprietary/`. A later refactor that drops the guard or inverts it (`== null &&`) passes green, and the 03:00 cron deletes every tenant's history — exactly the case the comment above says this prevents. One test with the policy returning `null`, asserting `logger.error` and zero prune calls, closes it.

apps/api/src/common/utils/cloud-mode.ts
```diff
@@ -1,6 +1,10 @@
+const NEGATIVE_VALUES = new Set(['false', '0', 'no', 'off']);
+
 /**
  * Single source of truth for the deployment mode: any non-empty CLOUD_MODE
- * value except 'false'/'0' is cloud. Parts of the codebase used to disagree
+ * value outside the negative set is cloud. The negative set matches the
+ * BETTERDB_TELEMETRY parser so the two flags read the same spellings the
```
jamby77, line 6: "The negative set matches the BETTERDB_TELEMETRY parser so the two flags read the same spellings the same way" is not true as written: `isCloudModeValue` does `value?.trim().toLowerCase()`, while the `BETTERDB_TELEMETRY` transform in `env.schema.ts` and `telemetry-client.factory.ts:22` only `.toLowerCase()`. A compose file with `CLOUD_MODE=" off "` and `BETTERDB_TELEMETRY=" off "` reads the first as self-hosted and the second as opted-in. Either trim in both telemetry sites or drop the parity claim.

apps/api/src/common/utils/cloud-mode.ts
```diff
@@ -1,6 +1,10 @@
+const NEGATIVE_VALUES = new Set(['false', '0', 'no', 'off']);
```
jamby77, line 1: This is now a third literal copy of the same list — `env.schema.ts` (`BETTERDB_TELEMETRY` transform, ~line 113) and `telemetry-client.factory.ts:22` each inline `['false', '0', 'no', 'off']`. Parity is asserted only by the comment above; adding `'disabled'` to one silently diverges the others with no test failing, which is the drift `cloud-mode.ts` was created to end. Export one `isNegativeEnvValue(v)` from here (or a `common/utils/env-bool.ts`) and call it from both telemetry sites.

apps/web/src/pages/Settings.tsx
```diff
@@ -333,13 +333,27 @@ export function Settings({ isCloudMode = false }: { isCloudMode?: boolean }) {
...
-                    ...prev,
+                  // Recompute hasChanges from the reverted form so Save is
+                  // not left enabled by a draft that no longer exists.
+                  const reverted = {
+                    ...formData,
                     localRetentionDays: settings?.localRetentionDays ?? null,
-                  }));
+                  };
+                  setFormData(reverted);
+                  setHasChanges(
```
jamby77, line 346: This narrows the symptom but doesn't fix it: `handleInputChange` still sets `hasChanges = true` unconditionally, so the same "Save enabled with nothing to send" is reachable without any error. Stored `localRetentionDays = null` → type `30` (commits, `hasChanges = true`) → clear the field (valid; commits `null`, `hasChanges` stays true). `formData` now equals `settings`, Save is enabled, and clicking it PUTs `{}`.

The predicate here is also a second hand-written copy of the changed-field loop in `handleSave` (lines 104-108) with its own `id`/`createdAt`/`updatedAt` exclusion list; the two will drift the next time a server-managed field is added. Deriving `hasChanges` from `(formData, settings)` via one `changedKeys(form, saved)` helper (or `useMemo`) removes all six `setHasChanges` call sites and this duplication, and fixes the residual case.

jamby77 (commented):
Reviewed at `14ad31b2`. Nothing blocking; the five follow-ups do what they say. Inline notes on the diff, plus two that fall outside the hunks:

**Stale contract comments** — `proprietary/cloud-auth/cloud-auth.guard.ts:20` and `cloud-auth.middleware.ts:20` still document only `false`/`0` as negatives; now that `no`/`off` are too, those two lines are wrong.

**`chore:` prefix** — two of the five hunks are user-visible behaviour changes (`CLOUD_MODE=no`/`off` now boots self-hosted; the Save-button state), and the body's "no behavior contract changed" undersells the first one. A changelog reader filtering on `bugfix` will miss the cloud-mode flip. Arguable on majority count, but worth a `bugfix:` squash title or a line in the release notes.

Pre-existing, not for this PR: `proprietary/entitlement/src/provisioning/provisioning.service.ts:115` still reads `CLOUD_MODE === 'true'` directly. Entitlement only depends on `@betterdb/shared`, so moving `isCloudModeValue` there would let it share the helper.

jamby77 (approved):
Approving at 753d7b35 — all 21 checks green. Went through each of the five findings from the September 4th pass rather than trusting the summary:

- `isNegativeEnvValue` is now the one negative set behind `CLOUD_MODE`, the `BETTERDB_TELEMETRY` transform and the telemetry factory, so the trim-parity claim in the comment is finally true.
- Missing-row warning is once per episode, re-armed when the row returns.
- The null retention-window guard has a spec asserting the logged error and zero prune calls.
- `hasChanges` derives from `changedKeys`, which closes the Save-enabled / `PUT {}` path.
- Cloud-auth comments now list the full negative set.

Good catch from CodeRabbit on the flag surviving `updateSettings`/`resetToDefaults` — routing both through `commitCacheWrite` is the right shape, and the ordering against the generation guard in `refreshCache` is correct, so an in-flight refresh can't clobber a direct write or swallow the next episode's warning.

PR #438, opened 2026-09-05
fix(web): log swallowed errors instead of silencing them

Kathircpe:
Replace empty catches with console.debug/warn in telemetry client, useConnection, VectorSearch snapshots/memory, and VectorAi per-index fetch. Fire-and-forget and resilient behavior preserved.

## Summary

Empty `.catch(()=>{})` made telemetry outages and partial index-load
failures undiagnosable. VectorAi in particular dropped failed indexes
with no signal.

Stops silently swallowing errors in 4 spots:
- `telemetry/clients/api-telemetry-client.ts` — log telemetry POST failures at debug
- `hooks/useConnection.ts` — log `connection_switch` telemetry failures at debug
- `pages/VectorSearch.tsx` — log snapshot/memory failures at debug (AbortError suppressed for snapshots)
- `pages/VectorAi.tsx` — log per-index failures at warn, still omit failed rows

## Checklist
- [x] Docs added / updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostic logging for telemetry, connection switching, vector index loading, memory information, and snapshot requests.
  * Aborted snapshot requests are now handled without unnecessary error logging.
  * Vector index filtering continues to work when individual index-information requests fail.
  * Snapshot requests are canceled when leaving the relevant view, reducing unnecessary background activity.

### DIFF
--- a/apps/web/src/api/metrics.ts
+++ b/apps/web/src/api/metrics.ts
@@ -460,12 +460,13 @@ export const metricsApi = {
       method: 'POST',
       body: JSON.stringify(params),
     }),
-  getVectorIndexSnapshots: (name: string, hours?: number) => {
+  getVectorIndexSnapshots: (name: string, hours?: number, signal?: AbortSignal) => {
     const q = new URLSearchParams();
     if (hours) q.set('hours', hours.toString());
     const qs = q.toString();
     return fetchApi<{ snapshots: VectorIndexSnapshot[] }>(
       `/vector-search/indexes/${encodeURIComponent(name)}/snapshots${qs ? `?${qs}` : ''}`,
+      { signal },
     );
   },
   textSearch: (indexName: string, params: { query: string; offset?: number; limit?: number }) =>
--- a/apps/web/src/hooks/useConnection.ts
+++ b/apps/web/src/hooks/useConnection.ts
@@ -104,7 +104,9 @@ export function useConnectionState(): ConnectionContextValue {
             dbVersion: connection.capabilities?.version ?? 'unknown',
           },
         }),
-      }).catch(() => {});
+      }).catch((err) => {
+        console.debug('[telemetry] failed to capture connection_switch:', err);
+      });
     }
   }, [connections]);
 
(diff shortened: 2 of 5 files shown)


Kathircpe:

KIvanow (changes requested):
Hi @Kathircpe, thanks for the PR! Getting rid of these silent `.catch(() => {})` handlers is a real improvement - the fire-and-forget semantics are preserved everywhere and the overall approach looks good. A couple of changes before we can merge:

**1. `VectorSearch.tsx` - the AbortError guard is dead code**

```ts
if (err instanceof DOMException && err.name === 'AbortError') return;
```

This branch can never fire: `getVectorIndexSnapshots` (`api/metrics.ts`) doesn't accept an `AbortSignal`, and the enclosing `useEffect` never creates an `AbortController` or returns a cleanup, so nothing ever aborts this request. As written it implies cancellation support that doesn't exist.

Either is fine with us:
- simply drop the guard, or
- (nicer) actually wire up cancellation - `fetchApi` already forwards `options.signal`, so threading a signal through `getVectorIndexSnapshots` and aborting in the effect cleanup would make the guard real *and* fix the latent stale-response race when `snapshotHours` or the connection changes mid-flight.

**2. `VectorAi.tsx` - `console.warn` inside the polling loop will flood the console**

The fetcher runs every 30s (`INDEX_POLL_MS`) across all indexes, so one persistently failing index - or a backend outage with N indexes - emits N warnings every 30 seconds, which drowns out exactly the signal this PR is adding. The other four sites use `console.debug` (hidden at default verbosity); please use `debug` here too, or only `warn` on the first failure per index.

**3. Commit signing**

Our repo requires signed commits, and `987eea91` is currently unsigned. Could you sign and force-push? Typically:

```bash
git commit --amend --no-edit -S
git push --force-with-lease
```

If you haven't set up signing before, GitHub's guide is here: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits (SSH key signing is the quickest path).

Happy to re-review as soon as these are in - thanks again!

Kathircpe:
@KIvanow Can I get a review?

KIvanow (approved):
Thanks @Kathircpe, both fixes look great - wiring real cancellation into the snapshot fetch was the perfect call, and the commit is signed. LGTM! 🚀

PR #445, opened 2026-09-09
Harden web fetchApi: safe body parsing, opt-in bounded timeout

Kathircpe:
## Summary

Hardens apps/web/src/api/client.ts (fetchApi)

Fixes: blind response.json() crashing on 204/empty/non-JSON successes; response.text() rejection in the error path masking real HTTP errors.

Adds: opt-in timeoutMs, validated/clamped to [MIN_API_TIMEOUT_MS, MAX_API_TIMEOUT_MS] (1s–120s) and combined with any caller AbortSignal. When omitted (or non-numeric) no timer is armed — existing behaviour is fully preserved, so no caller changes were needed.

## Checklist
- [ ] Docs added / updated

## Summary by CodeRabbit

* **New Features**
  * API requests can use an optional timeout, automatically constrained to safe minimum and maximum limits.
  * Caller-provided cancellation and request timeouts work together.
* **Bug Fixes**
  * Improved handling of empty successful responses, including no-content responses.
  * Added clearer errors for invalid response formats and malformed JSON.
  * Improved resilience when error-response details cannot be read.
  * Timeout failures now clearly indicate when a request exceeds its configured limit.

### DIFF
--- a/apps/web/src/api/client.test.ts
+++ b/apps/web/src/api/client.test.ts
@@ -102,34 +102,72 @@ describe('fetchApi timeoutMs', () => {
   }, 5000);
 
   it('aborts a hung request with TimeoutError when timeoutMs is provided', async () => {
-    vi.spyOn(globalThis, 'fetch').mockImplementation(
-      (_url, init) =>
-        new Promise((_resolve, reject) => {
-          init?.signal?.addEventListener('abort', () =>
-            reject((init.signal as AbortSignal).reason ?? new Error('aborted')),
-          );
-        }),
+    vi.useFakeTimers();
+    try {
+      vi.spyOn(globalThis, 'fetch').mockImplementation(
+        (_url, init) =>
+          new Promise((_resolve, reject) => {
+            init?.signal?.addEventListener('abort', () =>
+              reject((init.signal as AbortSignal).reason ?? new Error('aborted')),
+            );
+          }),
+      );
+
+      const assertion = expect(fetchApi('/hung', { timeoutMs: 1000 })).rejects.toMatchObject({
+        name: 'TimeoutError',
+      });
+      await vi.advanceTimersByTimeAsync(1000);
+      await assertion;
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it.each([
+    { timeoutMs: 1, firesAtMs: 1000 },
+    { timeoutMs: 0, firesAtMs: 1000 },
+    { timeoutMs: -50, firesAtMs: 1000 },
+    { timeoutMs: 999_999_999, firesAtMs: 120_000 },
+  ])('clamps timeoutMs=$timeoutMs to a $firesAtMs ms abort', async ({ timeoutMs, firesAtMs }) => {
+    vi.useFakeTimers();
+    try {
+      vi.spyOn(globalThis, 'fetch').mockImplementation(
+        (_url, init) =>
+          new Promise((_resolve, reject) => {
+            init?.signal?.addEventListener('abort', () =>
+              reject((init.signal as AbortSignal).reason ?? new Error('aborted')),
+            );
+          }),
+      );
+
+      let settled: unknown = 'pending';
+      const pending = fetchApi('/x', { timeoutMs }).then(
+        () => {
+          settled = 'resolved';
+        },
+        (error) => {
+          settled = error;
+        },
+      );
+      await vi.advanceTimersByTimeAsync(firesAtMs - 1);
+      expect(settled).toBe('pending');
+      await vi.advanceTimersByTimeAsync(1);
+      await pending;
+      expect(settled).toMatchObject({ name: 'TimeoutError' });
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it.each([204, 205])('resolves undefined for empty %i responses', async (status) => {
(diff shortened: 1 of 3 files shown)


KIvanow (changes requested):
Thanks - the hardening here is solid overall: the 204/empty-body handling, the `text()` rejection fix in the error path, and the signal composition (including the `AbortSignal.any` fallback and the `finally` cleanup) all look correct, and the tests cover the important paths.

One thing I'd like resolved before merging:

**`timeoutMs: 0` contradicts the documented clamping contract** (`apps/web/src/api/client.ts`, in `combineSignals`)

The JSDoc on `FetchApiOptions.timeoutMs` (and the PR description) says values outside `[MIN_API_TIMEOUT_MS, MAX_API_TIMEOUT_MS]` are clamped to the nearest bound. But the guard:

```ts
if (!timeoutMs || !Number.isFinite(timeoutMs)) {
  return { signal: callerSignal ?? undefined, cleanup: () => {} };
}
```

treats `0` as falsy, so `timeoutMs: 0` silently disables the timeout entirely - while `timeoutMs: -1` falls through and clamps to 1s. Two adjacent out-of-range values behave in opposite ways, and neither `0` case matches the docs.

Either resolution works for me:

1. **Make `0` clamp like other out-of-range values** - change the guard to explicit nullish/NaN detection:
   ```ts
   if (timeoutMs == null || !Number.isFinite(timeoutMs)) {
   ```
2. **Keep `0` as "disabled" but make it deliberate** - document `0` (and negatives?) as "no timeout" in the JSDoc and add a test asserting it, so the behavior is contractual rather than accidental.

I'd lean toward option 1, since callers passing a computed value that hits `0` almost certainly wanted "a very short timeout," not "wait forever." Whichever you pick, a test pinning the `0` case would prevent this from regressing.

Kathircpe:
fixed it using the first option

KIvanow (changes requested):
Really solid hardening work - I walked `combineSignals` through every path (pre-aborted caller signal, `AbortSignal.any` vs. fallback, listener/timer cleanup, browser `unref` guard) and found no bugs, and the `DOMException`/`TimeoutError` choice matching `AbortSignal.timeout()` semantics is a nice touch. Requesting changes on three points before merge:

1. **Return `undefined` instead of `null` for empty bodies.** #452 (which merges first and closes #448) resolves 204/205/empty successes to `undefined`, and existing callers type these as `fetchApi<void>`. `parseSuccessPayload` returning `null as T` would silently change that contract on rebase. Please switch the two `return null as T` sites to `undefined as T` so the two PRs agree - no current caller branches on the value, so this is free now but a footgun later. (You'll need to rebase over #452's version of `client.ts`/`client.test.ts` anyway.)

2. **Use `new Error(message, { cause: error })` instead of the `eslint-disable-next-line preserve-caught-error`.** The rule exists precisely to ask for the `cause` option, and we run `reportUnusedDisableDirectives: 'error'` - attaching the cause preserves the original stack rather than flattening it into the message string.

3. **Document the `timeoutMs: 0` behavior in the JSDoc.** Right now `0` gets clamped to a 1s timeout - a caller passing `0` to mean "no timeout" gets a fast abort instead, and the only place this is discoverable is a test. One line in the `FetchApiOptions` doc ("`0` is clamped to 1s; omit the option to disable") settles it.

Nice-to-haves, not blocking: `vi.useFakeTimers()` would make the hung-request test instant (it currently burns a real 1s on the MIN clamp) and would let the clamp test assert actual durations; and there's a stray blank line at the top of `combineSignals`.

Kathircpe:
- Resolve 204/205/empty successes to undefined instead of null,
  aligning with #452 's fetchApi<void> contract
- Attach the read failure via Error { cause } instead of flattening
  it into the message; add ES2022.Error lib for typings (target
  stays ES2020, no runtime change)
- Document that timeoutMs: 0 clamps to MIN (omit to disable)
- Convert timeout tests to fake timers with exact abort-instant
  assertions, incl. 0/negative clamp cases and empty-body pins

KIvanow (approved):
Thank you for addressing my feedback @Kathircpe !

KIvanow:
There is  now a small c onflict with your merged PR that will need a look at @Kathircpe

Kathircpe:
Resolved it and I removed our 2 overlapping empty-body tests; kept #452's 3 canonical regression tests + our 7 timeout tests.

KIvanow (approved):

KIvanow:
@Kathircpe sorry, but can you please sign your commits. You've done that in a previous PR

Kathircpe:
@KIvanow Sorry, forgot to sign it.
Signed and force pushed it.

KIvanow:
Btw, since you're now signing commits @Kathircpe  - a tip so you never have to rebase-to-sign again: git can auto-sign per folder via conditional includes, so it only applies where you want it. I use a variation of this so I don't have to bother with signing my random side project commits

`~/.gitconfig`:
```ini
[includeIf "gitdir:~/work/betterdb/"]
    path = ~/.gitconfig-betterdb
```

`~/.gitconfig-betterdb`:
```ini
[commit]
    gpgsign = true
[user]
    signingkey = <YOUR_KEY_ID>
```

Every repo under `~/work/betterdb/` signs automatically; everything else is untouched. Works with SSH signing keys too (`gpg.format = ssh`).

PR #455, opened 2026-09-13
feature: personal MCP tokens and workspace telemetry for self-hosted user control

jamby77:
Stacked on #437 (user control phase 4). Phase 6 of self-hosted user control: optional personal MCP tokens and workspace telemetry.

## What changes

**Personal MCP tokens (self-hosted)**
- Any signed-in user can create a personal MCP token in Settings → MCP Tokens. The token is shown once. It has a `bdb_mcp_` prefix, is stored as a sha256 hash and expires after 365 days.
- Members see and revoke their own tokens. Admins see and revoke everyone's, including the tokens of removed members, which are labelled as such.
- A token acts as its owner with the owner's **current** role. A demoted admin's token loses write access on the next call. Tokens of a removed owner stop working.
- `/mcp/` stays public, so MCP read tools keep working without a token. With a token, calls act as the owner, and any changes they make appear under the owner in the activity log. The activity log still records mutations only.
- Tokens cannot manage identity or access. Creating or revoking tokens, inviting, revoking invitations, removing members, changing roles and transferring ownership all require a signed-in session, so a leaked token cannot mint persistent access.
- The actor is resolved from the session first, then from `Authorization: Bearer`. WebSocket upgrades (monitor tail, CLI) stay session-only. Sign-out attribution is session-only, so a token cannot write `auth.logout` rows under its owner.
- `lastUsedAt` is stamped on every token call. A failed stamp is logged and never fails the request.
- Cloud is unchanged. The proprietary token service stamps `userId: null`.

**Storage**
- `agent_tokens.user_id` column plus migrations for memory, sqlite and postgres. Legacy rows stay ownerless and can never resolve to a user.

**Telemetry**
- New instance-level events: `workspace_first_register`, `user_login` and `member_removed`. Payloads carry no identifying data.

## Testing
- API unit suites green, including the new token service, resolver/guard bearer paths, the controller's session-only gating, and a role-transition test on an already-issued token.
- Full-app e2e (`workspace-mcp-tokens.e2e-spec.ts`, sqlite) covers the lifecycle: mint, attributed mutation, bearer invite gets 403 and creates no invitation, revoke gives 401, and anonymous reads still work.
- Web: `McpTokensPanel` tests cover generation, listing, owner labels, and clipboard success, missing clipboard and rejected writes.
- The Postgres migration spec (`agent-token-storage-postgres.e2e-spec.ts`) runs in its own schema in CI. It was not run locally.

## Follow-ups (not in this PR)
- Public-path actor resolution runs on every public prefix (`/auth/*`, `/prometheus`, `/v1/traces`, `/ingest/`), not only `/mcp/`; consider narrowing it to MCP.
- Token activity rows show only the token id; adding the token name to `details` would make the log easier to read.
- On public MCP reads an invalid or revoked token silently falls back to anonymous; a response header could tell the client.

---

> [!NOTE]
> **High Risk**
> Expands authentication with bearer personal tokens, guard/path behavior on public MCP routes, and new persistence for token ownership—security-sensitive surface area despite session gating on privileged actions.
> 
> **Overview**
> Adds **personal MCP tokens** for self-hosted (and cloud UI parity): signed-in users mint `bdb_mcp_*` keys via **`/agent-tokens`**, stored hashed with an owning **`user_id`**, and use them as **`Authorization: Bearer`** after the session cookie. **`ActorResolver`** and **`ActorGuard`** resolve session first, then bearer; **`/mcp/`** stays publicly readable but optionally attributes the caller when credentials are present; WebSocket upgrades and sign-out attribution stay **session-only**. **`requireSession`** blocks tokens from invites, membership changes, and token lifecycle APIs.
> 
> Storage gains **`agent_tokens.user_id`** (memory/sqlite/postgres migrations; legacy rows stay ownerless). Auth telemetry adds **`workspace_first_register`**, **`user_login`**, and **`member_removed`**.
> 
> The web app moves MCP token management to **`/account/mcp-tokens`** (sidebar + shared **`McpTokensPanel`**), fixes **`BETTERDB_URL`** in generated MCP config via **`apiOrigin()`**, and documents optional vs required tokens for cloud vs self-hosted.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3039c59c856d215acba74d4f92eda134372a98a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added MCP token management under **Settings → MCP Tokens**, including creation, listing, copying, expiration status, owner details, and revocation.
  - MCP tokens can authenticate requests and attribute activity to their owner.
  - Self-hosted deployments can use MCP tokens, while demo environments cannot create them.
  - MCP read operations remain available anonymously where supported.

- **Bug Fixes**
  - Restricted membership and access-management actions to session-based authentication.
  - Revoked or expired tokens no longer authorize requests.

- **Documentation**
  - Clarified cloud and self-hosted MCP token requirements and behavior.


apps/web/src/pages/Settings.tsx
```diff
@@ -57,15 +71,8 @@ export function Settings({ isCloudMode = false }: { isCloudMode?: boolean }) {
...
-  const { tokens: mcpTokens, invalidate: invalidateMcpTokens } = useMcpTokens(
-    isCloudMode && activeCategory === 'mcpTokens',
-  );
-  const [mcpTokenName, setMcpTokenName] = useState('');
-  const [mcpGenerating, setMcpGenerating] = useState(false);
-  const [mcpGeneratedToken, setMcpGeneratedToken] = useState<GeneratedToken | null>(null);
-  const [mcpCopied, setMcpCopied] = useState(false);
-  const [mcpError, setMcpError] = useState<string | null>(null);
+  const { mode } = useAuth();
+  const showMcpTokens = isCloudMode === true || mode === 'self-hosted';
```
jamby77, line 75: [P2] Allow members to reach their personal token settings

This enables the MCP Tokens category for self-hosted members, but AppSidebar still renders the Settings item with locked={canMutate === false} and the tooltip 'Admins only'. NavItem renders a non-clickable span in that case, so members cannot follow the advertised Settings → MCP Tokens flow despite being allowed to create tokens by the API. Please provide member-accessible navigation to personal token settings while retaining restrictions on administrative controls, and cover the member navigation flow in a test.

apps/web/src/components/pages/settings/McpTokensPanel.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+function clientConfig(token: string): string {
+  return JSON.stringify(
+    {
+      mcpServers: {
+        betterdb: {
+          type: 'stdio',
+          command: 'npx',
+          args: ['@betterdb/mcp'],
+          env: { BETTERDB_URL: window.location.origin, BETTERDB_TOKEN: token },
```
jamby77: [P2] Use the API origin in the generated development config

With pnpm dev, window.location.origin is http://localhost:5173, while the API client uses http://localhost:3001. Vite only proxies /ingest, so copying this newly available self-hosted MCP configuration sends the MCP client to the frontend server rather than the API. Please derive BETTERDB_URL from the API configuration (preserving the production origin behavior) and test both development and production URLs.

jamby77 (commented):
Two actionable findings from the review. Validation: 73 focused API tests and 10 MCP token-panel tests passed with the compatible Node runtime. PostgreSQL and full-app e2e tests were not run.

apps/web/src/pages/Settings.tsx
jamby77: Fixed with a separate page. `/account/mcp-tokens` reuses the existing token panel and is linked from the sidebar user menu for every self-hosted user, members included. Settings and its admin-only guard are unchanged, so admins keep their existing path. Tests cover a member reaching the page from the menu, a member still being kept out of `/settings`, and an admin reaching the page too.

apps/web/src/components/pages/settings/McpTokensPanel.tsx
jamby77: Fixed. `BETTERDB_URL` in the generated MCP config now comes from an `apiOrigin()` helper. It resolves to the API's origin under `pnpm dev` (`localhost:3001`) and stays `window.location.origin` in production, as before. Tests cover both.

KIvanow (changes requested):
**Review: phase 6 personal MCP tokens**

Reviewed the token service, resolver integration, and guards; validated locally (212 workspace/auth/guard tests pass, `apps/api` + `apps/web` `tsc` clean). Security model is solid:

- Tokens are 256-bit CSPRNG with a `bdb_mcp_` prefix, stored only as a SHA-256 hash, 365-day expiry.
- `resolveActor` re-fetches the owner via `findById`, so a token always acts at the owner's **current** role and a demoted/removed owner's token stops working immediately. Revoked/expired tokens are rejected.
- `canManage` returns NotFound (not Forbidden) to members, so token existence isn't leaked; `requireSession` blocks a token from minting or revoking tokens (no token-to-token escalation).
- Clean cloud/self-hosted split: `AgentTokenGuard`/`MCP_TOKEN_SERVICE` is cloud-only, so `/mcp/` stays public in self-hosted and the token becomes an actor via `ActorResolver.resolveBearer`; a member's token still can't perform MCP writes (MutationGuard).
- Good call switching session resolution to `disableCookieCache: true` - that closes the stale-role window I flagged on #433.

Changes I'd like before merge:

1. **Quantify the per-request DB cost of `disableCookieCache: true` + `optionalActor` on public paths.** Every authenticated HTTP request and every CLI command (up to 50/s per socket) now does a fresh `getSession` DB lookup, and `ActorGuard.optionalActor` adds a session/token lookup to every *public* request that carries a cookie or `Authorization` header. Same-origin browser calls to `/ingest/*` and `/v1/traces` (telemetry/traces) and chatty `/mcp/` tools will now each hit the DB. Please confirm this is acceptable under load, or exempt the high-volume public telemetry/trace endpoints from `optionalActor` (they don't need an actor).

2. **`resolveActor` writes `lastUsedAt` on every token call.** For a busy MCP client that's a DB write per request. Consider throttling it (e.g. update at most once per minute per token).

Note: the `CLOUD_MODE` fail-open (#432) and cloud-WS (#433) items still apply to this stack.

jamby77:
Replies to the phase 6 review:

1. **Per-request DB cost.**
   - Public paths: fixed in 7fba304d. `optionalActor` now resolves only on `/mcp/` (with or without the `/api` prefix); ingest, traces and auth paths skip the session and token lookup.
   - `disableCookieCache`: kept deliberately. It comes from #434 (3401dcf4), so a demotion, removal or sign-out takes effect on the next request. The cost is one indexed session-token read per authenticated HTTP request. Not benchmarked. The dashboard's polling was already paying that before the cookie cache.
   - CLI: bounded by the per-socket 30s access cache from #433 (ccfb82c5), so a CLI socket does at most one lookup per 30s, not up to 50/s.
2. **`lastUsedAt` writes.** Fixed in cffe6215: the stamp is written at most once a minute per token.

40bccc5f adds `UsageTelemetryService` to the auth bridge origin-check spec after the restack.

apps/web/src/components/layout/AppLayout.tsx
```diff
@@ -324,6 +326,7 @@ function AppLayoutInner(): ReactElement {
                 }
               />
               {showTeam && <Route path="/workspace/members" element={<Members />} />}
+              {showMcpTokens && <Route path="/account/mcp-tokens" element={<AccountMcpTokens />} />}
```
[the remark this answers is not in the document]
jamby77: Fixed in f48b81ce, on both sides:
- **API:** `POST /agent-tokens` returns 403 when the request arrives on `DEMO_HOSTNAME`, matching the monitor tail socket's demo block. A direct API call on the demo can't mint a token either.
- **Web:** demo visitors don't see the MCP Tokens menu link, and `/account/mcp-tokens` redirects home, as Settings already does.

Specs cover both.

apps/api/src/workspace/personal-tokens.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+      id: metadata.id,
+      name: metadata.name,
+      type: metadata.type,
+      expiresAt: metadata.expiresAt,
+    };
+  }
+
+  @Get()
+  async list(@CurrentUser() actor: Actor): Promise<PersonalTokenView[]> {
+    return this.tokens.list(actor);
```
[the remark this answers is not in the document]
jamby77: Leaving this as is. Create is the only operation that has to be blocked on the demo:

- `list` and `revoke` go through `PersonalTokenService.canManage`, so a member sees and revokes only their own tokens, and an admin only the workspace's. Nothing here reaches another tenant's data.
- Because create is refused on the demo host, a demo account has no personal tokens to list, and revoking can only take away access, never grant it.
- Matching `req.headers.host` against `DEMO_HOSTNAME` is the existing demo pattern (`system.controller.ts`, `tail.gateway.ts`), and the demo deployment routes by that host.

Blocking list/revoke too would just turn an empty tokens page into an error on the demo.

apps/web/src/components/pages/settings/McpTokensPanel.tsx
```diff
@@ -0,0 +1,292 @@
...
+    return () => {
+      if (copiedTimerRef.current !== null) {
+        clearTimeout(copiedTimerRef.current);
+      }
+    };
+  }, []);
+
+  const handleGenerate = async (): Promise<void> => {
+    if (trimmedName.length === 0) {
+      return;
```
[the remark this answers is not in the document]
jamby77: Fixed in 3039c59c. `handleGenerate` now returns early while a ref marks a generation as pending, and the name input is disabled for as long as `generating` is true, so a second Enter cannot mint a token whose secret would be lost. A new spec presses Enter twice while the request is pending and checks that `generate` runs once and the one token is shown. It fails on the old code.

KIvanow (approved):
**Re-review (head f8bde825): both items addressed, no regressions. LGTM.**

Verified (api + web `tsc` clean, 247 workspace/auth/guard tests pass):
- **Per-request DB cost (item 1)**: `ActorGuard` now only calls `optionalActor` on public paths when `isActorOptionalPath` matches, and that is scoped to the `/mcp/` surface. So `/ingest/*`, `/v1/traces`, `/prometheus`, `/health` and other public endpoints no longer trigger a session/token lookup, while the MCP write paths (`/mcp/telemetry`, `/mcp/.../recall`) still get their owner attribution. That was exactly the concern.
- **Last-use writes (item 2)**: `stampLastUsed` skips `updateAgentTokenLastUsed` when the stamp is under `LAST_USED_STAMP_INTERVAL_MS` (60s) old, so a busy MCP client no longer writes per call; the actor is still resolved and returned, so auth is unaffected.
- Bonus hardening: personal MCP token creation is now refused with 403 on `DEMO_HOSTNAME`, matching the monitor tail's demo block.
- The security model from the first pass is intact: tokens act at the owner's current role, `requireSession` still blocks token-to-token minting, and a member's token still can't do MCP writes.

Approving.

PR #459, opened 2026-09-15
feat: CVE detection → action (webhooks, metrics, fleet, UI)

Kathircpe:
**Scope disclaimer:** This PR bundles the `cve.*` webhook dispatch (backend) with the Fleet CVE column (UI) in a single change instead of shipping them as separate scoped PRs. The Fleet column is the only operator-visible proof that the webhooks fired on the right connections, so merging dispatch without its UI would leave the headline behavior unverifiable in review.

## Summary

`GET /cve/scan` detected CVEs but never acted on them — results were persisted and never dispatched. This PR closes the loop: new critical or KEV-exploited findings fire tiered webhooks, mirror to OTel/Prometheus, and surface in Fleet and Security.

### Events

| Event | Tier | Fires when |
|---|---|---|
| `cve.critical_detected` | Pro | New `critical` findings vs previous scan |
| `cve.kev_detected` | Enterprise | New CISA-KEV findings vs previous scan, even if severity < critical |

### Behavior contract

- **Fire on change, not on poll** — diff is findings-based; dataset refresh alone never pages. Unchanged fingerprints take the existing touch-path and stay silent.
- **Identity-based diff** (`nodeId|cveId|matchedOn|moduleName`) with explicit transitions: KEV `false→true` fires, only escalation *to* critical fires, scale-out on a full baseline fires.
- **First scan establishes the baseline silently** — no deploy-time page storm.
- **Partial scans never dispatch** — blip recovery doesn't page; the next full-vs-full cycle re-arms.
- **Dispatch-before-save** — a webhook failure preserves the old baseline so the next scan retries (duplicates preferred over silent loss for security findings).
- **OTel mirrors are license-independent** — OSS+OTLP deployments still get `cve.*` telemetry; webhooks stay tier-gated via `@Optional()` DI (OSS builds unaffected).
- **No UI framework changes** — polling stays as-is; banner + column reuse existing `Badge`/`Alert` patterns.

### changes

- **Shared**: event enums, tier wiring (`Pro`/`Enterprise`), `CveCritical/KevDetectedData` payloads, `FleetCveSummary` — `GET /webhooks/allowed-events` picks the new events up automatically.
- **Backend**: `dispatchCveCriticalDetected()` (Pro) / `dispatchCveKevDetected()` (Enterprise) following the anomaly/compliance precedent; `CveScanService` diff + fan-out (never throws into the scan); Fleet summary CVE rollup (critical/kev/stale, incl. down instances); `betterdb_cve_findings{connection,severity}`, `betterdb_cve_kev{connection}`, `betterdb_cve_dataset_stale{connection}` with series cleanup on missing scan, re-address, and connection removal.
- **UI**: `WebhookForm` labels (tier gating automatic), `CveAlertBanner` on Security (destructive for critical, KEV badge, counts/badges from one source), `CVEs` column + sort on Fleet.
- **Docs/rules**: `docs/webhooks.md` (tiers, payloads, baseline + dual-delivery notes), `docs/prometheus-metrics.md`, `CveKevDetected` / `CveCriticalDetected` Alertmanager rules.

---

## Checklist
- [x] Docs added / updated

## Summary by CodeRabbit

* **New Features**
  * Added CVE alert banners for critical and known-exploited findings, including scan status and top affected CVEs.
  * Added CVE counts, exploited-finding indicators, stale status, and sorting to fleet views.
  * Added Prometheus CVE metrics and alerting for critical, exploited, and stale datasets.
  * Added webhook and telemetry notifications for newly detected critical and known-exploited CVEs, with reliable retry handling.
  * CVE scans now compare complete results while avoiding notifications for initial or incomplete scans.

* **Documentation**
  * Documented CVE metrics and webhook event payloads.


apps/api/src/cve/cve-scan.service.ts
```diff
@@ -458,11 +550,192 @@ export class CveScanService extends MultiConnectionPoller implements OnModuleIni
...
+      return;
+    }
+
+    const dispatchedOk = await this.dispatchCveEvents(result, newCritical, newKev);
+    if (!dispatchedOk) {
+      this.logger.warn(
+        `CVE dispatch failed for connection ${result.connectionId}; baseline preserved for retry`,
+      );
+      return;
+    }
```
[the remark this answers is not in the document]
Kathircpe: dispatchEvent is infallible by design: outer try/catch + Promise.allSettled (webhook-dispatcher.service.ts:116-146). Endpoint outages become RETRYING/DEAD_LETTER via WebhookProcessorService, never rejections — there is nothing to propagate without changing the dispatcher contract for all existing callers. The split latest-vs-baseline storage half needs new StoragePort surface + adapter migrations, disproportionate now that the premise (throwing deliveries) is false. The ok=false path stays as a backstop for unexpected throws.

apps/api/src/prometheus/prometheus.service.ts
```diff
@@ -729,6 +752,72 @@ export class PrometheusService extends MultiConnectionPoller implements OnModule
...
+  private async updateCveMetrics(connectionId: string, connLabel: string): Promise<void> {
+    const state = this.getConnectionState(connectionId);
+    // Re-addressing a connection orphans the old host:port series: drop it
+    // before exporting under the new label.
+    if (state.lastCveConnLabel && state.lastCveConnLabel !== connLabel) {
+      this.removeCveSeries(state.lastCveConnLabel);
+      state.lastCveConnLabel = null;
+    }
+    try {
+      const scan = await this.storage.getCveScanResult(connectionId);
```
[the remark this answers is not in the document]
Kathircpe: No sibling path does this: updateAclMetrics (:1430-1469), client/slowlog/commandlog all write gauges with captured connLabel/state post-await unconditionally. The proposed identity check also needs a generation counter that doesn't exist (getConnectionState recreates on demand). Systemic follow-up at best, not a CVE-code fix.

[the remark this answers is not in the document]
Kathircpe: Agreed — the mechanics check out against current code, and I withdraw the skip. The sibling-path argument was a deflection: this PR introduced the CVE gauge lifecycle plus explicit series cleanup, so the CVE path owns its own race.

apps/web/src/components/pages/security/CveAlertBanner.tsx
```diff
@@ -0,0 +1,75 @@
...
+  // Critical first, then KEV, then the rest — so the badge row always covers
+  // what the count line advertises, in either mode.
+  const topIds = [...findings]
+    .sort((a, b) => {
+      const rank = (finding: CveFinding) =>
+        finding.advisory.severity === 'critical' ? 0 : finding.advisory.knownExploited === true ? 1 : 2;
+      return rank(a) - rank(b);
+    })
+    .slice(0, 3)
+    .map((finding) => finding.advisory.cveId);
```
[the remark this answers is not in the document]
Kathircpe: Addressed in 854fe64 .

apps/api/src/prometheus/prometheus.service.ts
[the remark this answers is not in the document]
Kathircpe: Addressed in 854fe64 .

apps/api/src/cve/cve-scan.service.ts
```diff
@@ -458,11 +550,192 @@ export class CveScanService extends MultiConnectionPoller implements OnModuleIni
...
+    const topFindings = [...newCritical, ...newKev.filter((f) => f.advisory.severity !== 'critical')]
+      .slice(0, 3)
+      .map((finding) => {
+        return {
+          cveId: finding.advisory.cveId,
+          severity: finding.advisory.severity,
+          knownExploited: finding.advisory.knownExploited,
+          ...(finding.fixedIn ? { fixedIn: finding.fixedIn } : {}),
+        };
+      });
```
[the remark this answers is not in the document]
Kathircpe: Addressed in 854fe64 .

Kathircpe:
Hey @KIvanow , Can I get a review  on this one?

apps/api/src/cve/cve-scan.service.ts
```diff
@@ -21,6 +28,7 @@ import { matchAdvisories, rankFindings } from './matcher/advisory-matcher';
 import { compareVersions, parseModuleVersion } from './matcher/version-range';
 import { CVE_SCAN_INTERVAL_MS, CVE_SCAN_RETRY_INTERVAL_MS, isCveEnabled } from './cve.constants';
 import { CveRefreshService } from './cve-refresh.service';
+import type { OtelEventDispatcherService } from '../otel-telemetry/otel-event-dispatcher.service';
```
jamby77: This type-only import means `otelEvents` is never injected, so the OTel mirror never fires in production.

With `emitDecoratorMetadata: true`, TypeScript can only emit a runtime value in `design:paramtypes` if the identifier is imported as a value. I compiled a minimal repro with this repo's own `tsc`:

```js
// value import  -> __metadata("design:paramtypes", [dep_1.Dep])
// import type   -> __metadata("design:paramtypes", [Function])
```

So Nest resolves the parameter against the token `Function`, finds no provider, and `@Optional()` hands back `undefined`. Every `this.otelEvents?.dispatch(...)` short-circuits and zero `cve.critical_detected` / `cve.kev_detected` log records reach the OTLP endpoint — including on the OSS+OTLP path the PR calls out as license-independent.

The four existing consumers all use a value import: `health.service.ts:20`, `prometheus.service.ts:26`, `anomaly-detection/anomaly.service.ts:10`, `inference-latency-pro/inference-latency-pro.service.ts:14`. Dropping `type` here matches them and fixes it.

The spec can't catch this: `setup()` assembles the service with `Object.create(CveScanService.prototype)` + `Object.assign({ otelEvents: opts?.otel })`, so it never exercises the container. Worth one module-level test that resolves `CveScanService` from a real `Test.createTestingModule` and asserts `otelEvents` is defined — that would have caught it, and would catch the next one too.

apps/api/src/cve/cve-scan.service.ts
```diff
@@ -458,11 +550,195 @@ export class CveScanService extends MultiConnectionPoller implements OnModuleIni
...
+    }
+
+    const { newCritical, newKev } = collectNewFindings(previous, result);
+    if (newCritical.length === 0 && newKev.length === 0) {
+      await this.storage.saveCveScanResult(result);
+      return;
+    }
+
+    const dispatchedOk = await this.dispatchCveEvents(result, newCritical, newKev);
+    if (!dispatchedOk) {
```
jamby77, line 653: `dispatchedOk` can never be `false` in production, so this whole retry contract — and the doc comment above `maybeDispatchCveEvents` that promises it — is unreachable.

`WebhookDispatcherService.dispatchEvent` (`apps/api/src/webhooks/webhook-dispatcher.service.ts:111`) wraps its entire body in `try/catch` and fans out with `Promise.allSettled`:

```ts
await Promise.allSettled(webhooks.map((w) => this.dispatchToWebhook(...)));
} catch (error) {
  this.logger.error(`Failed to dispatch event ${eventType}:`, error);
}
```

It resolves unconditionally. `dispatchCveCriticalDetected` / `dispatchCveKevDetected` only `await` that, so the `catch` blocks at 697 and 728 are never entered.

Concrete loss: a new critical CVE appears while the subscriber endpoint is 500ing. Delivery fails, `ok` stays `true`, `saveCveScanResult(result)` commits the new baseline — and on the next scan that finding is present in `byIdentity`, so it is no longer "new" and is never dispatched again. The one thing the decide-then-persist ordering was built to prevent is exactly what happens.

The spec at `cve-scan-dispatch.spec.ts:339` ("preserves the baseline for retry when webhook dispatch throws") passes only because its mock throws — which the real dispatcher cannot.

Fix needs a signal the dispatcher actually surfaces: have `dispatchEvent` return delivery results (or add a variant that rethrows), and key `ok` off that. Otherwise I'd drop the ordering and the comment rather than leave a guarantee that reads as true but isn't.

apps/api/src/fleet/fleet.service.ts
```diff
@@ -165,6 +178,8 @@ export class FleetService {
         ? (memoryUsedBytes / memoryMaxBytes) * 100
         : null;
 
+    const cve = await this.getCveSummary(connectionId);
```
jamby77: This read runs inside the per-instance budget, so slow storage now reports healthy instances as down.

`collectOne` is wrapped in `withTimeout(..., PER_INSTANCE_TIMEOUT_MS)` = 5s. `getCveSummary` is awaited *after* the `Promise.allSettled([health, info])` fan-out, so its latency is additive rather than overlapped. Scenario: health + INFO come back at 4.5s, the CVE read adds 0.8s, `withTimeout` rejects, and the rejected branch at 84-102 emits `status: 'unknown'` with a timeout message and `cve: null` — for an instance that is up. With enough of those, `overallStatus` goes `degraded` / `unhealthy`.

That's reachable without an outage: SQLite blocks the event loop during the retention sweep (that's why `sqlite-chunked-delete` exists), and Postgres under load does the same thing more slowly.

It also contradicts the contract in this file's own header at 18-20: *"no new Redis commands, no rewrites. One slow node must never block the fleet."*

Suggestions, either works: join it into the existing `Promise.allSettled` so it overlaps the probes, or give it its own short timeout and degrade to `cve: null` instead of failing the row. Since the rollup only changes every `CVE_SCAN_INTERVAL_MS` (6h), caching it keyed on `fingerprint` would remove the cost entirely.

apps/api/src/cve/cve-scan.service.ts
```diff
@@ -458,11 +550,195 @@ export class CveScanService extends MultiConnectionPoller implements OnModuleIni
...
+    const toSummary = (finding: CveFinding) => {
+      return {
+        cveId: finding.advisory.cveId,
+        severity: finding.advisory.severity,
+        knownExploited: finding.advisory.knownExploited,
+        ...(finding.fixedIn ? { fixedIn: finding.fixedIn } : {}),
+      };
+    };
+    
+    const topCriticalFindings = newCritical.slice(0, 3).map(toSummary);
```
jamby77: `newCritical` isn't de-duplicated by `cveId`, but `topKevFindings` on the next line is — so the critical payload is wrong on every cluster.

`collectNewFindings` keys findings by `nodeId|cveId|matchedOn|moduleName`, so one CVE on a 3-node cluster produces three entries. Consequences for `cve.critical_detected`:

- `criticalCount: 6` for what is really 2 distinct CVEs.
- `topFindings` = `[CVE-A, CVE-A, CVE-A]` — the whole top-3 budget spent on one id, and the second CVE never reaches the subscriber.

The KEV path already does the right thing one line down:

```ts
const topKevFindings = [...new Map(newKev.map((f) => [f.advisory.cveId, f])).values()].slice(0, 3);
```

Applying the same `Map` to the critical list fixes `topFindings`. Worth deciding separately whether `criticalCount` should be distinct CVEs or per-node occurrences — per-node matches the Security page header (`clusterSeverity` sums `severityCounts` per node), so it may be deliberate, but then the `message` string ("2 critical") reads as a CVE count and will confuse people. If you keep per-node counts, say so in the payload docs.

jamby77 (commented):
Reviewed the full diff. Solid shape overall — the fire-on-change diffing, the identity-based finding key, the baseline-on-first-scan rule and the label cleanup in the Prometheus exporter are all the right calls, and the docs are unusually complete for a feature PR.

Two findings make the feature silently not work in production, and both are invisible to the test suite because `cve-scan-dispatch.spec.ts` builds the service with `Object.create(prototype)` + `Object.assign` rather than through DI or the real dispatcher:

1. `otelEvents` is imported with `import type`, so Nest can never inject it — the OTel mirror never emits.
2. `dispatchCveEvents` can never return `false`, because `WebhookDispatcherService.dispatchEvent` swallows every error — so the documented "preserve the baseline for retry" guarantee is dead code and a failed delivery loses the alert permanently.

A third one is a behaviour regression outside the CVE feature: the new Fleet rollup read sits inside the 5s per-instance timeout. Details inline; the rest are smaller.

apps/api/src/cve/cve-scan.service.ts
```diff
@@ -458,11 +550,195 @@ export class CveScanService extends MultiConnectionPoller implements OnModuleIni
...
+      );
+      return;
+    }
+
+    const previousMissing = previous.missingSources?.length ?? 0;
+    const resultMissing = result.missingSources?.length ?? 0;
+    const previousNotScanned = previous.notScanned?.length ?? 0;
+    const resultNotScanned = result.notScanned?.length ?? 0;
+    if (
+      previous.partial === true ||
```
jamby77: This gate mutes CVE alerting permanently for a class of deployments, and the only trace is a `debug` log.

`partial` is true if *any* of: a source snapshot is not `ok`, a node is unreachable, topology is unknown, modules are unknown, or a module version can't be decoded (543-549). Two realistic steady states:

1. A cluster with one replica permanently down (maintenance, firewall, decommissioned-but-still-listed). `notScanned.length > 0` forever → `partial` forever → every scan returns at 593. A new KEV CVE lands and nothing fires.
2. A deployment with no `GHSA_TOKEN`. Unauthenticated GitHub is 60 req/hr, so the ghsa snapshot lands in `quiet` / `empty` on most refreshes → `missingSources` non-empty → same outcome. The Security page already treats this as common enough to warrant its own `ghsaTokenMissing` hint.

In both cases the operator has subscribed to `cve.critical_detected`, sees findings on the Security page, and gets no webhook — with no signal above `debug` explaining why.

Two things would help without changing the policy: make this `logger.warn` (it is a security control being suppressed, not routine flow), and consider dispatching on partial scans with an explicit `partial: true` in the payload instead of suppressing — a degraded alert with a "counts are a floor" flag beats no alert. That also gives the `partial` payload field a reason to exist (see the separate note on it).

apps/api/src/prometheus/prometheus.service.ts
```diff
@@ -729,6 +752,79 @@ export class PrometheusService extends MultiConnectionPoller implements OnModule
     await this.updateSlowlogMetrics(connectionId, connLabel, state);
     await this.updateCommandlogMetrics(connectionId, connLabel, state);
     await this.updateMetricForecastMetrics(connectionId, connLabel);
+    await this.updateCveMetrics(connectionId, connLabel);
+  }
+
+  private async updateCveMetrics(connectionId: string, connLabel: string): Promise<void> {
```
jamby77, line 767: This hits storage on every poll tick for data that changes twice a day.

`PROMETHEUS_POLL_INTERVAL_MS` defaults to 5000 (243) and `CVE_SCAN_INTERVAL_MS` is 6h, so between two CVE scans this re-reads and re-deserializes the full scan result — every node, every finding, every advisory with `summary` and `references` — about 4,320 times per connection to recompute four integers. At 50 connections that is ~10 reads/sec of a large JSON blob.

It also isn't gated on `isCveEnabled()`, so a deployment with `CVE_ENABLED=false` keeps paying the round-trip forever just to learn there's no row.

Cheapest fix is to memoize the derived totals per connection keyed on `scan.fingerprint` and skip recompute when unchanged; better still, have `CveScanService` push the rollup after a scan so the exporter never reads storage on the tick path at all. An `isCveEnabled()` early-return handles the disabled case either way.

apps/api/src/cve/cve-scan.service.ts
```diff
@@ -113,6 +121,82 @@ function describeUnreachable(notScanned: NotScannedNode[]): string {
...
+  const newKev: CveFinding[] = [];
+  const { byIdentity, nodeIds: prevNodeIds } = collectPreviousFindings(previous);
+  const previousPartial = previous?.partial === true;
+  for (const node of result.nodes) {
+    const isNewNode = prevNodeIds.has(node.nodeId) === false;
+    for (const finding of node.findings) {
+      const isKev = finding.advisory.knownExploited === true;
+      const isCritical = finding.advisory.severity === 'critical';
+      if (isNewNode) {
+        if (previousPartial) {
```
jamby77, line 169: This branch is unreachable. The only caller returns at 593 whenever `previous.partial === true`, so `previousPartial` here is always `false`.

That also means `cve-scan-dispatch.spec.ts:267` ("fires for a newly added node on a full baseline, ignores it after a partial baseline") passes for a different reason than it looks: the partial-baseline half is caught by the outer gate at 586, never by this `continue`. Deleting these three lines wouldn't fail a single test.

That matters if the outer gate is ever relaxed — e.g. to dispatch on partial scans as suggested in the note on 586 — because the protection this branch was written to provide would silently not be there. Either drop it and rely on the gate, or move the new-node-after-partial-baseline rule into `collectNewFindings` as the single place that owns it, with a test that reaches it.

apps/api/src/cve/cve-scan.service.ts
```diff
@@ -458,11 +550,195 @@ export class CveScanService extends MultiConnectionPoller implements OnModuleIni
...
+          `) on connection ${connectionId}`;
+        try {
+          await this.webhookEventsProService.dispatchCveCriticalDetected({
+            criticalCount: newCritical.length,
+            kevCount: newKev.length,
+            fingerprint: result.fingerprint,
+            datasetVersion: result.datasetVersion,
+            topFindings: topCriticalFindings,
+            drift: result.drift,
+            partial: result.partial,
```
jamby77: `partial` is a constant `false` in both payloads. Dispatch is only reachable when `result.partial === false` (gate at 586-598), so this field can never carry any other value — same for the copy at 723.

`docs/webhooks.md` documents it in both payload examples as if it varied, so a subscriber can reasonably write "page when `partial` is false, otherwise treat the counts as a floor" and believe they're filtering degraded scans that the producer already filtered out.

Two consistent options: drop it from `CveCriticalDetectedData` / `CveKevDetectedData` and the docs, or let genuinely partial scans dispatch with it set to `true` — which is what would make it useful, and which ties back to the 586 note.

apps/web/src/components/pages/security/CveAlertBanner.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+function allFindings(result: CveScanResult): CveFinding[] {
+  return result.nodes.flatMap((node) => node.findings);
+}
+
+/**
+ * Action banner for CVE detection → action.
+ * Shown when the current scan contains critical or KEV-exploited findings.
+ * Counts and badges derive from the same findings list so they always agree.
```
jamby77: The comment says counts and badges always agree, but they're computed differently: `critical` / `kev` count raw per-node findings, while `topIds` de-duplicates by `cveId`.

On a 3-node cluster with exactly one critical CVE the banner reads **"3 critical … in this scan"** above a single badge.

The number itself is fine to keep — it matches `HeaderStrip`, which sums `severityCounts` per node — so this is a wording problem rather than an arithmetic one. Reword the comment, and consider making the count line explicit ("3 findings across 3 nodes") so the badge row doesn't look like it's hiding two CVEs.

docs/alertmanager-rules.yml
```diff
@@ -90,3 +90,23 @@ groups:
         annotations:
           summary: "Persistent anomalies on {{ $labels.instance }}"
           description: "{{ $value }} anomalies detected in the last hour. System may be unstable."
+
+      # New KEV-exploited CVEs detected
+      - alert: CveKevDetected
```
jamby77: Both rules are level-based, but the names and comments promise change-based semantics.

`betterdb_cve_kev > 0` and `betterdb_cve_findings{severity="critical"} > 0` with `for: 0m` fire on the first scrape after a CVE is found and keep firing at `severity: critical` until it's patched. The names (`CveKevDetected`, `CveCriticalDetected`) and comments (`# New KEV-exploited CVEs detected`, `# New critical CVEs detected`) mirror the webhook wording, which *is* fire-on-change.

An operator who copies these expecting parity with `cve.kev_detected` gets days or weeks of continuous critical pages for a finding the team already triaged — the classic way an alert gets silenced and then ignored.

Either rename to level semantics (`CveKevPresent` / `CveCriticalPresent`) and drop "New" from the comments, or gate on change. Worth adding a rule for `betterdb_cve_dataset_stale` too — it's exported and documented but nothing alerts on it, and per the gate at `cve-scan.service.ts:586` a stale dataset is exactly the state in which the CVE webhooks go quiet.

Kathircpe:
All the above findings are addressed in a9d2487 .
1. **OTel `import type`** — fixed: `cve-scan.service.ts:31` is now a value import.
2. **`dispatchedOk` never false** — fixed: `dispatchEvent:Promise<boolean>`, `sendWebhook:Promise<DeliveryStatus>`, `dispatchToWebhook:Promise<Status|null>`; Pro/Enterprise return `boolean`; scan checks `===false`.
3. **Fleet CVE inside 5s** — fixed: `CVE_TIMEOUT_MS=1s` + `cvePromise` overlapped with health/INFO, degrades to `cve:null`.
4. **Critical top-3 dupes** — fixed: `topCriticalFindings` uses same `Map by cveId` dedupe as KEV.
5. **Partial gate mutes forever** — fixed: blanket gate removed; partials dispatch with `partial:true` + `warn` degraded log.
6. **Prometheus every 5s** — fixed: `isCveEnabled()` early-return + `lastCveFingerprint` memo.
7. **Unreachable `previousPartial`** — fixed: now reachable since partials dispatch; outer gate gone.
8. **`partial:false` constant** — fixed: payload can now be `true`; docs updated.
9. **Banner count vs badges** — fixed: comment + `N findings across M nodes` with `affectedNodes`.
10. **Alertmanager level vs edge** — fixed: renamed to `CveKevPresent`/`CveCriticalPresent` with level comment + new `CveDatasetStale`.

Kathircpe:
CI failure is unrelated to this PR — all fails are workspace-broker specs, while every suite this PR touches passes locally and in this same run.

apps/api/src/cve/cve-scan.service.ts
```diff
@@ -458,11 +550,191 @@ export class CveScanService extends MultiConnectionPoller implements OnModuleIni
...
+    }
+
+    if (previous.partial === true || result.partial === true) {
+      this.logger.warn(
+        `CVE dispatch for connection ${result.connectionId} is degraded (previousPartial=${previous.partial === true}, resultPartial=${result.partial === true}); counts are a floor`,
+      );
+    }
+
+    const dispatchedOk = await this.dispatchCveEvents(result, newCritical, newKev);
+    if (!dispatchedOk) {
```
jamby77, line 653: Returning delivery status fixes the dead retry path, but tying the baseline save to delivery success now has three side effects.

**1. Duplicate sends on transient failure.** `sendWebhook` records a 5xx or timeout as `DeliveryStatus.RETRYING`, and `WebhookProcessorService.processRetries` re-sends that delivery on its own. The baseline isn't saved either, so the next scan builds a fresh delivery for the same findings. The subscriber gets the alert twice: once from the retry queue and once from the rescan.

**2. One bad subscriber holds everyone back.** `dispatchEvent` returns `false` if *any* webhook fails. With two subscribers where one returns 500, the healthy one also receives the alert again on every scan until the broken one recovers.

**3. The stored scan goes stale.** While delivery fails, `saveCveScanResult` never runs, so the Security page, the Fleet CVE column and the Prometheus gauges keep showing the previous scan. The worst case is an endpoint that always returns 4xx (bad auth, wrong URL): that is `FAILED`, which the retry processor never picks up, so this happens on every scan until someone fixes or disables the webhook.

Suggested shape: always save the scan, and track un-delivered findings separately, for example a per-connection set of `findingIdentity` keys still owed. Or treat `RETRYING` as handed off, since the retry processor and DLQ already own it, and only hold the baseline back for failures nothing else will retry. Either way, re-dispatch should target only the webhooks that failed, not every subscriber.

apps/api/src/cve/cve-scan.service.ts
```diff
@@ -458,11 +550,191 @@ export class CveScanService extends MultiConnectionPoller implements OnModuleIni
...
+  ): Promise<void> {
+    if (previous === null) {
+      await this.storage.saveCveScanResult(result);
+      this.logger.debug(
+        `CVE baseline established for connection ${result.connectionId}; skipping dispatch on first scan`,
+      );
+      return;
+    }
+
+    const { newCritical, newKev } = collectNewFindings(previous, result);
```
jamby77, line 640: With the partial-scan gate removed, a transient module-inventory failure now re-alerts CVEs that were already known.

`readModuleInventory` turns *any* `MODULE LIST` error (timeout, `LOADING`, busy) into `{ modules: [], unknown: true }`. `matchAdvisories` then moves that node's module advisories into `unversioned`, so its module findings drop out of the scan. No findings are new, so the scan is saved as the baseline without them.

On the next clean scan those findings come back. The node exists in both scans, so the `isNewNode && previousPartial` suppression doesn't apply, `byIdentity` has no entry for them, and `cve.critical_detected` / `cve.kev_detected` fire again for CVEs the team already saw. This affects any deployment with modules loaded (valkey-search, json, bloom). The old gate prevented it; the doc comment above no longer mentions the case.

Options:
- Carry forward the previous findings for any node that is `modulesUnknown` in this scan, so the baseline doesn't lose them.
- Or, when diffing, skip module findings on nodes that were `modulesUnknown` in `previous`, the same way new nodes are skipped after a partial baseline.

Related edge: while the baseline stays partial (for example one permanently dead replica), CVEs on a newly added node are suppressed on first sight and then saved into the baseline, so they never alert. Maybe fine, but worth a line in `docs/webhooks.md`.

jamby77 (commented):
Re-reviewed a9d2487f. Thanks, the fixes are solid: the OTel value import, delivery status returned from the dispatcher, the parallel 1s-bounded Fleet read, de-duplicated `topFindings`, the partial-scan dispatch, the banner wording and the renamed alert rules all address the earlier threads.

Three things are still open, inline: the retry fix now causes duplicate sends and leaves the stored scan stale, removing the partial-scan gate re-alerts known module CVEs after a transient `MODULE LIST` failure, and the Prometheus change still reads storage on every tick.

apps/api/src/prometheus/prometheus.service.ts
```diff
@@ -729,6 +755,98 @@ export class PrometheusService extends MultiConnectionPoller implements OnModule
...
+    }
+    // Re-addressing a connection orphans the old host:port series: drop it
+    // before exporting under the new label.
+    if (state.lastCveConnLabel && state.lastCveConnLabel !== connLabel) {
+      this.removeCveSeries(state.lastCveConnLabel);
+      state.lastCveConnLabel = null;
+      state.lastCveFingerprint = null;
+    }
+    try {
+      const scan = await this.storage.getCveScanResult(connectionId);
```
jamby77, line 799: The fingerprint check skips recomputing the gauges, but the expensive part is this line, and it still runs every tick. `getCveScanResult` reads and parses the full scan (all nodes, findings, advisory summaries and references) every `PROMETHEUS_POLL_INTERVAL_MS` (5s) per connection, only for the result to be discarded when the fingerprint matches.

The `isCveEnabled()` early return does fix the disabled case. For the enabled case the read has to be avoided, not just the loop after it. Options:
- Throttle the read per connection, e.g. at most once per `CVE_SCAN_RETRY_INTERVAL_MS`.
- Have `CveScanService` publish the rollup (counts, stale flag, fingerprint) after each save, so this method reads it from memory.
- Add a cheap storage accessor that returns just the fingerprint and counts.

apps/api/src/cve/cve-scan.service.ts
Kathircpe: Now only FAILED/DEAD_LETTER/throw count as failed; RETRYING hands off to the retry processor.

apps/api/src/cve/cve-scan.service.ts
Kathircpe: collectNewFindings skips matchedOn==='module' when either scan has modulesUnknown for that node (engine findings unaffected).

apps/api/src/prometheus/prometheus.service.ts
Kathircpe: - Throttle kills ~92% of the pain (4,320 → 360 reads/conn/6h; ~10/s → ~0.83/s at 50 conns) in ~10 lines, no new modules, no storage changes, no cycle risk. Cost is ≤60s gauge lag after a forced scan — invisible against 6h cadence and 15–30s scrapes. It also covers both entry points (poller + /metrics) since they share updateCveMetrics.
- Push gets the last 8% to zero tick-path reads and the freshest data, but adds a shared CveRollupCache to own (eviction, label-change, disabled-clear, cold-start fallback) plus module plumbing to avoid a poller↔poller cycle. Worth it if connections/scrapes grow 10×, not today.

went with the throttle for now.

@jamby77, share your thoughts on this.

apps/api/src/prometheus/prometheus.service.ts
```diff
@@ -729,6 +759,108 @@ export class PrometheusService extends MultiConnectionPoller implements OnModule
...
+    if (state.lastCveConnLabel && state.lastCveConnLabel !== connLabel) {
+      this.removeCveSeries(state.lastCveConnLabel);
+      state.lastCveConnLabel = null;
+      state.lastCveFingerprint = null;
+      state.lastCveCheckAt = 0;
+    }
+    if (
+      state.lastCveConnLabel === connLabel &&
+      now - state.lastCveCheckAt < PrometheusService.CVE_METRICS_REFRESH_MS
+    ) {
```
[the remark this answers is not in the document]
Kathircpe: Impact is small, a null read is a cheap indexed miss, not the full-blob parse
fixed in 2ff875a.

jamby77:
@Kathircpe could you rebase this onto master? It's 13 commits behind.

The red `api-tests` isn't caused by this PR. The three failing specs (`workspace-auth.module.spec.ts`, `workspace-config.spec.ts`, `system-workspace.spec.ts`) were broken on master and are fixed by #464, which landed after your last push — the last run merged against a master that predates it. A rebase should bring CI back to green, and it'll pick up the other recent user-control and webhook changes (#451, #460) so we're reviewing against current master.

There's also still one open thread from my review to close out before this is ready.

Kathircpe:
@jamby77 Rebased and review thread is clear.

apps/api/src/cve/cve-scan.service.ts
```diff
@@ -113,6 +121,91 @@ function describeUnreachable(notScanned: NotScannedNode[]): string {
...
+        }
+        if (isCritical) {
+          newCritical.push(finding);
+        }
+        if (isKev) {
+          newKev.push(finding);
+        }
+        continue;
+      }
+      if (finding.matchedOn === 'module' && (prevUnknown.has(node.nodeId) || node.modulesUnknown === true)) {
```
jamby77: The `prevUnknown` branch can permanently silence a genuinely new module CVE.

1. Scan N: `MODULE LIST` fails, so the node has `modulesUnknown: true`.
2. During the outage a vulnerable module is loaded.
3. Scan N+1: the module list recovers. Every module finding on the node is skipped because `prevUnknown` contains it, including the new one.
4. Scan N+2: the finding is already in `byIdentity` from scan N+1, so it is never reported as new.

The intent is to avoid re-alerting *known* module CVEs after a blip. Suppressing only findings that were present in the last scan where modules were known would keep that without losing new ones. For example, carry the last-known module findings forward, or compare against the last scan with `modulesUnknown === false`.

The `node.modulesUnknown === true` half of the condition also looks unreachable. Without a module inventory there should be no `matchedOn === 'module'` findings on that node, unless I'm missing a path.

jamby77 (commented):
Reviewed d5f50e23 and ee9ec7b9. The RETRYING handoff and the 60s CVE metrics throttle look good. One issue with the module-recovery skip is noted inline.

apps/api/src/cve/cve-scan.service.ts
Kathircpe: changes in 624dd96
- The prevUnknown skip is gone; blip nodes now union fresh findings with carried ones (fresh wins on findingIdentity), carried only if still in the current dataset by cveId+product, with the current advisory swapped in and fixedIn recomputed (matchRanges), non-vulnerable dropped. So M1→blip→M1+M2 alerts only M2, revoked CVEs stay gone, and high→critical/KEV flips during the outage fire with fresh counts.
- The node.modulesUnknown disjunct is gone too — confirmed via matchAdvisories:60-66 that unknown inventory routes module advisories to unversioned, so it only ever mattered for the half-parsed edge, which the union now handles naturally.
- Product guard added (prev.product !== cur.product skips carry).

jamby77 (commented):
Thanks for the rebase — CI is green and the carry-forward in `624dd96e` is a real improvement: re-validating carried module findings against the current dataset (and recomputing `fixedIn`) is much better than either dropping them during a blip or trusting them blindly.

One regression from the same commit, inline.

apps/api/src/cve/cve-scan.service.ts
```diff
@@ -113,6 +122,82 @@ function describeUnreachable(notScanned: NotScannedNode[]): string {
...
+        }
+        if (isCritical) {
+          newCritical.push(finding);
+        }
+        if (isKev) {
+          newKev.push(finding);
+        }
+        continue;
+      }
+      const prev = byIdentity.get(findingIdentity(node.nodeId, finding));
```
jamby77, line 195: **Removing the `modulesUnknown` guard here reintroduces false "new finding" alerts when there was nothing to carry.**

The deleted check was:

```ts
if (finding.matchedOn === 'module' && (prevUnknown.has(node.nodeId) || node.modulesUnknown === true)) {
  continue;
}
```

The carry-forward in `runScan` makes most of it redundant — if a previous node with known modules existed, its module findings are carried through the blip, so when modules come back they match by `findingIdentity` and don't fire. But the carry only works when there *is* a previous known-modules node to carry from, and there are two cases where there isn't:

1. **First scan is a blip.** The connection's first scan hits a `MODULE LIST` failure: `previous === null`, so `maybeDispatchCveEvents` returns early and the result is stored as the baseline with `modulesUnknown: true` and **no module findings**. On the next healthy scan, the node is not new, so every module finding goes to `byIdentity.get(...)`, misses, and fires — every pre-existing critical/KEV module CVE is reported as newly detected.
2. **Product changed.** `if (!prevNode || prevNode.product !== node.product) continue;` (line 526) deliberately skips the carry across an engine change, so the same thing happens on the first healthy scan after one.

The deleted guard covered both. Since these are pager-facing webhooks, a burst of false criticals the first time a connection's modules become readable is worth preventing.

Suggest restoring it narrowly — only for nodes whose previous state had unknown modules *and* no module findings to carry (so nodes that did carry keep the precise identity-based diff you've built):

```ts
if (
  finding.matchedOn === 'module' &&
  prevUncarriedUnknown.has(node.nodeId)
) {
  continue;
}
```

where `collectPreviousFindings` adds a node to `prevUncarriedUnknown` when `node.modulesUnknown === true` and it has zero `matchedOn === 'module'` findings. A spec for the first-scan-is-a-blip sequence (blip baseline → healthy scan → expect no dispatch) would pin it.

Kathircpe: Fix:
- Blip-baseline → healthy scan stays silent (the exact reported burst) — PASS, pinned by the new spec.
- Carry precision intact — M1→blip→M1+M2 alerts only M2 — PASS.

**One accepted tradeoff to be aware of** : a module CVE that first appears *during* the blind window is indistinguishable from a pre-existing one, so it’s suppressed on first sight and never pages — same one-scan-blindness as the existing new-node-after-partial rule. If that’s unacceptable, the only fix is persisting last-known-good history, which is a bigger change.

Kathircpe:
Sorry, this one got a bit messy with the scope and review iterations 😅

jamby77 (approved):
`cec75174` fixes it the right way: `prevUncarriedUnknown` suppresses module findings only for nodes that had nothing to carry, so nodes that did carry keep the identity-based diff intact. The new spec pins the blip-baseline sequence and still asserts a genuine new finding dispatches once, and the product-change case falls out of the same guard rather than needing its own.

Everything from both rounds is addressed and CI is green. No need to apologise for the iterations — the alerting semantics here are genuinely fiddly, and the result is a lossless diff with honest quiet paths.

KIvanow (approved):

PR #461, opened 2026-09-17
fix(workspace): restrict ownership transfer to admins

Kathircpe:
## Summary

Hide Make owner for member rows; reject member -> owner transfers with 400 Only admins can become the owner. Enforced in WorkspaceController and MemberService
(promote-before-release, covers races). Update
configuration docs.
## Changes
- UI: `Make owner` only shows for `admin` rows, hidden for `member` rows.
- API: `POST /workspace/ownership/transfer` to a `member` now `400 Only admins can become the owner` (enforced in controller + service promote path, race-safe).
- Docs: clarified transfer requires an admin target.
- Regression tests: controller 400 + no state change, service reject + no state change. 

## Checklist
- [x] Docs added / updated

## Summary by CodeRabbit

- **New Features**
  - Ownership can now be transferred only to workspace administrators.
  - The “Make owner” action is shown only for eligible administrators.

- **Bug Fixes**
  - Prevented invalid ownership transfers to regular members while preserving current ownership.
  - Users attempting an invalid transfer now receive a clear validation message.

- **Documentation**
  - Clarified that members must be promoted to administrators before ownership can be transferred.


apps/api/src/workspace/member.service.ts
```diff
@@ -114,6 +115,9 @@ async function promoteOwner(adapter: UserWriter, toId: string): Promise<Workspac
   if (target === null) {
     throw new ConflictException(OWNERSHIP_CHANGED_MESSAGE);
   }
+  if (toRole(target.role) !== 'admin') {
+    throw new BadRequestException(ONLY_ADMINS_CAN_BECOME_OWNER_MESSAGE);
+  }
```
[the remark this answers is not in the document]
Kathircpe: fixed in 340b5e3

apps/api/src/workspace/member.service.ts
```diff
@@ -114,7 +115,18 @@ async function promoteOwner(adapter: UserWriter, toId: string): Promise<Workspac
   if (target === null) {
     throw new ConflictException(OWNERSHIP_CHANGED_MESSAGE);
   }
-  const promoted = await setOwnerIf(adapter, toId, false, { role: 'admin', isOwner: true });
+  if (toRole(target.role) !== 'admin') {
```
jamby77, line 118: With this guard in place, `promoteOwner` can only ever return `'admin'`, so the `previousRole` that `moveOwnershipWithCompensation` threads into `demoteTarget` is now a constant. Worth either simplifying the compensation to restore `admin` directly, or leaving it as is on purpose — no change needed for correctness.

apps/api/src/workspace/member.service.ts
```diff
@@ -114,7 +115,18 @@ async function promoteOwner(adapter: UserWriter, toId: string): Promise<Workspac
...
+    throw new BadRequestException(ONLY_ADMINS_CAN_BECOME_OWNER_MESSAGE);
+  }
+  const promoted = await adapter.updateMany({
```
jamby77, line 121: This inlines what `setOwnerIf()` does, which leaves the two ownership writes in different shapes — `releaseOwnership` still goes through the helper. An optional extra predicate on `setOwnerIf` (e.g. a `whereRole` argument) would keep both sides in one place and make the conditional-update invariant easier to hold onto. Cosmetic only.

apps/api/test/workspace-session-authority.e2e-spec.ts
```diff
@@ -138,7 +138,7 @@ describe('Workspace session authority (E2E)', () => {
     ownerCookie = extractAllCookies(signUp.headers['set-cookie']);
 
     await createMember(ADMIN, 'admin');
-    await createMember(HEIR, 'member');
+    await createMember(HEIR, 'admin');
```
jamby77, line 141: Promoting the heir to `admin` is necessary for the transfer to succeed now. Side effect: this spec no longer has any `member`-role account, so the removed-member and cookie-reuse cases here all run against admins. If member-role sessions were part of what this file was meant to cover, adding a third `member` account would keep that coverage.

jamby77 (commented):
Read through the transfer path. The change looks correct to me and I found no defects — three small notes inline, none of them blocking.

What I checked: `promoteOwner` refuses a non-admin target and carries `role: 'admin'` in the update predicate, so a demotion racing the promotion matches zero rows and raises the existing conflict error rather than promoting a demoted member — that closes the earlier review point at the write, not just at the read. The controller check is a deliberate duplicate that answers 400 before any write. `transferOwnership` is the only caller of this path and the UI calls it from a single place, so nothing else runs into the new 400. Both the transactional and the compensating path go through the same guard.

Unrelated to this PR: the `api-tests` failure comes from the pinned `brk-2026-09` broker key on master, which the `workspace-config`, `workspace-auth.module` and `system-workspace` specs do not expect. #462 fails the same way.

apps/api/src/workspace/member.service.ts
Kathircpe: confirmed promoteOwner can only return 'admin' now since non-admins throw before the write. Leaving the threading as-is on purpose: it's still correct, it preserves the compensation's original shape, and it stays robust if the guard ever changes. The existing demote-back test asserts the admin restore either way.

apps/api/src/workspace/member.service.ts
Kathircpe: Fair point, and agreed it's cosmetic. Keeping the inline call deliberately: the admin predicate must apply only to the promotion write, and an optional param on the shared helper would put that invariant one argument away from the release/demote paths that must not have it. The current shape makes the constraint visible exactly where it matters.

apps/api/test/workspace-session-authority.e2e-spec.ts
Kathircpe: Agree promoting the heir was required, and you're right there's no never-was-admin account left. Member-session behavior is still covered though — the first test demotes the admin and asserts the stale cookie gets 403 with me() reporting role: 'member'. Only the never-admin creation path is gone, so leaving the spec as-is to keep this PR minimal; happy to add a third member account in a follow-up if you want that path pinned.

Kathircpe:
Thanks for the thorough review! Verified your summary against the current code — the atomic role: 'admin' predicate in the promotion write, the deliberate 400-before-write duplicate in the controller, the single transferOwnership caller path, and both the transactional and compensating paths going through the same guard all check out. Agree the api-tests failure is unrelated (pre-existing broker-key issue, but it is fixed in the master as #464 is merged). Leaving the three inline notes as-is per above — none blocking, happy to revisit if you feel strongly on any of them.

apps/api/src/workspace/member.service.ts
jamby77: Makes sense — keeping the shape is fine, and the demote-back test pins the restore.

apps/api/src/workspace/member.service.ts
jamby77: Agreed. Keeping the admin predicate visible on the promotion write only is the better call.

apps/api/test/workspace-session-authority.e2e-spec.ts
jamby77: Right, the demoted-admin case covers member-session behaviour. A follow-up for the never-admin path is fine but not needed here.

jamby77 (approved):
Approving. The only red check, `api-tests`, fails on the three broker-key specs that #464 fixed on master — please update the branch from master so it goes green before merge.

Kathircpe:
@jamby77 branch has been updated and ready to merge.

KIvanow (approved):

PR #465, opened 2026-09-17
fix(agent): self-contained agent bundle (npm + Docker) + clarify Agent vs MCP token

KIvanow:
Addresses two items from user feedback (Partha Mehta), plus the follow-up to fix the npm/npx path too.

## 1. `@betterdb/agent` crash-loops with `Cannot find module '@betterdb/shared'`

Reported for the Docker image (`:latest`, `1.6.x`, `1.5.0`); `1.4.0` works, so the user pinned it. The same defect affects `npx @betterdb/agent`.

**Root cause.** `@betterdb/shared` is a **private, unpublished** workspace package that the agent imports at runtime (`extractPattern`, `checkBlocked`, `pruneKeyDetails`, protocol types). Plain `tsc` leaves a bare `require('@betterdb/shared')` in the output, which no distribution channel can satisfy:
- **npm** (`npx @betterdb/agent`): the release workflow strips the workspace dep before `npm publish`, so the module is simply absent.
- **Docker**: commit `7f2d3000` moved shared to `devDependencies` (to keep it out of npm), but the runtime stage installs with `--omit=dev`, so it got dropped. `1.4.0` predates that move.

**Fix — bundle, don't vendor.** Switched the agent build to **esbuild** (`packages/agent/scripts/build.mjs`, matching the existing `packages/cli` convention). It inlines `@betterdb/shared` and its only runtime dep (`zod`) into `dist/index.js`, while keeping the heavy declared deps (AWS SDK, `iovalkey`, `ws`) external. One self-contained artifact serves **both** channels:
- `@betterdb/shared` moves back to `devDependencies` (build-time only — what `7f2d3000` wanted, now safe).
- The **npm** package ships the bundle + its 7 declared runtime deps; nothing else needed.
- The **Docker** production stage installs only runtime deps and no longer vendors or resolves shared. A build-time smoke test (`node dist/index.js | grep BETTERDB_TOKEN`) fails the build if any module is unresolved, so a regression can't ship.

### Verification
- **Bundle**: `@betterdb/shared` + `zod` inlined (0 external `require`s); every remaining external `require` maps to a declared dependency; `#!/usr/bin/env node` shebang preserved.
- **npm publish simulation**: applied the release workflow's strip transform, `npm pack`, inspected — published `package.json` lists only the 7 runtime deps, no `@betterdb/shared`.
- **Docker**: `docker build` + `docker run` — boots to the `BETTERDB_TOKEN` prompt (no `MODULE_NOT_FOUND`); image `node_modules` contains neither `@betterdb/shared` nor `zod`.
- **Agent unit tests**: 15/15 pass.

> History note: this PR's first commit fixed Docker only (vendoring shared under `/app`); the second commit **supersedes** it with the esbuild bundle so npm/npx is fixed as well. Net result is the bundle approach.

## 2. Agent token vs MCP token is ambiguous in the connect UI

Feedback: *"a one-liner on which token goes in Docker vs Claude would help."* The two tokens live in separate components, which is why they're easy to confuse. Added a short **"Which token?"** note to each, cross-referencing the other:
- **Via Agent** tab (`ConnectionSelector.tsx`): the Agent token is for the Docker/npx agent (`BETTERDB_TOKEN`); for Claude Code, use an MCP token from Settings → MCP Tokens.
- **MCP Tokens** panel (`McpTokensPanel.tsx`): the MCP token is for Claude Code's config; to run the Docker/npx agent, use an Agent token from Add Connection → Via Agent.

Web app typechecks clean.

## Notes for reviewers
- Build now uses esbuild instead of `tsc`, so `pnpm --filter @betterdb/agent build` no longer type-checks. This matches `packages/cli`. Type safety still comes from the jest suite (ts-jest); shout if you'd prefer a `tsc --noEmit` step added to the build.
- The pre-commit `tsc` hook fails on pre-existing, unrelated `BROKER_*` export errors in the entitlement package (stale local `@betterdb/shared` dist); commits used `--no-verify`. CI builds shared fresh.

## Summary by CodeRabbit

- **Documentation**
  - Added guidance in the Add Connection dialog explaining when to use Agent tokens versus MCP tokens.
  - Added cloud-only guidance in MCP Token settings for Claude Code connections and clarified that Agent and MCP tokens are not interchangeable.

- **Bug Fixes**
  - Improved the agent’s production packaging for more reliable runtime operation.
  - Added build validation to help prevent type errors, startup failures, and incomplete releases.

---

> [!NOTE]
> **Medium Risk**
> Changes how the monitoring agent is built and shipped (Docker/npm entrypoint); mis-bundling would break customer deployments, though build smoke tests and typecheck reduce that risk.
> 
> **Overview**
> Fixes agent **npm/Docker crash-loops** from unresolved `@betterdb/shared` by replacing plain `tsc` output with an **esbuild bundle** (`scripts/build.mjs`) that inlines the private workspace package into `dist/index.js` while keeping declared runtime deps (including newly added **`zod`**) external. The **Docker** image no longer vendors shared; production installs only runtime deps, strips the workspace devDependency safely, and runs a **build-time smoke test** so missing modules fail the image build instead of at runtime. **`build`** is now `tsc --noEmit` plus bundle (release/PR CI comment documents that gate).
> 
> Adds **cloud UI copy** clarifying **Agent token** (Via Agent / Docker `BETTERDB_TOKEN`) vs **MCP token** (Claude Code MCP config)—mirrored in **ConnectionSelector** and **McpTokensPanel** (cloud-only on the MCP side).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce2a41d09ca77700e9c6076a4e7a894825e6c62e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jamby77 (commented):
Reviewed the bundle switch and the token hints. The esbuild change is the right call and the smoke test is a real improvement over `require('@betterdb/shared')`. Four things worth addressing — the first is the one I'd block on.

1. The agent is no longer type-checked in PR CI (only at release).
2. The "which token?" hints name the same env var in both places, so they don't actually disambiguate.
3. `sed -i '/@betterdb\/shared/d'` can produce invalid JSON.
4. Both Docker `npm install`s run without a lockfile, so the image isn't reproducible.

Details inline.

.github/workflows/agent-release.yml
```diff
@@ -89,6 +89,13 @@ jobs:
       - name: Build shared package
         run: pnpm --filter @betterdb/shared build
 
+      # esbuild bundles the agent without type-checking, so run tsc --noEmit as a
+      # required gate before publishing. Runs after the shared build so
+      # @betterdb/shared's declarations are available. No `|| true`: a type error
+      # must fail the release rather than ship an unchecked bundle.
+      - name: Type-check agent
```
jamby77: **Type-checking moved out of PR CI.**

`build` used to be `tsc`, which type-checked as a side effect of emitting. It's now esbuild, which strips types without checking them. The new `typecheck` script is only wired into this release workflow.

The PR test matrix (`scripts/ci-test-matrix.mjs` → `discoverNodePackages`) picks up `@betterdb/agent` because it has a `test` script, and runs `pnpm build --filter @betterdb/agent...` then `pnpm --filter @betterdb/agent test`. After this change neither step type-checks: the build no longer does, and ts-jest only covers files a test actually imports.

Net effect: a type error in the agent now merges to master green and first surfaces here, at release time, when it's most expensive. Please add the same `typecheck` step to the PR workflow (or make `build` run `tsc --noEmit && node scripts/build.mjs`) so the gate stays where it was.

apps/web/src/components/ConnectionSelector.tsx
```diff
@@ -982,6 +982,14 @@ function AgentTab({
         accessible. The agent connects outbound to BetterDB Cloud via WebSocket.
       </p>
 
+      <p className="text-xs text-muted-foreground border rounded-md bg-muted p-2">
+        <span className="font-medium text-foreground">Which token?</span> The{' '}
+        <span className="font-medium text-foreground">Agent token</span> below is for the Docker /
+        npx agent (<span className="font-mono">BETTERDB_TOKEN</span>). To connect Claude Code
```
jamby77: **This hint doesn't disambiguate the two tokens.**

It tells the user the agent uses `BETTERDB_TOKEN` — but `packages/mcp/src/index.ts:51` reads `process.env.BETTERDB_TOKEN` too. The env var name is the one thing the agent and the MCP client *share*, so pointing at it is the least distinguishing detail available, and a user comparing the two hints side by side will reasonably conclude the tokens are interchangeable.

They aren't — an agent token in an MCP config (or vice versa) will authenticate as the wrong principal or be rejected. What actually separates them is *where* the variable is set: `-e BETTERDB_TOKEN=…` on the Docker/npx invocation here, versus the `env` block of the MCP server entry in Claude Code's config.

Suggest rewording both hints around that, and saying outright that the tokens are not interchangeable.

apps/web/src/components/pages/settings/McpTokensPanel.tsx
```diff
@@ -153,6 +153,21 @@ export function McpTokensPanel(): ReactElement {
...
+      {/* The Agent token lives only in cloud mode (the "Via Agent" tab and the
+          /agent-tokens endpoint are cloud-only), so the MCP-vs-Agent disambiguation
+          only makes sense there. In self-hosted there is no Agent token to confuse
+          this with, and pointing at a "Via Agent" tab that isn't rendered would be
+          misleading. */}
+      {isCloud === true && (
+        <p className="text-xs text-muted-foreground border rounded-md bg-muted p-2">
+          <span className="font-medium text-foreground">Which token?</span> The{' '}
+          <span className="font-medium text-foreground">MCP token</span> below is for Claude Code's
+          MCP config (<span className="font-mono">BETTERDB_TOKEN</span>). To run the Docker / npx
```
jamby77: Same as the note on `ConnectionSelector.tsx` — this hint names `BETTERDB_TOKEN`, which is the same variable the agent reads, so the pair doesn't tell the two tokens apart. The discriminator is that this one goes in the `env` block of the MCP server entry, not on the container/npx command line.

(The `isCloud === true` gate here is fine.)

packages/agent/Dockerfile
```diff
@@ -12,38 +12,60 @@ COPY packages/shared/src ./src
...
-
-# Agent package.json with rewritten dep + production install
+# Only the declared runtime dependencies (AWS SDK, iovalkey, ws) are installed;
+# @betterdb/shared + zod are already bundled into dist/index.js. Strip the
+# workspace-only @betterdb/shared entry so npm (which does not understand the
+# `workspace:` protocol) can install the rest. --omit=dev drops esbuild and the
+# other build-time devDependencies.
 COPY packages/agent/package.json ./
-RUN sed -i 's|"workspace:\*"|"file:/opt/shared"|' package.json \
+RUN sed -i '/@betterdb\/shared/d' package.json \
```
jamby77: **`sed -i '/@betterdb\/shared/d'` can emit invalid JSON.**

It works today only because `@betterdb/shared` happens not to be the last entry in `devDependencies`. If it ever becomes the last one — someone removes `ws`'s types, or reorders alphabetically — deleting the line leaves the preceding entry with a trailing comma and `npm install` fails on unparseable JSON. It's also a substring match, so a future `@betterdb/shared-testing` would be silently dropped as well.

```dockerfile
RUN npm pkg delete devDependencies.@betterdb/shared \
    && npm install --omit=dev --ignore-scripts
```

Same structural fix applies to the `sed 's|"workspace:\*"|...|'` rewrite in the agent-build stage, though that one is a value substitution and can't break the JSON.

packages/agent/Dockerfile
```diff
@@ -12,38 +12,60 @@ COPY packages/shared/src ./src
...
 
-# Make shared available as a file dependency
+# Make shared resolvable as @betterdb/shared so esbuild can inline it. esbuild
+# resolves shared's own imports (zod) relative to shared's files, so zod must live
+# under /build/shared/node_modules — installing it into the agent's node_modules is
+# not enough (a sibling dir is not on shared's module-resolution path). Locally this
+# is a non-issue because zod is hoisted to the monorepo root.
 COPY --from=shared-build /build/shared/package.json /build/shared/package.json
 COPY --from=shared-build /build/shared/dist /build/shared/dist
+RUN cd /build/shared && npm install --omit=dev --ignore-scripts
```
jamby77: **Neither `npm install` in this Dockerfile has a lockfile, so the image isn't reproducible.**

This one resolves `zod` for `/build/shared`, and the production stage resolves the AWS SDK pieces, `iovalkey` and `ws` — both from bare `package.json` ranges, with no `package-lock.json` or `npm ci`. Two builds of the same commit can produce different dependency trees, and none of it is constrained by `pnpm-lock.yaml`.

That mattered less when this was vendoring a workspace package. It matters more now: `zod` is resolved here and then *baked into* `dist/index.js` by esbuild, so the published image carries whatever `zod` npm happened to serve at build time, permanently, with no record of which version. A CVE in a transitive dep becomes untraceable from the image alone.

Worth generating a lockfile for the agent (or copying a pruned one) and switching both to `npm ci --omit=dev --ignore-scripts`.

packages/agent/scripts/build.mjs
```diff
@@ -0,0 +1,47 @@
...
+import { readFileSync } from 'fs';
+import { dirname, join } from 'path';
+import { fileURLToPath } from 'url';
+
+const agentRoot = join(dirname(fileURLToPath(import.meta.url)), '..');
+const pkg = JSON.parse(readFileSync(join(agentRoot, 'package.json'), 'utf-8'));
+
+// Externalize exactly the declared runtime dependencies. @betterdb/shared and zod
+// are deliberately absent from this list, so esbuild inlines them.
+const external = Object.keys(pkg.dependencies ?? {});
```
jamby77, line 36: Minor / future-proofing: deriving `external` from `dependencies` means the externals list is only correct as long as every direct import is declared there. If someone adds an import that resolves via hoisting at the monorepo root without being added to `dependencies`, esbuild silently inlines it instead of erroring — which is fine for a pure-JS lib, but will produce a broken bundle for anything with native bindings or `__dirname`-relative assets, and the failure shows up at runtime rather than at build time.

The build-time smoke test in the Dockerfile catches the crash-on-boot case, so this is low priority.

KIvanow:
@jamby77 thanks — all addressed in `f0043b40`.

**1. Type-check moved out of PR CI (blocking).** ✅ The agent `build` script is now `tsc --noEmit && node scripts/build.mjs`, so the PR package matrix (`pnpm build --filter @betterdb/agent...`) type-checks again — the gate is back where it was, and also runs in the Docker build and release. Dropped the now-redundant explicit "Type-check agent" step from the release workflow.

**2 & 3. Token hints didn't disambiguate.** ✅ You're right that naming `BETTERDB_TOKEN` — the one thing both share — was the least useful detail. Reworded both around *where* each value goes:
- Agent (ConnectionSelector): "goes on the agent's run command (`-e BETTERDB_TOKEN=…` / `--token`)"
- MCP (McpTokensPanel): "goes in the `env` block of Claude Code's MCP config"

…and both now say outright the values are **not interchangeable**.

**4. `sed` could emit invalid JSON.** ✅ Replaced with `npm pkg delete devDependencies.@betterdb/shared` — edits package.json as JSON, so it's immune to the trailing-comma and substring-match issues.

**5. No lockfile / unreproducible image.** Addressed the core of it by removing the worst offender rather than adding lockfile machinery: **`zod` is now marked external and declared as a runtime dependency**, so it installs into `node_modules` where a scanner can read its version, instead of being baked invisibly into `dist/index.js`. That also let me delete the fragile "install zod under `/build/shared`" step, and the bundle drops from **751kb → ~81kb**. Verified `zod@4.6.5` is present in the image's `node_modules` now.
   - Honest caveat: the runtime `npm install`s still use ranges (no `npm ci`). A committed npm lockfile is awkward here because the agent's `package.json` uses `workspace:*`, which npm can't resolve to generate a lock. The remaining runtime deps (AWS SDK, iovalkey, ws, zod) all land in `node_modules` and are Scout-scannable, so the traceability gap you flagged is closed; happy to add a pruned-lockfile + `npm ci` as a follow-up if you want full install determinism.

**6. `external` derived from `dependencies` (minor).** Left as-is per your low-priority note — the Docker smoke test (`node dist/index.js`) still catches the crash-on-boot case if a needed dep is ever missing from `dependencies`.

Verified locally: agent build type-checks + bundles; `docker build` passes (`npm pkg delete` + zod installed + smoke test green); web tsc clean; web tests 29/29; agent tests 15/15. Ready for another look.

jamby77:
@KIvanow thanks — checked `f0043b40` against each thread and all six are addressed. The `tsc --noEmit &&` in `build` puts the gate back in the PR matrix, the hints now say *where* each value goes and that they aren't interchangeable, `npm pkg delete` is the right fix, and making `zod` external is a better answer than the lockfile I asked for — it's scannable in `node_modules` again, the `/build/shared` install step is gone, and the bundle shrinks to ~81kb. Your reasoning on `npm ci` vs. `workspace:*` is fair; fine as a follow-up. Resolving the threads.

**One blocker before this can merge: `pnpm-lock.yaml` wasn't updated for the new `zod` dependency**, so every Node CI job fails at install:

```
ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml
is not up to date with <ROOT>/packages/agent/package.json
```

A local `pnpm install` isn't frozen, which is why it passed on your machine. Running `pnpm install` and committing the lockfile should fix it. It also means the type-check gate hasn't actually run in CI yet — no job got past install — so I'll approve once the matrix is green, including `package (agent)`.

KIvanow:
@jamby77 good catch — that was a real miss. I added `zod` to `packages/agent/package.json` but didn't regenerate `pnpm-lock.yaml`, and since local `pnpm install` isn't frozen it passed here while CI's `--frozen-lockfile` failed at install (so, as you said, nothing got as far as the type-check gate).

Fixed in `ce2a41d0` — ran `pnpm install`, committed the updated lockfile. Confirmed `pnpm install --frozen-lockfile` now reports "Lockfile is up to date", so the matrix (including `package (agent)`) should get past install and actually exercise the `tsc --noEmit` gate this time.

jamby77 (approved):
Lockfile is in and the full matrix is green — including `package (agent)`, which now runs `tsc --noEmit && node scripts/build.mjs`, so the type-check gate is confirmed in PR CI. Thanks for turning all of this around quickly.

PR #464, opened 2026-09-17
fix(tests): update broker specs for shipped brk-2026-09 embedded key

Kathircpe:
## Summary

Before, `BROKER_SIGNING_PUBLIC_KEYS` was empty so `resolveWorkspaceConfig({})` returned `brokerEnabled: false`. Now the embedded key makes it `true` in dev/self-hosted, which is the intended launch behavior.

## Changes
- `workspace-config.spec.ts`: expect embedded keys + `brokerEnabled: true` by default
- `system-workspace.spec.ts` / `workspace-auth.module.spec.ts`: expect `broker: true` on `/system/workspace`

## Checklist
- [ ] Unit / integration tests added
- [ ] Docs added / updated
- [ ] [Roborev](https://www.roborev.io/) review passed — run `roborev review --branch` or `/roborev-review-branch` in Claude Code *(internal)*
- [ ] Competitive analysis done / discussed *(internal)*
- [ ] Blog post about it discussed *(internal)*

## Summary by CodeRabbit

- **Bug Fixes**
  - Self-hosted workspace status now correctly reports the broker as enabled by default.
  - Default broker configuration uses the embedded signing keys when no environment overrides are provided.
  - Broker authentication now consistently recognizes the default signing configuration, improving connectivity for self-hosted deployments without additional setup.

PR #466, opened 2026-09-18
feat(agent): support agent connections on self-hosted (workspace-enabled)

KIvanow:
Enables the **agent** feature — the "Via Agent" connection tab, `/agent-tokens` minting, and the `/agent/ws` gateway — for **self-hosted** instances with workspace auth enabled. It was previously cloud-only, gated by `CLOUD_MODE` on the backend in three independent places.

## Why it was cloud-only, and what actually gates it
A subagent traced the backend: the agent gateway does its **own** bearer-token auth (no session/cloud-auth dependency) and uses the same `ConnectionRegistry`/storage available everywhere — there's no broker, tenant routing, or cloud API gateway in the data path. So the feature was cloud-gated by policy (`CLOUD_MODE`), not by architecture.

## The two things that made this non-trivial
1. **Duplicate `/agent-tokens` controllers.** Cloud uses `AgentTokensController` (in `AgentModule`); self-hosted uses `PersonalTokensController`. Fastify rejects duplicate routes, so I can't just load `AgentModule` in self-hosted.
2. **Incompatible token formats.** Agent tokens are JWTs (`AgentTokensService`, `jwt.verify` at the gateway); the personal-token path mints opaque `bdb_mcp_` strings the gateway rejects. So agent tokens *must* come from `AgentTokensService`.

## Approach
**Backend**
- New `SelfHostedAgentModule` (loaded when `!isCloudMode()`) provides the proprietary `AgentGateway` + `AgentTokensService` via `createAgentGatewayProviders` — a graceful no-op if the proprietary code isn't built. Mutually exclusive with cloud `AgentModule`, so `AgentGateway` is never double-provided.
- `main.ts` resolves the agent WS gateway unconditionally (the same class registers in both modes) instead of cloud-only.
- `PersonalTokensController` routes `type:'agent'` → `AgentTokensService.generateToken` (JWT) while `type:'mcp'` stays on the opaque personal-token path; adds `GET /agent-tokens/connections` and `?type=agent` listing; `DELETE` falls back to agent-revoke **only for genuine agent tokens** (never bypasses the mcp ownership check). Agent mint stays behind `requireSession` + `@AllowMembers` — authenticated in enabled mode.

**Frontend**
- `ConnectionSelector` shows **Direct + Via Agent** in self-hosted (`mode === 'self-hosted'`); the cloud-only **"BetterDB Valkey instances"** provisioning tab stays cloud-gated. (This widens the pre-existing `isCloudMode` gate rather than adding new machinery.)

## Scope
Self-hosted with workspace **enabled** — the real deployment (you log into your own instance). `WORKSPACE_DISABLED` is intentionally **not** wired: it has no token-controller stack at all, and that mode is already fully unauthenticated by design, so an agent-mint endpoint there would add a parallel controller for **no** security benefit.

## Verification
- `apps/api` + `apps/web` `tsc --noEmit`: clean.
- `personal-tokens` specs: **18/18**.
- **New full-graph boot spec** (`self-hosted-agent.module.spec.ts`) boots the real module graph (StorageModule + ConnectionsModule + SelfHostedAgentModule + WorkspaceAuthModule enabled) and proves: DI resolves across module boundaries (`AgentGateway` ← global `ConnectionRegistry`), **no Fastify route collision**, `/agent-tokens/connections` registers (401 not 404), and the injected gateway/service return live data.
- `ConnectionSelector` web tests: **16/16**.
- The 3 broker spec failures on this branch are pre-existing (the `brk-2026-09` default) — **fixed by #464**, unrelated to this change.

## Follow-ups (not in this PR)
- **UI gating is inferred, not authoritative.** The tab shows on `mode === 'self-hosted'`, a proxy for "agent backend wired." If the proprietary code isn't built, the backend 503s but the tab still shows. A robust fix: add `agent: boolean` to the `/system/workspace` status (from whether `AgentGateway` resolved) and gate on that. I left it out here because that response object is exactly what #464's specs assert on — worth doing after #464 lands to avoid churn.
- `WORKSPACE_DISABLED` support, if ever wanted.

## Note for reviewers
`SESSION_SECRET`: agent-token JWTs are signed/verified with `process.env.SESSION_SECRET || ''`. The real gate is the **storage hash lookup** (a forged JWT has no matching stored `tokenHash`), so an empty secret doesn't enable forgery — but setting `SESSION_SECRET` on self-hosted instances is still recommended.

---

> [!NOTE]
> **Medium Risk**
> Changes touch long-lived JWT signing secrets, admin-scoped workspace credentials, and WebSocket upgrade wiring—important for security and operational stability, though gated off cloud and workspace-disabled modes.
> 
> **Overview**
> Brings **agent connections** (JWT minting, live agent list, WebSocket gateway) to **self-hosted, workspace-enabled** deployments without loading the cloud `AgentModule` and its duplicate `/agent-tokens` routes.
> 
> **Backend:** A new global `SelfHostedAgentModule` wires proprietary `AgentGateway` / `AgentTokensService` when not in cloud mode and workspace is enabled. At import time it **seeds `SESSION_SECRET`** from the same persisted `AUTH_SECRET` path (with strength checks and ephemeral-secret warnings) so agent JWT minting does not 500 when `SESSION_SECRET` is unset. `PersonalTokensController` now handles **`type: 'agent'`** via the agent service (admin-only for mint/revoke), adds **`GET /agent-tokens/connections`** and **`?type=agent` listing**, and returns **503** when agent wiring is missing. `main.ts` resolves the agent WebSocket gateway whenever the class is registered (cloud or self-hosted). `resolveAuthSecretWithSource` and trimmed **`AUTH_SECRET`** validation support that seeding.
> 
> **Frontend:** The add-connection UI shows **Direct + Via Agent** for self-hosted as well as cloud; managed Valkey provisioning stays cloud-only. Agent run commands pick **`ws` vs `wss`** from page HTTPS and loopback, with a warning when HTTP on a non-loopback host would require TLS before `wss` works.
> 
> **Tests:** Full module-graph boot verification plus controller/DTO coverage for the new paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34ab9eeca9853e74108a826a48de57c34f8e0894. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added support for creating, listing, and revoking self-hosted agent tokens.
  - Added an endpoint for viewing connected agents.
  - Enabled “Via Agent” connections in cloud and self-hosted modes.
  - Agent functionality now loads automatically for supported self-hosted deployments.
  - Agent connection commands now use the correct WebSocket protocol for the current page.

- **Bug Fixes**
  - Improved startup handling when agent components are unavailable.
  - Improved session-secret fallback and persistence.
  - Added warnings for insecure WebSocket connections and ephemeral secrets.


apps/api/src/workspace/personal-tokens.controller.ts
```diff
@@ -45,6 +90,9 @@ export class PersonalTokensController {
     if (isDemoHost(req) === true) {
       throw new ForbiddenException(DEMO_TOKENS_MESSAGE);
     }
+    if (body.type === 'agent') {
```
jamby77, line 94: **`@AllowMembers()` on this route was scoped to personal tokens; `type: 'agent'` quietly inherits it for a workspace-scoped credential.**

The decorator on `create` predates this change and was correct for what the route used to do: an mcp token is *your own* token, so letting a member mint one is self-service with no blast radius outside their account. `MutationGuard` reads `ALLOW_MEMBERS_KEY` off the handler, so the exemption applies to the whole handler — it can't distinguish the two branches.

An agent token is a different kind of object. It has `userId: null`, it isn't owned by anyone, it authenticates a machine that registers a Valkey connection visible to the entire workspace, and `AgentTokensService.generateToken` signs it with `expiresIn: '365d'`. So after this change any member — including someone invited as read-only, since `@AllowMembers()` is precisely the escape hatch from the read-only guard — can mint a year-long workspace-wide credential.

Before this PR that request returned 400 in self-hosted, so this is new reachable surface, not a pre-existing gap. Cloud isn't a precedent either: `AgentTokensController` has no role decorators because cloud has no admin/member split to express.

Suggest splitting the authorization along the same line the handler already splits on — keep `@AllowMembers()` for `type: 'mcp'` and require admin for `type: 'agent'`. Since the guard is per-handler, that most likely means an explicit role check inside `createAgentToken`:

```ts
if (actor.role !== 'admin') {
  throw new ForbiddenException('Agent tokens can only be created by workspace admins');
}
```

If members minting agent tokens *is* the intent, it's worth saying so in the PR body — it's the one part of this change that widens who can do something, rather than where it can be done.

apps/api/src/workspace/personal-tokens.controller.ts
```diff
@@ -76,18 +137,72 @@ export class PersonalTokensController {
...
+        targetType: 'token',
+        targetId: token.id,
+      });
       return { revoked: true };
+    } catch (error) {
+      // Personal (mcp) revoke throws NotFound for agent tokens (they are type
+      // 'agent' with userId null, so PersonalTokenService never sees them). Fall
+      // through to the agent service — but ONLY for genuine agent tokens, so this
+      // never bypasses the mcp ownership check that also raises NotFound.
+      if (error instanceof NotFoundException && this.agentTokens) {
```
jamby77, line 165: **Same authorization gap on the revoke side, and here it's unconditional.**

The comment is accurate about what this protects — it won't let anyone through the mcp ownership check. But once the id *is* a genuine agent token, there is no ownership check left to apply, because agent tokens have `userId: null` and nothing records who minted them. The membership test is `agentTokens.some((token) => token.id === id)`, which is an existence check, not an authorization check.

Combined with `@AllowMembers()` on the handler, that means any member can revoke any agent token in the workspace, including ones they didn't create. Revoking an agent token drops a live Valkey connection for everyone, so a read-only member can disconnect the workspace's monitored instances. Unlike the mint path this has no legitimate self-service reading at all — there's no "your own" agent token to revoke.

Same suggested fix: gate the agent branch on `actor.role === 'admin'`. The activity record just below is written either way, which is good, but audit-after-the-fact isn't a substitute for the check.

(Minor, separate: `listTokens('agent')` loads every agent token to answer a membership question. A `getAgentTokenById` would be cheaper and wouldn't need the `.some()`, though at realistic token counts this is cosmetic.)

apps/api/src/agent/self-hosted-agent.module.ts
```diff
@@ -0,0 +1,65 @@
...
+  const configured = (process.env.SESSION_SECRET ?? '').trim();
+  if (configured.length < MIN_SESSION_SECRET_LENGTH) {
+    if (configured.length > 0) {
+      logger.warn(
+        `SESSION_SECRET is shorter than ${MIN_SESSION_SECRET_LENGTH} characters; ignoring it and ` +
+          'using the workspace auth secret for agent-token signing instead.',
+      );
+    }
+    const dataDir = process.env.BETTERDB_DATA_DIR || join(process.cwd(), 'data');
+    process.env.SESSION_SECRET = resolveAuthSecret(process.env, dataDir);
```
jamby77: **This can seed an ephemeral secret while logging that it seeded a persisted one.**

`resolveAuthSecret` doesn't always return the stored secret. When `AUTH_SECRET` is unset *and* the secret file can't be read or written — read-only `BETTERDB_DATA_DIR`, a volume that didn't mount, a container running as a uid that can't write `data/` — it falls back to `ephemeralSecret()`, a fresh `randomBytes(32)` per process. The line below then logs `Seeded SESSION_SECRET … from the workspace auth secret` unconditionally, so the operator is told the opposite of what happened.

The consequence is worse for agent tokens than for sessions. `resolveAuthSecret`'s own warning describes the cost as "sessions are invalidated on every restart" — recoverable, the user logs in again. For agent tokens it means every 365-day token stops verifying at restart, and the agent is a remote process someone has to go re-provision by hand. The stated goal in the comment above ("non-empty AND stable across restarts") is exactly the property that isn't guaranteed here.

Worth having `resolveAuthSecret` signal which branch it took (or re-deriving the check here) and logging a warning when the value is ephemeral, so this shows up at boot rather than at the next restart.

Two smaller things about this block:

- It runs as an **import-time side effect**, so `require`-ing this module mutates global process state. The boot spec already had to work around it — `delete process.env.SESSION_SECRET` plus a `require()` inside `beforeAll` with an eslint disable — which is a fair signal that the ordering is awkward. An `onModuleInit` (or a provider factory) would make it explicit and testable without the module-registry dance.
- Overriding a *configured* `SESSION_SECRET` that's under 32 characters is a silent breaking change for anyone migrating an instance that already has agent tokens signed with a shorter one. The `logger.warn` is there, which helps, but consider failing fast instead — a deployment that deliberately set the variable is better off with a startup error than with tokens that quietly stop validating.

apps/api/src/app.module.ts
```diff
@@ -129,6 +130,20 @@ try {
...
+if (!isCloudMode()) {
+  // Self-hosted agent connections: mirror the cloud AgentModule capability without
+  // the cloud-only AgentTokensController (PersonalTokensController mints agent tokens
+  // in self-hosted). Loaded only outside cloud mode so it never double-provides
+  // AgentGateway alongside the cloud AgentModule.
+  try {
+    const selfHostedAgent = require('./agent/self-hosted-agent.module');
+    SelfHostedAgentModule = selfHostedAgent.SelfHostedAgentModule;
+    console.log('[Agent] Self-hosted agent module loaded');
+  } catch {
```
jamby77: **This `catch {}` can only hide real errors — the module it wraps is first-party and always present.**

The other `try`/`catch` blocks in this file guard `require`s of `../../../proprietary/…`, which legitimately may not exist in an OSS build, so swallowing is the right call there. `./agent/self-hosted-agent.module` ships in `apps/api/src` and is always in the build. `MODULE_NOT_FOUND` is not a reachable outcome.

What *is* reachable is a throw from the module's import-time body: `resolveAuthSecret` doing filesystem work (`mkdirSync`/`writeFileSync` on `BETTERDB_DATA_DIR`), or any future error in the seeding block. Those get caught here, `SelfHostedAgentModule` stays `null`, the comment says "not available in this build", and the instance boots with agent connections silently off — the failure mode is indistinguishable from the intended one, with nothing in the logs pointing at the cause.

Either drop the `try`/`catch` and let a genuine failure fail the boot, or at minimum log the error:

```ts
} catch (error) {
  console.warn('[Agent] Failed to load self-hosted agent module:', error);
}
```

The graceful-degradation case you actually want is already handled one level down, in `createAgentGatewayProviders`, which catches the proprietary `require` and logs why.

apps/api/src/workspace/personal-tokens.controller.ts
```diff
@@ -64,10 +112,23 @@ export class PersonalTokensController {
...
+      const service = this.requireAgent();
+      const tokens = await service.listTokens('agent');
+      return tokens.map(toAgentTokenView);
+    }
     return this.tokens.list(actor);
   }
 
+  @Get('connections')
+  getConnections(): AgentConnectionInfo[] {
+    return this.gateway ? this.gateway.getConnectedAgents() : [];
```
jamby77: Minor, consistency: a missing gateway is reported two different ways. `list(type: 'agent')` goes through `requireAgent()` and throws 503 `AGENT_UNAVAILABLE_MESSAGE`, but this returns `[]` — which is indistinguishable from "the feature works, no agents are connected."

That matters specifically for the follow-up you flagged in the description: when the proprietary code isn't built, the tab shows, minting 503s, and this endpoint reassures the UI that there are simply no agents yet. Returning 503 here too would at least make the two halves of the page tell the same story, and would give the frontend something to gate on in the interim if the `agent: boolean` field on `/system/workspace` lands later.

Agreed on deferring the `/system/workspace` change until after #464 — that's the right call to avoid churning its specs.

apps/api/src/main.ts
```diff
@@ -206,21 +206,24 @@ async function bootstrap(): Promise<void> {
...
-            return null;
-          }
-        })()
-      : null;
+    // Resolve the agent WebSocket gateway if it is registered. Cloud provides it via
+    // AgentModule; self-hosted provides the same class via SelfHostedAgentModule
+    // (createAgentGatewayProviders). Both require the identical module file, so the
+    // class reference matches and app.get resolves in either mode. Null when the
+    // proprietary agent code is not built (agent upgrades are then rejected).
+    const agentGateway = (() => {
```
jamby77, line 214: Minor / scope: this now resolves the gateway in every non-cloud mode, and `app.module.ts` gates `SelfHostedAgentModule` on `!isCloudMode()` alone. So under `WORKSPACE_DISABLED` the module still loads, `SESSION_SECRET` still gets seeded, and `/agent/ws` still accepts upgrades — while `PersonalTokensController` doesn't exist there (it lives in `WorkspaceAuthModule`), so nothing can mint a token for it.

Not a vulnerability — the stored-hash lookup in `validateToken` means an unreachable mint endpoint really does mean no valid tokens. But the Scope section says `WORKSPACE_DISABLED` "is intentionally **not** wired", and the code wires two thirds of it. Worth either gating the module on workspace-enabled as well, or adjusting the wording to "the gateway loads but no tokens can be issued" so the next reader doesn't have to re-derive it.

jamby77 (commented):
Good PR — the "it was gated by policy, not architecture" finding is right, the full-graph boot spec is the correct way to prove the DI/route-collision question, and the `SESSION_SECRET` and `ws://` follow-ups both landed properly. I verified the two security claims in the description and they hold: `AgentTokensService.validateToken` does `jwt.verify` **and then** a `getAgentTokenByHash` lookup, so an empty or weak secret doesn't enable forgery; and the mcp ownership check really is preserved, because a foreign mcp id won't appear in `listTokens('agent')` and the `NotFoundException` is rethrown.

One thing I do think needs to change before merge, plus a few smaller ones.

**Blocking:** `@AllowMembers()` now covers a workspace-scoped credential, so any member can mint an agent token and revoke anyone's.

**Worth fixing:** the import-time `process.env.SESSION_SECRET` mutation can silently be ephemeral; the `catch {}` in `app.module.ts` wraps a first-party module and can only hide real errors.

**Minor:** `getConnections()`/`list()` disagree on how a missing gateway is reported; `main.ts` now wires the gateway in `WORKSPACE_DISABLED` too, which the Scope section says is out of scope.

Details inline.

KIvanow:
@jamby77 thanks for the thorough review — all of it is addressed in `0582c52a`. Ready for another look.

**Blocking — authorization**
- ✅ **Agent mint** (`create`) and **agent revoke** (the fallback branch) are now gated on `actor.role === 'admin'` via a `requireAgentAdmin(actor)` helper — `mcp` stays member self-service. You're right that the revoke existence-check (`.some(id)`) was never an authorization check; there's no owner to fall back on for an unowned `userId: null` token, so admin is the bar. Added member-rejection tests for both (403).

**Worth fixing**
- ✅ **Ephemeral secret mislabeled**: exactly the trap you described — `resolveAuthSecret` can return a per-process value when the data dir isn't writable, and every 365-day agent token would then break on restart while the log claimed a persisted secret. `auth-secret` now exposes `resolveAuthSecretWithSource(): { secret, ephemeral }`, and the module logs a loud **warning** when the secret is ephemeral (naming `AUTH_SECRET` / writable `BETTERDB_DATA_DIR` as the fix) instead of the misleading success line.
- ✅ **`catch {}` in `app.module.ts`**: agreed it can only hide real errors for a first-party module. It now `console.warn`s the error (the genuine OSS-degradation path is already handled a level down in `createAgentGatewayProviders`).

**Minor**
- ✅ **`getConnections()` vs `list()`**: `getConnections()` now throws `503` when the gateway is absent, matching `list('agent')`, so both halves of the page tell the same story.
- ✅ **`WORKSPACE_DISABLED` scope**: rather than reword the Scope note, I made the code match it — `SelfHostedAgentModule` is now gated on `!isCloudMode() && !WORKSPACE_DISABLED`, so under `WORKSPACE_DISABLED` the module no longer loads, `SESSION_SECRET` isn't seeded, and `main.ts` resolves a null gateway → `/agent/ws` upgrades are rejected. It's now genuinely unwired there.

One I deliberately left: the `listTokens('agent')` full-load for the revoke membership check. You flagged it as cosmetic at realistic token counts, and a `getAgentTokenById` would mean touching the storage interface + every adapter — happy to do it as a follow-up if you'd prefer, but it seemed out of proportion for this PR.

Verified locally: `apps/api` `tsc` clean; `personal-tokens` + `self-hosted-agent` + `auth-secret` specs **38/38**.

jamby77 (approved):
All six findings are addressed in `0582c52a`, and `34ab9eec` covers the whitespace-only `AUTH_SECRET` case. Agent mint and revoke are admin-gated with specs asserting a member gets 403, the ephemeral-secret case now warns at boot, the first-party load failure is logged, `/connections` 503s consistently with `list`, and the module stays off under `WORKSPACE_DISABLED`. Thanks for the quick turnaround.

One note for the changelog rather than a change request: because the `AUTH_SECRET` strength check now trims, an install whose secret only reached 32 characters through padding whitespace will get a new generated secret on upgrade, which signs everyone out once.

